PR #25, opened 2026-04-20
[DEV-1484] Retrospective tool access via kapacitor mcp judge

alexeyzimarev:
## Summary

- Replaces the retrospective judge's embedded compacted trace (~1.1 MB prompts that triggered mid-call auto-compaction on default Sonnet — see DEV-1491 for the stabilisation step) with on-demand MCP tool access. The judge now pulls `get_session_recap` / `get_session_errors` / `get_transcript` slices instead of us pre-shipping everything.
- New subcommand `kapacitor mcp judge --session <sessionId>` — session-scoped MCP server exposing the three tools. Each tool validates its `session_id` arg against the one bound at launch.
- `ClaudeCliRunner.RunAsync` gains two optional parameters (`mcpConfigJson`, `allowedTools`). When MCP is opted in, the `--strict-mcp-config` / `--tools \"\"` / `--disallowedTools LSP` lockdown is dropped in favour of `--mcp-config` + `--allowedTools`. `--disable-slash-commands` stays on unconditionally.
- `EvalService.RunRetrospectiveAsync` engages the new path: builds an inline MCP config (`JsonObject` + `JsonArray`, same pattern as `ReviewCommand.cs`), bumps `--max-turns` 3 → 10 and timeout 5 min → 15 min, restricts the judge to the three `mcp__kapacitor-review__*` tools.
- **Per-question judges and title generation are unchanged.** They stay text-only; this PR is retrospective-only.
- Retrospective prompt drops `{TRACE_JSON}` and gains a tool-use guide with an "at most 6 tool calls" budget line.

Linear: [DEV-1484](https://linear.app/kurrent/issue/DEV-1484) (parent: [DEV-1474](https://linear.app/kurrent/issue/DEV-1474)). Unblocks DEV-1485 (session-scoped summary/search/untruncated-tool-result endpoints) and DEV-1486 (per-question opt-in).

## Implementation notes

- **Plan:** `docs/superpowers/plans/2026-04-20-dev-1484-retrospective-tool-access.md` — 8 tasks in two batches, subagent-driven execution with spec + code-quality review per batch and a final cross-cutting review before this PR.
- **Squash-merge expected** — the 10 commits on this branch are for per-task reviewability; history becomes one commit on `main`.
- **Server-side:** no changes. All three endpoints (`/api/sessions/{id}/recap`, `/errors`, `/api/review/sessions/{id}/transcript`) already existed — the new MCP tools just wrap them.

## What the smoke test showed

Ran against session `6458c581…` (1.1 MB compacted trace — would have hit compaction before DEV-1491). The retrospective:

- produced a 127 KB prompt (was ~1.1 MB);
- ran in a single assistant turn with a clean `StructuredOutput` call;
- emitted no `compact_boundary` event;
- synthesised from the 13 per-question verdicts directly, without calling any `mcp__kapacitor-review__*` tool — within spec (the prompt says \"ground every point in the verdicts below **OR** in evidence pulled via the MCP tools\"), but it means the allowed-tool name-match is live-but-unexercised by smoke. First real eval where the judge reaches for a tool will confirm the prefix form. If wrong, the judge will error on first call and it's a one-line fix.

The retrospective rendering bug the smoke also surfaced was a server-side omission in the `EvalRunRegistry` snapshot path — fixed in Kurrent.Capacitor PR #495 (DEV-1494), already merged.

## Test plan

- [x] 253/253 unit tests pass (`dotnet run --project test/kapacitor.Tests.Unit/kapacitor.Tests.Unit.csproj`).
- [x] Release publish clean — zero IL3050 / IL2026 AOT warnings.
- [x] `kapacitor mcp judge --session <id>` probed interactively — `tools/list` returns the three tools, stdio JSON-RPC responds correctly.
- [x] Locally-built binary exercised end-to-end against a real session; retrospective completed cleanly.
- [x] Final review flagged a latent double-encoding defect on meta-session slugs (`ctx.EncodedSessionId` was passed where raw `ctx.SessionId` should go) — fixed in `0c0597d`.

## Out of scope (parked)

- Session-scoped `get_session_summary` / `search_session` / `get_tool_result` endpoints → DEV-1485.
- Per-question opt-in to tools (`needs_tools` flag on the catalog) → DEV-1486.
- `--max-budget-usd` cost cap → revisit once real 1M-context spend is measurable.
- Built-in `Read` against the repo working copy → deferred until retrospective tool usage stabilises.

### DIFF
--- a/docs/superpowers/plans/2026-04-20-dev-1484-retrospective-tool-access.md
+++ b/docs/superpowers/plans/2026-04-20-dev-1484-retrospective-tool-access.md
@@ -1,0 +1,520 @@
+# DEV-1484: Retrospective-only tool access (PR A)
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Replace the ~1.1 MB compacted trace embedded in the retrospective prompt with an on-demand MCP tool guide. The judge pulls what it needs (recap, errors, transcript slices) instead of us pre-shipping everything.
+
+**Architecture:** Add a new session-scoped MCP server subcommand (`kapacitor mcp judge`) that exposes three tools wrapping existing server endpoints. `RunRetrospectiveAsync` launches the judge with `--mcp-config` pointing at that subcommand, `--allowedTools` listing the three MCP tools, `--max-turns 10`, 15-min timeout. Per-question judges and title generation are untouched. The server-side follow-up (default `sonnet[1m]` from Kurrent.Capacitor) stays out of scope of this plan.
+
+**Tech Stack:** .NET 10 AOT, TUnit + WireMock.Net, `claude` CLI MCP protocol (stdio JSON-RPC 2024-11-05), existing HTTP endpoints.
+
+---
+
+## File structure
+
+- Create: `src/kapacitor/Commands/McpJudgeServer.cs` — session-scoped MCP server (three tools, all accept `session_id` as tool arg; validates against a single expected session ID bound at launch time).
+- Modify: `src/kapacitor/Program.cs:233-257` — add `kapacitor mcp judge --session <id>` subcommand dispatch.
+- Modify: `src/kapacitor/Resources/prompt-eval-retrospective.txt` — drop `{TRACE_JSON}`, add tool-use guide block, tighten budget language.
+- Modify: `src/kapacitor/Eval/EvalService.cs` — `RunRetrospectiveAsync` gets new ClaudeCliRunner parameters (MCP config, allowed tools, turn/timeout overrides); drop `traceJson` from `BuildRetrospectivePrompt` signature.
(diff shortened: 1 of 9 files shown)


alexeyzimarev:
All three Qodo findings addressed in `52b653b`:

1. **Correctness — `KAPACITOR_URL` injection.** `RunRetrospectiveAsync` now takes `baseUrl`, and the MCP config for the spawned `kapacitor mcp judge` carries `env: { KAPACITOR_URL: baseUrl }` alongside `command`/`args` — same pattern as `ReviewCommand.cs`. Child process reaches the exact server the parent daemon resolved, regardless of whether the parent used `--server-url` or env.
2. **Reliability — `HttpResponseMessage` disposal.** `McpJudgeServer.HandleToolCallAsync` wraps the switch-assigned response in `using var` so the connection-pool slot releases as soon as the tool call returns. Same shape exists in `McpReviewServer` but it's pre-existing — out of scope for this PR; happy to file a small follow-up.
3. **Maintainability — `allowedTools` contract.** `ClaudeCliRunner.RunAsync` now throws `ArgumentException(paramName: nameof(allowedTools))` when `mcpConfigJson` is supplied without a non-empty allowlist. Today's only MCP caller (retrospective) supplies one, so no behaviour change — but a future misuse surfaces as a loud error rather than a silent tool-surface widening. Two unit tests cover null and empty forms.

Tests: 255/255 unit. AOT clean.

PR #47, opened 2026-05-07
[AI-62] Daemon ends hosted-agent sessions on stop/exit

alexeyzimarev:
## Summary

Fixes orphan "active" sessions for hosted agents (AI-62). When the daemon stops a hosted claude (or observes it exiting unexpectedly), nothing was guaranteed to fire claude's `session-end` hook. With no `SessionEnded` event in the stream, the session stayed marked active forever in the read model — and no session summary or what's-done was generated.

Three changes in one PR:

1. **Graceful shutdown attempt before SIGTERM** — `HandleStopAgent` now writes `/exit\r` to the PTY and waits 15s for claude to fire its own `session-end` (which still runs the existing what's-done generation when configured) before falling back to SIGTERM/SIGKILL.

2. **Daemon-driven `EndAgentSession` call** — both `HandleStopAgent` and `ReadAgentOutputAsync`'s `finally` block now call the new `EndAgentSession` hub method (with reasons `agent_stopped` and `agent_exited` respectively). The server-side handler is idempotent: a no-op if claude already fired session-end, otherwise drains the watcher, appends `SessionEnded`, and computes summary stats.

3. **`WatcherDrainComplete` cts bug** — the watcher was passing `cts.Token` (already cancelled when the final-drain block runs) to `InvokeAsync`, so the call threw `OperationCanceledException` synchronously and the server-side `StopAndDrainAsync` always timed out at 10s. Switched to a fresh 5s token.

## Test plan

- [x] CLI unit tests pass (404 ✓)
- [x] AOT publish for `kapacitor` produces no IL3050/IL2026 warnings
- [x] AOT publish for `kapacitor-daemon` produces no IL3050/IL2026 warnings
- [ ] Manual: hosted agent stopped via UI → SessionEnded appears in stream, session shows Ended in dashboard
- [ ] Manual: hosted agent killed externally (`kill -9`) → daemon emits SessionEnded with reason `agent_exited`

## Server-side

This PR depends on a server change in `kurrent-io/Kurrent.Capacitor` that adds the `EndAgentSession` hub method and the `SessionEndReason.AgentStopped` / `AgentExited` enum values. The server PR also bumps this submodule reference. Server-side defaults gracefully when the daemon isn't yet upgraded (no client calls the new method).

### DIFF
--- a/src/Kapacitor.Core/Models.cs
+++ b/src/Kapacitor.Core/Models.cs
@@ -470,6 +470,7 @@ record RepoEntry {
 [JsonSerializable(typeof(AgentRunStopped))]
 [JsonSerializable(typeof(AgentRunHeartbeat))]
 [JsonSerializable(typeof(PermissionDecision))]
+[JsonSerializable(typeof(EndAgentSessionResult))]
 [JsonSerializable(typeof(int))]
 [JsonSerializable(typeof(string))]
 [JsonSerializable(typeof(string[]))]
@@ -742,3 +743,17 @@ record AgentRunStopped(
 record AgentRunHeartbeat(
         string? SessionId
     );
+
+/// <summary>
+/// Returned by the server's <c>EndAgentSession</c> SignalR hub method. Mirrors the
+/// server-side record of the same name. SessionId is surfaced because the daemon
+/// only knows agentId — it can't spawn <c>kapacitor generate-whats-done</c> without
+/// the sessionId, which the server resolves via FindAgentSessionIdAsync.
+/// </summary>
+public record EndAgentSessionResult {
+    [JsonPropertyName("generate_whats_done")]
+    public bool GenerateWhatsDone { get; init; }
+
+    [JsonPropertyName("session_id")]
+    public string? SessionId { get; init; }
+}
--- a/src/Kapacitor.Daemon/DaemonConfig.cs
+++ b/src/Kapacitor.Daemon/DaemonConfig.cs
@@ -14,6 +14,15 @@ public class DaemonConfig {
 
     public string ClaudePath { get; set; } = "claude";
 
+    /// <summary>
+    /// Path to the kapacitor CLI binary. Used by the daemon to spawn auxiliary
+    /// processes (e.g. <c>generate-whats-done</c>) when claude didn't fire its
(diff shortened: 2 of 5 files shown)


alexeyzimarev:
Both Qodo "Review Recommended" findings addressed in cb7fbf8.

**1. Graceful timeout undetected** — confirmed. PTY `WaitForExitAsync(timeout)` returns silently on timeout (verified in `UnixPtyProcess.cs:169-184` and `ConPtyProcess.cs:331-346` — both `while`-loop until the deadline and return). Added an explicit `HasExited` check after the wait that emits a new `LogGracefulExitTimedOut` warning when the graceful window elapses. The existing try/catch around the wait stays for genuine I/O exceptions.

**2. Stop reason may be lost** — confirmed and fixed via the `PendingEndReason` approach Qodo suggested. New `string PendingEndReason` field on `AgentInstance` defaults to `"agent_exited"`. `HandleStopAgent` flips it to `"agent_stopped"` before its first server call. Both end-session call sites (HandleStopAgent and the read-loop finally) now read from `agent.PendingEndReason`, so whichever call ends up succeeding writes the correct semantic reason — even if a transient SignalR error makes the finally-path call the only one that lands.

Verification: 404/404 unit tests pass, daemon AOT publish clean (no IL3050/IL2026).

alexeyzimarev:
Addressed all three findings in e40da61.

**P1 — `agent.SessionId` never populated.** Confirmed: searched the daemon for any `SessionId =` assignment and there are none. The previous `agent.SessionId is not null` guard was always false, so the generator never ran on the AI-62 path. Server PR #573 now returns `EndAgentSessionResult { GenerateWhatsDone, SessionId }`; daemon spawns `kapacitor generate-whats-done` using the sessionId from the response.

**P3 — `/exit\r` single write.** Confirmed against `HandleSendInput` (same file, ~line 641) which already documents that text and Enter must be split. Now sends `"/exit"`, 50ms delay, `"\r"` — matching the existing pattern.

CLI unit tests: 404/404 pass. Daemon AOT publish clean (no IL3050/IL2026).

(P2 — race during `StopAndDrainAsync` — was server-side; addressed in kurrent-io/Kurrent.Capacitor#573 with a re-check of the stream tail after the drain.)

PR #51, opened 2026-05-08
[AI-78] Refuse foreground `kapacitor agent start` when another daemon is alive

alexeyzimarev:
## Summary

- **CLI half of the AI-78 fix.** The detached path (`-d`/`--detach`) has guarded against duplicate launches via the PID file since day one (`StartDetached` line 67 — `Agent daemon already running (PID …). Use \`kapacitor agent stop\` first.`). Foreground was the only hole. A second `kapacitor agent start` — run by mistake, by an automation, or by a parallel terminal — happily spawns a fresh `kapacitor-daemon`. Cold-start fresh daemons call `DaemonConnect` with empty `live_agents` (the orchestrator hasn't been constructed yet, `GetLiveAgentIds` is null, the `?? []` fallback fires), the server silently replaces the active daemon's `(owner, name)` slot via `DaemonRegistry.Register`, then `ReconcileDaemon([])` flips every running hosted agent to Failed. The displaced real daemon keeps its WebSocket open and never notices.
- **The fix**: mirror the detached PID-file check into `StartForegroundAsync`, write the PID file when the foreground daemon launches so subsequent starts in any mode see it, and clean it up on exit. `IsOurDaemon`'s StartTime check already handles the recycled-PID case if the parent dies hard.
- **Help text + README updated** to document the new behaviour, per the project convention of keeping `help-usage.txt` / `help-<cmd>.txt` / README in sync.

Tested locally:

```
$ kapacitor agent --help        # new "Notes:" section visible
$ kapacitor agent start         # exits 1, message: "Agent daemon already running (PID 54115)…"
```

Independent of the server-side belt (kurrent-io/Kurrent.Capacitor PR 590) which tightens `ReconcileDaemon` to ignore empty `live_agents`. Either alone closes the cascade; both together make it double-safe.

See AI-78 for the full investigation, including macOS process-log evidence of 5 distinct `kapacitor-daemon` PIDs today and confirmation the real daemon's HubConnection never reconnected.

## Test plan

- [x] AOT publish clean (no IL3050/IL2026 warnings).
- [x] Smoke test: `kapacitor agent start` with the existing daemon alive exits 1 with the expected message.
- [x] Help text renders the new "Notes:" section correctly.
- [ ] Reviewer: confirm the PID file is properly cleaned up on Ctrl+C in foreground mode (the `try/finally` deletes it; verified manually but worth a second look).

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,8 +163,10 @@ The agent daemon connects to the Capacitor server and runs Claude Code agents in
 kapacitor agent start              # start in foreground
 kapacitor agent start -d           # start in background (daemonize)
 kapacitor agent status             # check if daemon is running
-kapacitor agent stop               # stop the background daemon
+kapacitor agent stop               # stop the daemon (foreground or background)
 ```
+
+`agent start` refuses to launch a second daemon when one is already alive, regardless of mode. Run `kapacitor agent stop` to terminate the existing one first. Two concurrent daemons under the same identity used to silently take down each other's hosted agents on the server.
 
 ### Repository paths
 
--- a/src/Kapacitor.Core/Resources/help-agent.txt
+++ b/src/Kapacitor.Core/Resources/help-agent.txt
@@ -4,7 +4,7 @@ Usage: kapacitor agent <subcommand>
 
 Subcommands:
   start [-d]              Start the agent daemon (foreground, or -d for background)
-  stop                    Stop the background agent daemon
+  stop                    Stop the agent daemon (foreground or background)
   status                  Show agent daemon status
   logs                    Show recent daemon log output
 
@@ -14,3 +14,9 @@ Options for start:
   --max-agents <n>        Max concurrent agents (default: 5)
   --log-file <path>       Log to file instead of console
   -d, --detach            Run in background (logs to file automatically)
+
+Notes:
+  start refuses to launch if another kapacitor-daemon is already running
+  (whether started in foreground or background mode). Use `kapacitor agent stop`
+  to terminate the existing one first. Multiple concurrent daemons under the
+  same identity were taking down each other's hosted agents on the server.
--- a/src/Kapacitor.Core/Resources/help-usage.txt
+++ b/src/Kapacitor.Core/Resources/help-usage.txt
@@ -21,8 +21,8 @@ Configuration:
   config set <key> <value>         Update a config value
 
 Agent Daemon:
-  agent start [-d]                 Start agent daemon (foreground, or -d for background)
-  agent stop                       Stop background agent daemon
+  agent start [-d]                 Start agent daemon (foreground, or -d for background; refuses if one is already running)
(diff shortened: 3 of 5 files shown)


alexeyzimarev:
Thanks Qodo. Both fixed in 532000d:

**#1 PID file deleted incorrectly** — Real race. Fixed by re-reading `agent.pid` in the finally and only deleting if it still matches the process we spawned. The window is: our daemon exits cleanly → a concurrent `kapacitor agent start` passes the guard (because `IsOurDaemon` correctly returns false on our exited PID) → it writes its own PID-B → our finally deletes the file → daemon-B is orphaned. Now `if (ReadPidFile() is { } current && current.Pid == process.Id) File.Delete(...)`.

I'm leaving the more atomic acquisition (lock file / create-new) as a follow-up; it's a bigger refactor and the conditional-delete closes the window without changing the file-format contract. The remaining TOCTOU between guard-read and `WritePidFile` is benign on the new behavior — the worst case is the second daemon overwrites our PID file (we then no-op the delete, and the second daemon's stop/status keeps working).

**#2 Status parses PID file wrong** — Fixed. `StatusCommand` now takes the first non-empty line of `agent.pid` instead of trying to parse the whole multi-line `PID\nStartTicks` content as one int. Bug existed before this PR too (any `-d` daemon would also hit it) but the foreground guard makes it user-visible for foreground users, which is what Qodo flagged.

I considered consolidating into a single PID-file reader (Qodo's "shared helper" suggestion) and decided against it for this PR — `AgentCommands.ReadPidFile` returns a private `PidEntry` record with `StartTicks`, which `StatusCommand` doesn't need; promoting it costs more API surface than the bug warrants. Worth doing if/when we add a third reader.

alexeyzimarev:
You're right — TOCTOU was real, not benign. Pushed 23a9871 with an actual exclusive lock.

`agent.start.lock` is opened with `FileShare.None` — POSIX `flock(LOCK_EX)` on macOS/Linux, native sharing-mode constraint on Windows. Foreground holds it for the whole daemon lifetime; detached holds it just for the check + spawn + WritePidFile window before the parent exits.

Two concurrent `kapacitor agent start` invocations now serialize at the OS level: the loser's `TryAcquireStartLock` returns null and it refuses with `Another \`kapacitor agent start\` is already in progress or holds the daemon lock.` The lock is per-open-handle, so SIGKILL on the parent releases it cleanly — no zombie locks.

Refactored `StartForegroundAsync` into a thin lock-acquiring wrapper plus a new `SpawnForegroundAsync` that holds the original spawn body, to keep the acquire/release scope obvious.

Re: CI — the test failure (`ImportChainsAsync_dispatches_independent_chains_in_parallel`) is a known-flaky timing assertion (`spreadMs < 160` after a 200ms-delayed parallel HTTP fan-out). It hit `main` itself on May 6 (run 25422621258) with the same assertion. Rerun passed; would be worth bumping the threshold or relaxing the test in a separate PR.

PR #53, opened 2026-05-08
[AI-566] Daemon round-trip Ping replaces fire-and-forget heartbeat

alexeyzimarev:
Pairs with the server-side `DaemonPing` hub method + `EvalRunOrchestrator` fast-fail on slot replacement (server PR https://github.com/kurrent-io/Kurrent.Capacitor pending).

## Why

The current `DaemonHeartbeat` is fire-and-forget `SendAsync`, so when `DaemonRegistry.Register` silently displaces this connection's `(owner, name)` slot (the staging incident — an auto-reconnect that lands a new conn id), the daemon never notices. It keeps pumping heartbeats the server drops, while the orchestrator's in-flight `IClientProxy.InvokeAsync` hangs on the dead conn id for the full 3 / 12 min per-question timeout. See AI-566 for the full root cause.

## Change

- `DaemonHeartbeatLoop` (small testable class behind `IDaemonHeartbeatPort`) replaces the SendAsync heartbeat:
  - tick interval **15 s** (was 60 s) — under the server's default 30 s `ClientTimeoutInterval`
  - per-tick **10 s ping deadline** so a hung WebSocket can't park the loop
  - Ping returns `false` → `ReRegisterAsync` (slot was displaced; re-run `DaemonConnect` so the new conn id takes the slot)
  - Ping throws / times out → `ForceReconnectAsync` (transport hung; `_hub.StopAsync` triggers `OnClosed → ConnectWithRetryAsync` which builds a fresh conn and re-registers)
  - outer cancel → exit cleanly (no reconnect storm at process shutdown)
- `ServerConnection` exposes `PingAsync`, `ReRegisterAsync`, `ForceReconnectAsync` (virtual so the unit test double overrides without needing a real SignalR client) and implements `IDaemonHeartbeatPort` directly.
- `AgentOrchestrator.RunDaemonHeartbeatLoopAsync` delegates each tick to the new loop.

## Tests

`DaemonHeartbeatLoopTests` (5 cases):

- healthy → no side effects
- server says false → ReRegister, no reconnect
- ping throws → force reconnect
- ping exceeds deadline → force reconnect
- outer cancellation (shutdown) → no force reconnect

`dotnet run --project test/kapacitor.Tests.Unit … 428/428 pass`. AOT publish clean for both `kapacitor` and `Kapacitor.Daemon` (no IL3050 / IL2026).

## Test plan

- [x] Unit tests cover the four loop branches + shutdown
- [x] AOT publish clean
- [ ] Smoke: run daemon against staging, simulate a `Register` displacement (or just trigger an eval that previously stuck) and confirm UI advances within ~1 s instead of waiting out the per-question timeout

### DIFF
--- a/src/Kapacitor.Daemon/Services/AgentOrchestrator.cs
+++ b/src/Kapacitor.Daemon/Services/AgentOrchestrator.cs
@@ -75,7 +75,12 @@ internal partial class AgentOrchestrator : IAsyncDisposable {
     readonly LocalPermissionBridge                       _permissionBridge;
     readonly ILogger<AgentOrchestrator>                  _logger;
     readonly PeriodicTimer                               _heartbeatTimer  = new(TimeSpan.FromSeconds(30));
-    readonly PeriodicTimer                               _daemonHeartbeat = new(TimeSpan.FromMinutes(1));
+    // AI-79: heartbeat tightened from 60 s SendAsync to 15 s round-trip Ping.
+    // Server's default ClientTimeoutInterval is 30 s, and the staging incident
+    // showed daemons holding a displaced slot for nearly a minute before
+    // anyone noticed; 15 s puts us comfortably under both.
+    readonly PeriodicTimer                               _daemonHeartbeat = new(TimeSpan.FromSeconds(15));
+    static readonly TimeSpan                             _pingDeadline    = TimeSpan.FromSeconds(10);
     readonly CancellationTokenSource                     _shutdownCts     = new();
 
     public AgentOrchestrator(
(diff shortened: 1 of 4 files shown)


alexeyzimarev:
Addressed both findings in 45983e7 ([diff](https://github.com/kurrent-io/kapacitor/pull/53/commits/45983e7)).

**#1 Heartbeat task can fault** — `TickAsync` is now total. Both recovery actions go through guarded helpers (`SafeReRegisterAsync` / `SafeForceReconnectAsync`); a failed re-register escalates to forced reconnect, a failed forced reconnect logs and lets the next 15 s tick retry. Belt-and-suspenders `try/catch` in `RunDaemonHeartbeatLoopAsync` as defence-in-depth. New tests `Tick_ForceReconnectThrows_DoesNotRethrow` and `Tick_ReRegisterAndForceReconnectBothThrow_DoesNotRethrow` pin the contract.

**#2 Dead heartbeat method** — `ServerConnection.SendHeartbeatAsync` removed. Confirmed no callers in the daemon repo. The server-side `DaemonHeartbeat` hub method stays as a wire-compat alias for older deployed daemons.

Server PR https://github.com/kurrent-io/Kurrent.Capacitor/pull/593 has the submodule bump.

PR #73, opened 2026-05-15
[AI-642] Tighten SignalR keepalive on daemon and watcher hubs

alexeyzimarev:
## Summary

- Halve `HubConnection.KeepAliveInterval` (15s → 7s) and `ServerTimeout` (30s → 15s) on both the daemon's `ServerConnection` and the watch command's hub, so a silently hung WebSocket is surfaced before Cloudflare's recurring edge connection rotation can mask it.
- Halve `AgentOrchestrator._daemonHeartbeat` tick (15s → 7s) and `_pingDeadline` (10s → 5s) so the daemon-level liveness probe stays comfortably under the new 15s server `ClientTimeoutInterval`.

Paired with kurrent-io/Kurrent.Capacitor#625 — both must merge together.

Context: Linear [AI-642](https://linear.app/kurrent/issue/AI-642/stabilize-daemon-signalr-connection-under-cloudflared) — daemon logs show `WebSocketException: remote party closed the WebSocket connection without completing the close handshake` every ~20–30 min through cloudflared. The companion server PR force-switches cloudflared from QUIC to HTTP/2 (the root cause); this PR tightens the keepalive cadence so transport hangs surface faster regardless.

## Test plan

- [ ] AOT publish passes with no IL3050/IL2026 warnings (verified locally)
- [ ] Existing daemon unit tests pass on CI
- [ ] After merge: deploy server companion + this CLI, monitor staging daemon logs for an hour, confirm the `~30 min` reconnect cadence disappears

### DIFF
--- a/src/Kapacitor.Daemon/Services/AgentOrchestrator.cs
+++ b/src/Kapacitor.Daemon/Services/AgentOrchestrator.cs
@@ -76,12 +76,12 @@ internal partial class AgentOrchestrator : IAsyncDisposable {
     readonly IReadOnlyDictionary<string, IHostedAgentLauncher> _launchers;
     readonly ILogger<AgentOrchestrator>                        _logger;
     readonly PeriodicTimer                                     _heartbeatTimer  = new(TimeSpan.FromSeconds(30));
-    // AI-79: heartbeat tightened from 60 s SendAsync to 15 s round-trip Ping.
-    // Server's default ClientTimeoutInterval is 30 s, and the staging incident
-    // showed daemons holding a displaced slot for nearly a minute before
-    // anyone noticed; 15 s puts us comfortably under both.
-    readonly PeriodicTimer                               _daemonHeartbeat = new(TimeSpan.FromSeconds(15));
-    static readonly TimeSpan                             _pingDeadline    = TimeSpan.FromSeconds(10);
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
Addressing Qodo's findings in 8642a5bb (CLI) + kurrent-io/Kurrent.Capacitor@b8fccd60 (server):

1. **15s ServerTimeout rollout risk** — fixed. Reverted `ServerTimeout` and `ClientTimeoutInterval` to the 30s default on both sides. Daemons ship via npm and roll out independently from the server, so the zero-slack scenario was a real risk. Only `KeepAliveInterval=7s` is needed for cloudflared stability; the daemon's application-level heartbeat (7s tick / 5s deadline) already provides fast hang detection independent of SignalR's transport timeout.

2. **StopAsync can stall heartbeat** — fixed. `ForceReconnectAsync` now caps the `StopAsync` await at 5s with a linked CTS; if the transport is wedged, we log and let the next tick retry instead of blocking the heartbeat loop.

3. **Duplicated timeout literals** — declining. Two call sites with one `TimeSpan` literal each; the abstraction cost (a shared `SignalRTimeouts` class referenced from two assemblies) exceeds the drift risk for now. Can revisit if a third caller appears.

PR #75, opened 2026-05-16
[AI-647] Close orphaned session when watcher's parent coding agent exits

alexeyzimarev:
## Summary

- When the watcher's parent-PID monitor detects the parent coding agent died without firing `SessionEnd`, it now POSTs `/hooks/session-end/{vendor}` with `reason: "parent_exited"`, reads back `generate_whats_done`, and spawns the what's-done generator when needed.
- Only fires for session watchers (`agentId is null`) that crossed the transcript threshold; subagent watchers and trivial sessions are skipped.
- Disposes the SignalR connection before posting so the server's `StopAndDrainAsync` skips its 10s drain wait.

## Context

Closes the local-watcher equivalent of the gap the daemon already handles for hosted agents via `AgentOrchestrator.EndAgentSessionAsync`. Without this, users who close their terminal mid-session see the session stuck "Active" in the UI even though the agent is clearly gone.

Paired with kurrent-io/Kurrent.Capacitor#628 — the server change adds `SessionEndReason.ParentExited` plus idempotency guards on `HandleSessionEnd` that protect against the duplicate-POST race when this fallback fires alongside a successful claude session-end. Merge server PR first.

Linear: [AI-647](https://linear.app/kurrent/issue/AI-647/close-orphaned-sessions-when-watchers-parent-coding-agent-exits)

## Test plan

- [x] `WatcherParentExitPostTests` (WireMock) — 3 new tests: payload shape, optional repository, vendor-specific route (`claude` vs `codex`).
- [x] Full CLI integration suite — 12/12 pass.
- [x] Full CLI unit suite — 750/750 pass.
- [x] `dotnet publish -c Release` — clean, no IL3050/IL2026 AOT warnings.

### DIFF
--- a/src/kapacitor/Commands/WatchCommand.cs
+++ b/src/kapacitor/Commands/WatchCommand.cs
@@ -1,5 +1,7 @@
 using System.Runtime.InteropServices;
+using System.Text;
 using System.Text.Json;
+using System.Text.Json.Nodes;
 using System.Text.RegularExpressions;
 using kapacitor.Auth;
 using Microsoft.AspNetCore.SignalR.Client;
@@ -36,6 +38,12 @@ static partial class WatchCommand {
         };
         PosixSignalRegistration.Create(PosixSignal.SIGTERM, _ => cts.Cancel());
 
+        // Tracks whether shutdown was triggered by the parent coding-agent process
+        // dying without firing session-end. When true, the watcher takes over the
+        // server's session-end POST (which the parent normally fires) so the read
+        // model doesn't keep the session "active" forever.
+        var parentExited = false;
+
         // Watch the spawning claude process. If it dies without firing session-end
         // (crash, force-kill, IDE-detach), self-terminate within ~5s instead of orphaning.
         if (parentPid is { } ppid && ProcessHelpers.IsProcessAlive(ppid)) {
@@ -50,6 +58,7 @@ static partial class WatchCommand {
 
                     if (!ProcessHelpers.IsProcessAlive(ppid)) {
                         Log($"Parent pid {ppid} exited; shutting down watcher");
+                        parentExited = true;
                         cts.Cancel();
                         return;
                     }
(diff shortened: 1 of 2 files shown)


alexeyzimarev:
Thanks Qodo. Reviewed all four findings:

**#1 Shutdown POST can hang — accepted, fixed in e85ddbb.**

Real reliability concern. Wrapped the entire helper in a `CancellationTokenSource(TimeSpan.FromSeconds(10))` budget, threaded into `CreateAuthenticatedClientAsync`, `PostWithRetryAsync` (as both `ct` and `timeout`), and the response body read. `OperationCanceledException` becomes a best-effort log + return. Watcher self-terminates in ≤10s on a stalled server, preserving the "self-terminate within ~5s" promise of the parent-PID watchdog.

**#2 Unvalidated vendor in URL — accepted, fixed in e85ddbb.**

Added a `KnownVendors` whitelist (`claude`, `codex`) and a skip + log if vendor is anything else. CLI runs locally so the user would already have code execution to exploit this, but a verbatim path-segment concat is still worth guarding. New WireMock test covers the skip path with a path-traversal vendor.

**#3 Claude route compatibility — declined.**

The server route is `MapPost("/session-end/{vendor=claude}", …)` — `{vendor=claude}` is an optional parameter with default `"claude"`, so `/hooks/session-end` and `/hooks/session-end/claude` both match the same handler. There's no 404 risk against the current server. And since `parent_exited` is a new wire value (introduced by the paired server PR #628), older servers without that PR wouldn't recognize the reason anyway — version skew of this exact form isn't a real scenario.

**#4 `parentExited` visibility race — accepted, fixed in e85ddbb.**

Awaits between write and read already act as memory barriers in practice, but Qodo's right that the C# memory model doesn't formally guarantee it. Promoted to `int` + `Interlocked.Exchange` on write, `Volatile.Read` on read. Cheap and removes the ambiguity.

PR #86, opened 2026-05-21
[AI-666] Reuse SQLite connection for Cursor blob fetches

alexeyzimarev:
## Summary

- Replace per-blob open/query/close with `CursorStateReader.GetContentBlobsAsync` — one `SqliteConnection`, one `SELECT ... WHERE key IN (...)` per batch (chunks of 500 to stay under `SQLITE_MAX_VARIABLE_NUMBER`).
- Typical 6–12 blobs per Cursor session: 1 connection + 1 query instead of 6–12 of each.
- Keeps the single-blob `GetContentBlobAsync` API for back-compat (still used by tests).

Follow-up from Qodo review on #85 (bug #4 — performance).
Linear: [AI-666](https://linear.app/kurrent/issue/AI-666)

## Test plan

- [x] New `GetContentBlobs_*` unit tests: returns requested keys, omits missing keys, dedupes input, empty-input → empty dict.
- [x] All 26 Cursor unit tests pass.
- [x] All 814 CLI unit tests pass.
- [x] AOT publish clean (no IL3050/IL2026 warnings).

### DIFF
--- a/src/Kapacitor.Cli.Core/Cursor/CursorStateReader.cs
+++ b/src/Kapacitor.Cli.Core/Cursor/CursorStateReader.cs
@@ -174,7 +174,13 @@ public static class CursorStateReader {
             for (var i = 0; i < batch.Count; i++) cmd.Parameters.AddWithValue("@k" + i, batch[i]);
 
             await using var reader = await cmd.ExecuteReaderAsync(ct);
-            while (await reader.ReadAsync(ct)) result[reader.GetString(0)] = reader.GetString(1);
+            while (await reader.ReadAsync(ct)) {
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
Pushed fix for the NULL-value finding in a622b194b:

- `GetContentBlobsAsync` now calls `reader.IsDBNull(1)` and skips the row instead of `GetString(1)` throwing `InvalidCastException`. Matches the singular `GetContentBlobAsync`'s null-on-NULL contract that `BuildPayload` already accommodated.
- Added `GetContentBlobs_skips_rows_with_null_value` regression test (NULL row alongside a populated row → only the populated key appears in the result).
- All 12 `CursorStateReaderTests` pass.

PR #87, opened 2026-05-21
[AI-665] Normalize workspaceFolder before Cursor path redaction

alexeyzimarev:
## Summary

- `CursorCommand.ResolveWorkspaces` now normalises `folderPath` once (`Path.GetFullPath` + `TrimEnd`) and yields the canonical native form to all call sites.
- `CursorPayloadAssembler.RedactPaths` tries three variants of the (trimmed) `repoPath` — forward-slash, native backslash, and JSON-escaped backslash (`\\\\`) — replacing the escaped form first so we don't consume half of an escape sequence. Comparison is `OrdinalIgnoreCase` on Windows, `Ordinal` on POSIX.
- Closes the Windows-only path-leak gap where bubble JSON used backslashes and `workspace.json` used forward slashes (or vice versa).

Follow-up from Qodo review on #85 (bug #3 — Windows correctness).
Linear: [AI-665](https://linear.app/kurrent/issue/AI-665)

## Test plan

- [x] `Redacts_paths_with_trailing_separator_on_repoPath` — trailing-slash variant matches.
- [x] `Redacts_backslash_paths_when_repoPath_uses_forward_slash` — escaped Windows form in bubble JSON is stripped.
- [x] `Redacts_forward_slash_paths_when_repoPath_uses_backslash` — inverse case.
- [x] `Redacts_unescaped_backslash_paths_in_object_typed_params` — object-typed params via GetRawText().
- [x] Existing `Redacts_absolute_paths_to_relative` still passes.
- [x] All 26 Cursor unit tests + 814 CLI unit tests pass.
- [x] AOT publish clean.

### DIFF
--- a/src/Kapacitor.Cli.Core/Cursor/CursorPayloadAssembler.cs
+++ b/src/Kapacitor.Cli.Core/Cursor/CursorPayloadAssembler.cs
@@ -93,8 +93,36 @@ public static class CursorPayloadAssembler {
 
     static string? RedactPaths(string? json, string repoPath) {
         if (string.IsNullOrEmpty(json) || string.IsNullOrEmpty(repoPath)) return json;
-        // Cheap: textual replace. Bubbles are short; if false positives become a problem,
-        // upgrade to JSON walk.
-        return json.Replace(repoPath, "/repo");
+
+        // Trim trailing path separators so a repoPath ending in '/' or '\\'
+        // matches the same as one without.
+        var trimmed = repoPath.TrimEnd('/', '\\');
+        if (trimmed.Length == 0) return json;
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
Pushed fix for the silent-skip finding in 13f1c7981:

The new `catch` around `NormalizePath(folderPath)` now writes a one-line diagnostic to `Console.Error` with the workspace storage subdir, the problematic folder path, and the exception message before continuing — so users get an actionable clue when a workspace.json's folder URI is malformed instead of just the generic "No matching Cursor workspace found."

All 26 Cursor unit tests still pass.

PR #94, opened 2026-05-26
[AI-691] Auto-register kapacitor-sessions in plugin manifest

alexeyzimarev:
## Summary

Auto-register `kapacitor-sessions` (and consolidate `kapacitor-review`) via the Kapacitor plugin so they're available out of the box on **both** Claude Code and Codex CLI after `kapacitor setup` — no more manual `claude mcp add` or `~/.config/codex/mcp_servers.toml` edits.

Layout:

- `kapacitor/.mcp.json` — Claude format (`mcpServers` camelCase, with `cwd: ${CLAUDE_PROJECT_DIR}` for sessions). Single source of truth for Claude; picked up automatically by the plugin loader.
- `kapacitor/.codex-plugin/plugin.json` — new Codex manifest. Required `name/version/description` plus `mcpServers: "./.codex-mcp.json"` and `skills: "./codex-skills/"`.
- `kapacitor/.codex-mcp.json` — new file in Codex's `mcp_servers` (snake_case) wrapped shape per [Codex plugin docs](https://developers.openai.com/codex/plugins/build).
- `.claude-plugin/plugin.json` — version bump 1.5.0 → 1.6.0; reverted the inline `mcpServers` block I added in the first commit (now lives in `.mcp.json`).

## Why two MCP files?

Claude Code's `.mcp.json` requires a top-level `mcpServers` (camelCase). Codex's accepts either a bare server map or a `mcp_servers` (snake_case) wrapper, but **not** `mcpServers`. The schemas don't overlap, so a single file can't satisfy both loaders — confirmed against the official docs.

## Not in scope

- **`kapacitor-review` per-PR launcher** — `kapacitor review <pr-url>` keeps writing a temp `--mcp-config` with explicit `--owner/--repo/--pr`. The plugin's auto-registered `kapacitor-review` covers the *opportunistic* case (agent inside a repo with an open PR; CLI auto-detects).
- **`kapacitor mcp judge`** — kept opt-in; behavioral fit for blanket auto-register isn't obvious yet.

## Test plan

- [ ] Fresh `kapacitor setup` → `claude mcp list` shows `kapacitor-sessions` and `kapacitor-review`.
- [ ] Fresh `kapacitor setup` with Codex on PATH → `codex mcp list` shows the same two.
- [ ] Ask the agent *"have we worked on this before?"* in a captured repo — `search_sessions` is invoked without any manual MCP wiring.
- [ ] `kapacitor mcp --help` output is correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,13 +70,19 @@ You must pick an explicit scope (`--all`, `--org`, or `--repo`) so personal/priv
 
 Open the server URL in your browser. The dashboard shows repositories, sessions, and agents. It updates in real time as Claude Code sessions are active.
 
-### Installing the sessions MCP server (for agents)
+### Sessions MCP server for agents
 
-To let coding agents (Claude Code, Codex) search and recall past Capacitor sessions:
+The `kapacitor mcp sessions` stdio server lets coding agents search and recall past Capacitor sessions without leaving the chat.
 
-    claude mcp add kapacitor-sessions -- kapacitor mcp sessions
+For Claude Code users, the Kapacitor plugin (installed by `kapacitor setup`) **auto-registers it** — no extra step. The server is repo-aware: `cd` into a project before spawning your agent and `search_sessions` defaults to that repo's sessions.
 
-The server is repo-aware: `cd` into a project before spawning your agent, and `search_sessions` defaults to that repo's sessions.
+Codex CLI has no equivalent plugin-MCP convention, so Codex users wire it up manually once:
+
+    [kapacitor-sessions]
+    command = "kapacitor"
+    args    = ["mcp", "sessions"]
+
+(in `~/.config/codex/mcp_servers.toml`).
 
 ## What it records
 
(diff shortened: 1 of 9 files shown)


alexeyzimarev:
Addressed Qodo's review in 2d61e2e09:

**1. Unawaited token check always true** — fixed. Added the missing `await` to `TokenStore.LoadAsync()` in `HttpClientExtensions.CreateAuthenticatedClientAsync` so the "expired" vs "not authenticated" branches resolve correctly. Without it, the never-authenticated case would have printed the misleading "Authentication token has expired" message on every spawn.

**2. Sessions MCP eager network call** — fixed. `McpSessionsServer.RunAsync` no longer calls `CreateAuthenticatedClientAsync` at startup; the client is now wrapped in a `Lazy<Task<HttpClient>>` and only constructed on the first `tools/call`. `initialize` and `tools/list` don't need auth, so an auto-spawned session that the agent never queries pays zero network cost. Dispose in a `finally` to avoid leaking the client when the lazy was created.

**3. Plugin README misstates manifest** — already covered by 35b01baa6 (the plugin README rewrite earlier in this PR), which Qodo also caught and marked resolved.

AOT publish clean. 818 unit tests pass.

PR #127, opened 2026-06-07
fix: read default_visibility and excluded_repos from active V2 profile

alexeyzimarev:
## Summary

- A user reported that sessions were being shared to the org despite their profile having `default_visibility: private`.
- Root cause: `ClaudeHookCommand` read `default_visibility` and `excluded_repos` from `AppConfig.Load()` — the legacy v1 `CapacitorConfig` schema. On v2 configs (current format) those fields live under the active profile, not at the top level, so the legacy read silently fell back to defaults: visibility became `org_public` and per-profile `excluded_repos` was ignored entirely.
- Fix: read both settings from `AppConfig.GetActiveProfileAsync()`, matching the pattern already used two lines above for `excluded_paths`. The legacy `AppConfig.Load()` call is dropped from this code path.

The `ProfileCommand show` view already reads from the profile, so users would see `private` configured and have no indication the hook was ignoring it.

## Test plan

- [x] New failing integration test `SessionStartVisibilityTests.Stamps_private_visibility_from_active_profile_v2_config` reproduces the privacy bug (asserted `private`, got `org_public` before fix).
- [x] New failing integration test `Skips_session_start_when_repo_is_excluded_by_active_profile_v2_config` reproduces the excluded-repo bug (asserted 0 requests, got 1 before fix).
- [x] Both tests pass after the fix.
- [x] Full unit suite (1272 tests) and integration suite (25 tests) pass.
- [x] `dotnet publish -c Release` emits no IL3050/IL2026 AOT warnings.

### DIFF
--- a/src/Capacitor.Cli.Core/Config/AppConfig.cs
+++ b/src/Capacitor.Cli.Core/Config/AppConfig.cs
@@ -280,7 +280,34 @@ public static class AppConfig {
             }
         }
 
-        return result.Config;
+        return NormalizeProfileVisibilities(result.Config);
+    }
+
+    /// <summary>
+    /// Coerce each profile's <c>default_visibility</c> to the same set the
+    /// legacy <see cref="Load"/> path enforced (lowercase, restricted to
+    /// <see cref="ValidVisibilities"/>; fall back to <c>org_public</c>
+    /// otherwise). Manual edits and v1→v2 migrations bypass the validation
+    /// that <c>kcap config set</c> / <c>kcap setup</c> apply at write time,
+    /// so a profile on disk can carry through values like <c>"Private"</c>
+    /// or <c>"foo"</c> that the server would reject.
+    /// </summary>
+    static ProfileConfig NormalizeProfileVisibilities(ProfileConfig config) {
+        Dictionary<string, Profile>? rebuilt = null;
+
+        foreach (var (name, profile) in config.Profiles) {
+            var raw        = profile.DefaultVisibility ?? "org_public";
+            var normalized = raw.ToLowerInvariant();
+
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
Pushed `ce4680ed8` to address the validation regression: `LoadProfileConfig` now coerces each profile's `default_visibility` to lowercase and falls back to `org_public` for invalid values, matching what the legacy `AppConfig.Load()` did. Two new integration tests cover `"Private"` → `"private"` and `"totally-bogus"` → `"org_public"`.

## Codex / Cursor follow-up

Checked the server-side contracts in `kcap-server` to confirm scope before extending this PR:

- **Codex** (`/hooks/session-start/codex`): Server already accepts `default_visibility` — it shares `SessionStartHook` with Claude and `SessionHookHandlers.HandleSessionStart` reads `hook.DefaultVisibility` for both vendors (`SessionHookHandlers.cs:125,150`). The CLI never sends it. `SessionWriter.Writes.cs:100-104` has an explicit comment that the server side was fixed for codex visibility but notes the CLI side of the contract. → **Small follow-up PR**: inject `default_visibility` (and run profile-based repo/path exclusion) in `CodexHookCommand.HandleSessionStart`.

- **Cursor** (`/hooks/session-start/cursor`): Server is not ready — `CursorSessionStartHook` doesn't include `default_visibility`, and `CursorHookHandlers.HandleSessionStart` doesn't populate the vendor-neutral `CapacitorSessionBlock` at all, so the server's `VisibilityService` treats Cursor org-repo sessions as org-visible regardless of profile. → **Cross-repo follow-up**: server change first (add field + populate capacitor block), then CLI injection. Larger scope, doesn't belong in this PR.

Keeping this PR Claude-only so the privacy regression ships standalone.

PR #136, opened 2026-06-10
[AI-813] Cap session-end hook pre-POST drain so the session-end POST isn't starved

alexeyzimarev:
## Problem

Claude sessions can stay **Active** in the UI forever after a clean `/exit` — the `AgentSession-…` stream has no `SessionEnded` marker at all, because the terminal signal never reached the server.

The `SessionEnd` hook (`kcap hook --claude`) does `KillWatcher` (≤5s) + `InlineDrainAsync` **before** POSTing `/hooks/session-end`. `InlineDrainAsync`'s `/last-line` GET uses the retry helper's **30s** total timeout — double the **15s** SessionEnd hook budget. Against a slow/remote server this burns the whole budget, Claude kills the hook at its timeout, and the session-end POST is never sent. (The server-side write is fine — it uses `CancellationToken.None` — so the only way `SessionEnded` goes missing is the POST never arriving.)

## Fix

Cap the best-effort pre-POST drain (`KillWatcher` + `InlineDrainAsync`) at **8s** via a new `TimeBudget.RunCappedAsync`, for both `session-end` and `subagent-stop`. If the cap elapses the drain is abandoned and we proceed straight to the POST, guaranteeing it runs within the hook budget. The server's own `StopAndDrain` and the `kcap import` recovery hint cover anything an abandoned drain didn't finish.

Pairs with the server-side safety net in kcap-server (stale-watcher session reaper) — this PR reduces how often the signal is lost; that one guarantees a lost signal can't strand a session forever.

## Tests

`TimeBudgetTests` (new, 3): returns at the cap when work exceeds it; returns immediately when work completes; propagates exceptions raised within the cap.

- Full CLI unit suite: 1321/1321 ✅
- Hook-path integration tests (`ClaudeHookStdoutTests`, `WatcherParentExitPostTests`) ✅
- AOT publish: clean, no IL warnings ✅

### DIFF
--- a/src/Capacitor.Cli/Commands/ClaudeHookCommand.cs
+++ b/src/Capacitor.Cli/Commands/ClaudeHookCommand.cs
@@ -175,7 +175,7 @@ public static class ClaudeHookCommand {
                     var transcriptPath = node?["transcript_path"]?.GetValue<string>();
 
                     if (sessionId is not null) {
-                        await TimeBudget.RunCappedAsync(
+                        var drained = await TimeBudget.RunCappedAsync(
                             async () => {
                                 await WatcherManager.KillWatcher(sessionId);
 
@@ -185,6 +185,13 @@ public static class ClaudeHookCommand {
                             },
                             PreHookDrainCap
                         );
+
+                        if (!drained) {
+                            await Console.Error.WriteLineAsync(
+                                $"[kcap] session-end pre-drain cap ({PreHookDrainCap.TotalSeconds:0}s) elapsed; proceeding to POST. "
+                              + $"Transcript tail may be incomplete — recoverable via: kcap import --session {sessionId}"
+                            );
+                        }
                     }
                 } catch (Exception ex) {
                     Console.Error.WriteLine($"[kcap] session-end pre-hook failed: {ex.Message}");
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
Thanks @qodo-code-review — both findings addressed in e900ba8:

**1. Cap timeout not observable** — Valid. `RunCappedAsync` now returns `bool` (`true` = finished within cap, `false` = cap elapsed / work abandoned). `ClaudeHookCommand` logs a terse stderr line when the cap elapses for session-end and subagent-stop, e.g.:
> `[kcap] session-end pre-drain cap (8s) elapsed; proceeding to POST. Transcript tail may be incomplete — recoverable via: kcap import --session <id>`

So a cut-short drain is now diagnosable instead of silent.

**2. Timeout boundary race** — Valid edge case, fixed. Instead of deciding completed-vs-capped solely from `Task.WhenAny`'s winner, the helper now re-checks `workTask.IsCompleted` after the wait and **awaits the work whenever it has actually finished** — even if the delay "won" a near-boundary race — so a task that completes right at the cap still has its result/exception propagated (honoring the XML-doc contract). Only genuinely-still-running work is abandoned.

Tests updated to assert the `bool` result; full unit suite 1321/1321, hook integration tests pass, AOT publish clean.

PR #139, opened 2026-06-11
[AI-803] Keep --strict-mcp-config on for tool-using headless judges

alexeyzimarev:
## Problem

Daemon-based eval judges reported _"Unable to investigate the session transcript — all MCP tool calls were blocked by permission restrictions"_ and scored sessions low (see AI-803 screenshot).

## Root cause

The tool-using judges (retrospective + DEV-1486 per-question) run as headless `claude -p` subprocesses. `ClaudeCliRunner` passed an inline `--mcp-config` (session-scoped `kcap-review` → `kcap mcp judge`) plus an `--allowedTools` allowlist, but **dropped `--strict-mcp-config`** in that branch — it was only set in the text-only branch.

Without `--strict-mcp-config`, `--mcp-config` is *merged* with the user's global MCP servers. On a client the **`kcap` Claude Code plugin** is installed, whose `kcap/.mcp.json` globally registers `kcap-sessions` (`search_sessions` / `get_session_summary` / `get_session_transcript`) plus a colliding `kcap-review`. The judge reached for those leaked, un-allowlisted tools (the prompt asks for a "transcript" tool, and `get_session_transcript` matches more literally than the allowlisted `get_transcript`) → every call blocked by permission restrictions → degraded verdict.

Daemon-only because the leak needs the plugin installed: server-side evals use `IChatClient` (no CLI/plugin) and CI has no plugin, so only the inline server loads there.

## Fix

Emit `--strict-mcp-config` in **both** modes so the judge loads *only* the inline session-bound `kcap mcp judge` server (which also enforces the `expectedSessionId` binding). Extracted arg construction into an internal `BuildClaudeArgs` seam for unit testing.

## Tests
- New `BuildClaudeArgs_*` tests: assert `--strict-mcp-config` is present in MCP mode, the config/allowlist are passed correctly, and the text-only `--tools ""` / `--disallowedTools LSP` lockdown is intact. Verified RED (failed without the flag) → GREEN.
- Full CLI unit suite: 1357 passed.
- `dotnet publish -c Release`: clean, no IL warnings.

### DIFF
--- a/src/Capacitor.Cli.Core/ClaudeCliRunner.cs
+++ b/src/Capacitor.Cli.Core/ClaudeCliRunner.cs
@@ -66,14 +66,17 @@ static class ClaudeCliRunner {
     /// during the run (today: the DEV-1484 retrospective judge, which
     /// pulls session details via <c>kcap mcp judge</c> instead of
     /// having the full trace embedded in its prompt). When
-    /// <paramref name="mcpConfigJson"/> is supplied, the runner flips out
-    /// of the text-only lockdown (<c>--strict-mcp-config</c> / empty
-    /// <c>--tools</c> / <c>--disallowedTools LSP</c>) and instead loads
-    /// the caller-supplied MCP config via <c>--mcp-config</c>, restricting
-    /// the model to exactly <paramref name="allowedTools"/> via
-    /// <c>--allowedTools</c>. <c>--disable-slash-commands</c> stays on in
-    /// both modes. Leaving both null preserves the text-only behaviour
-    /// every other caller relies on.
+    /// <paramref name="mcpConfigJson"/> is supplied, the runner loads the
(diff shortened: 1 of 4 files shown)


alexeyzimarev:
Added a follow-up commit (c52c8b8): rename the inline judge MCP server key `kcap-review` → `kcap-judge`.

It was registered under the same key as the plugin's real `kcap mcp review` server. `--strict-mcp-config` already makes the collision moot at runtime, but the duplicate name was a latent footgun, and `kcap-judge` matches `kcap mcp judge`'s own `serverInfo` + the `help-mcp.txt` docs. The name now lives in `EvalService.JudgeMcpServerName`, the allowlist prefixes are built from it (no drift), and the duplicated inline config is extracted into `BuildJudgeMcpConfig`. New tests pin the name + key/allowlist consistency. Full suite 1360 passing, AOT publish clean.

Issue #141, opened 2026-06-11
Windows: subagent/session watcher self-reap is fragile — parent-PID watchdog may never arm


alexeyzimarev:
## Summary

Follow-up hardening identified while fixing kurrent-io/kcap-cli#140 / kurrent-io/kcap-cli#140. Independent of that fix, the transcript watcher's **self-reap path is fragile on Windows**: the parent-PID watchdog may never arm, so a watcher can orphan whenever its kill hook is missed.

## Detail

* `WatcherManager.SpawnWatcher` records `--parent-pid` from `ProcessHelpers.GetCodingAgentPid()`. On Windows this returns `GetParentPidWindows()` — the **immediate** parent of the `kcap hook` process. For hook invocations that's the transient per-hook process the agent spawns, not the long-lived coding-agent process.
* `WatchCommand.RunWatch` (`src/Capacitor.Cli/Commands/WatchCommand.cs:88-108`) only starts the parent-PID monitor task **if the parent is alive at watcher startup**:

  ```csharp
  if (parentPid is { } ppid && ProcessHelpers.IsProcessAlive(ppid)) { /* start monitor */ }
  ```

  If the transient parent has already exited by the time the watcher initializes, the monitor never starts and the watcher has **no parent-death self-reap at all**.
* Unix avoids this by using the process-group id (`getpgrp`), which resolves to the real agent PID. Windows has no equivalent and falls back to the immediate parent.

## Impact

After the kurrent-io/kcap-cli#140 fix, watchers are reaped via the `SubagentStop` hook's `KillWatcher` (subagents) and the `SessionEnd` hook (main session). But if either kill hook is missed — crash, force-kill, IDE detach, hook failure/timeout — on Windows the watcher can orphan because the parent-PID watchdog never armed. This is the defense-in-depth layer that was silently absent.

## Possible directions

* On Windows, walk up the process tree past transient hook executors to find the long-lived coding-agent PID; or
* have the agent pass its PID/session context explicitly; or
* add a server-side `StopWatcher` / TTL fallback as defense-in-depth (covers all platforms).

## References

* Originating bug: kurrent-io/kcap-cli#140 / kurrent-io/kcap-cli#140 (fixed by clearing `HANDLE_FLAG_INHERIT` on the watcher spawn + `async` subagent hooks)
* `src/Capacitor.Cli/Commands/WatchCommand.cs:88-108`
[...]

Issue #140, opened 2026-06-11
SubagentStart/SubagentStop hooks break Claude Code's Agent/Task tool and leak watcher processes


lev-parcelvision:
# Subagent (`SubagentStart`/`SubagentStop`) hooks break Claude Code's Agent/Task tool and leak watcher processes

## Summary

On Claude Code 2.1.173 (Windows), the `SubagentStart` and `SubagentStop` hooks in kcap `0.6.14` (still present and unfixed in the latest `0.6.15`) cause **every subagent spawn (the Agent/Task tool) to fail** — the tool result comes back as `<error>[Tool result missing due to internal error]</error>` instead of the agent's output. Each failed spawn also **leaks an orphaned `kcap` watcher process** that survives the (already-dead) subagent. Removing just the two `Subagent*` hook blocks fixes both problems with main-session recording fully intact.

## Environment

| | |
|---|---|
| kcap | `0.6.14+040b61fa3d2a1583b1c3e632a2ba5f04a097f852` (`@kurrent/kcap`, plugin `kcap@kcap`) |
| Claude Code | `2.1.173` |
| OS | Windows 11 Pro 10.0.26200 |
| Install | global npm: `C:\Users\<user>\AppData\Roaming\npm\node_modules\@kurrent\kcap` |
| Host | Claude Code running in a **PowerShell 7.5.5** (Core) terminal inside **VS Code Insiders** |
| Node | `v22.20.0` (npm `10.9.3`) |

## Symptom

Any Agent/Task tool call returns:

```
<error>[Tool result missing due to internal error]</error>
```

The subagent itself appears to start, but its result never makes it back to the main session. Separately, each subagent spawn leaves a `kcap watch … --agent-id …` process that is **never reaped** even after the subagent (and its whole session) ends — see the evidence below.

## Root cause (high confidence)

The plugin registers these two hook blocks in `kcap/hooks/hooks.json`:

```json
"SubagentStart": [
  { "hooks": [ { "type": "command", "command": "kcap hook --claude", "timeout": 5 } ] }
],
"SubagentStop": [
  { "hooks": [ { "type": "command", "command": "kcap hook --claude", "timeout": 5 } ] }
],
```

`kcap hook --claude` on the `Subagent*` events spawns a per-subagent watcher process of the form:

```
kcap.exe watch <session-id> <…>\subagents\agent-<agent-id>.jsonl --agent-id <agent-id> --parent-pid <claude-pid>
```

[...]

PR #142, opened 2026-06-11
fix: stop hook-spawned watcher inheriting Claude's std handles on Windows (AI-820)

alexeyzimarev:
Fixes the Windows-only bug in **AI-820 / #140**: the `SubagentStart`/`SubagentStop` hooks break Claude Code's Agent/Task tool (`[Tool result missing due to internal error]`) and leak orphaned `kcap watch … --agent-id` processes.

## Root cause

On Windows, `WatcherManager.SpawnWatcher` redirects stdio, so .NET's `CreateProcess` runs with `bInheritHandles: true` and the spawned watcher inherits the hook process's own std handles — which are **Claude's hook-stdout pipe write-ends**. Closing the parent-side copies doesn't touch the inherited copies living in the long-lived watcher, so:

- Claude's read of the **synchronous** `SubagentStart` hook never reaches EOF → it stalls to the 5s timeout → the Agent/Task tool reports the result as missing; and
- the disrupted flow never fires a clean `SubagentStop`, so `KillWatcher` never runs → the watcher is **orphaned** (the ~24h processes in the report).

Unix is immune: `fork`+`exec` `dup2`s the redirect pipes over fds 0/1/2 and drops the originals, so there's no inherited copy of Claude's pipe — which is why this was never reproduced on macOS/Linux, and why the async `SessionStart` watcher (reliably reaped by `SessionEnd`) was unaffected.

## Changes

- **Root cause:** clear `HANDLE_FLAG_INHERIT` on the process's std handles before spawning the watcher and the whats-done generator (`ProcessHelpers.PreventInheritedStdHandles`, Windows-only via `SetHandleInformation`; no-op on Unix). This fixes **both** symptoms — Claude's hook stdout EOFs immediately, and `SubagentStop` then fires normally to reap the watcher.
- **Belt-and-suspenders:** mark `SubagentStart`/`SubagentStop` hooks `async: true` so Claude never blocks on their stdout even on a build that doesn't honor the handle fix.
- **Regression test:** `ProcessHelpersHandleInheritanceTests` verifies the inherit bit is actually cleared on an inheritable handle (real assertion on Windows; safe no-op elsewhere).
- **CI:** run `build-and-test` on `[ubuntu-latest, windows-latest]` so the test genuinely executes on PRs.

## Testing

- Build clean (0 warnings); 1323 unit tests + watcher/hook integration tests pass locally.
- AOT publish (`dotnet publish -c Release`) — no IL2026/IL3050 warnings.
- ⚠️ The suite has **not** run on Windows CI before, so this PR's first Windows run may surface pre-existing Windows-incompatible tests unrelated to AI-820 — those get triaged separately.

## Follow-up

- #141 — Windows watcher self-reap fragility (parent-PID watchdog may never arm), independent of this bug.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Embedded prompt/help resources are baked into the binary at build time and
+# compared byte-for-byte at runtime (e.g. kcap-subsession prompt detection).
+# Pin them to LF so a Windows checkout (core.autocrlf) can't smuggle CRLF into
+# the embedded copy and break those comparisons. See AI-820.
+src/Capacitor.Cli.Core/Resources/*.txt text eol=lf
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,8 +16,12 @@ env:
 
 jobs:
   build-and-test:
-    name: Build and test
-    runs-on: ubuntu-latest
+    name: Build and test (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
 
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,12 @@ jobs:
         run: dotnet run --project test/Capacitor.Cli.Tests.Unit/Capacitor.Cli.Tests.Unit.csproj -- --maximum-parallel-tests 1
 
       - name: Run integration tests
+        # Linux only: the integration suite does real loopback HTTP (WireMock).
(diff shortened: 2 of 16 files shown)


alexeyzimarev:
Both Qodo findings addressed in fc5a6170d:

1. **Silent inherit-clear failures** — `PreventInheritedStdHandles` now emits a single stderr diagnostic per process (with the win32 error) when `SetHandleInformation` genuinely fails on a valid std handle, so a recurrence of the AI-820 symptom is no longer silent. Absent handles (NULL/INVALID) are still skipped quietly and the path still never throws.

2. **Test mutates std handles** — the unit test no longer calls `PreventInheritedStdHandles()` (which cleared the inherit flag on the test host's real STDIN/STDOUT/STDERR without restoring them). The actual clearing behaviour remains covered in isolation by the inheritable-pipe test.

PR #150, opened 2026-06-13
[AI-848] Guard Process.Start against launch failures in CLI runners

alexeyzimarev:
Fixes AI-848.

`ClaudeCliRunner.RunCoreAsync` / `CodexCliRunner.RunCoreAsync` called `Process.Start("claude"/"codex")` unguarded. When the executable isn't on the (often minimal, e.g. launchd) PATH of a detached hook / what's-done process, `Process.Start` throws `Win32Exception`, the throw escapes to `Main`, and the NativeAOT runtime aborts the whole CLI with SIGABRT. Observed as a 1-in-111 what's-done summary crash (`kcap-2026-06-13-112006.ips`) where `claude` (`~/.local/bin/claude`) wasn't reachable on the reparented process's PATH. The existing `if (process is null)` guard only covers `Start` *returning* null, not *throwing*.

## Change
- New `CliProcess.TryStart(psi, log)` — wraps `Process.Start` in try/catch, logs the failure, returns null, so callers fall through their existing `process is null → return null` path instead of crashing.
- Wired into both `ClaudeCliRunner` and `CodexCliRunner`.

## Test plan
- New `CliProcessTests.TryStart_MissingExecutable_ReturnsNullAndLogs` — TDD'd: RED reproduced the `Win32Exception`, GREEN returns null.
- Full unit suite green; `dotnet publish -c Release` (AOT) clean, no IL2026/IL3050.

### DIFF
--- a/src/Capacitor.Cli.Core/ClaudeCliRunner.cs
+++ b/src/Capacitor.Cli.Core/ClaudeCliRunner.cs
@@ -193,7 +193,7 @@ static class ClaudeCliRunner {
             psi.ArgumentList.Add(arg);
         }
 
-        using var process = Process.Start(psi);
+        using var process = CliProcess.TryStart(psi, log);
 
         if (process is null) {
             log("Failed to start claude process");
--- a/src/Capacitor.Cli.Core/CliProcess.cs
+++ b/src/Capacitor.Cli.Core/CliProcess.cs
@@ -1,0 +1,31 @@
+using System.Diagnostics;
+
+namespace Capacitor.Cli.Core;
+
+static class CliProcess {
+    /// <summary>
+    /// Starts a child process, converting a failed launch into a logged
+    /// <c>null</c> instead of letting the exception escape.
+    ///
+    /// <para>
+    /// A bare <see cref="Process.Start(ProcessStartInfo)"/> throws
+    /// <see cref="System.ComponentModel.Win32Exception"/> when the target
+    /// executable can't be found (e.g. <c>claude</c> not on PATH). In a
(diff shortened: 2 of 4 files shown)


alexeyzimarev:
Thanks @qodo-code-review — addressed in e8b0579ea:

**2. Catch-all masks start failures (Bug)** — Fixed the diagnosability half: the log now includes the exception type (`ex.GetType().Name: ex.Message`). I kept the catch broad **by design** and added a comment saying so: `TryStart` is a crash-guard whose contract is "no launch failure may abort the CLI". Narrowing to `Win32Exception`/`InvalidOperationException` would let an unanticipated launch exception escape to `Main` and re-introduce the SIGABRT this PR exists to prevent. Callers intentionally treat any null as "couldn't launch the helper → skip generation"; the logged type preserves the per-mode detail.

**3. Fixed temp path in test (Bug)** — Fixed: the test now uses `Path.Combine(Path.GetTempPath(), $"kcap-nonexistent-binary-{Guid.NewGuid():N}")` and asserts `File.Exists(...)` is false up front, so it can't depend on external filesystem state.

**1. README not updated (Rule violation)** — Dismissing as a false positive. The README-sync rule targets the *user-facing CLI surface* (new command, new/renamed/removed flag, changed default behavior, new prerequisite — updating `## Getting started` / `## CLI commands`). This PR adds none of those: it's an internal robustness fix to background title/what's-done generation that turns a process abort into a logged skip. `claude`/`codex` on `PATH` was already an implicit requirement for those features; this PR doesn't add a prerequisite, it just stops crashing when it's unmet. Nothing in the public CLI docs changes.

PR #151, opened 2026-06-13
[AI-847] Report local transcript line count on WatcherConnect (CLI)

alexeyzimarev:
Part 2 of 2 for AI-847 (CLI side; server side: kurrent-io/kcap-server#733).

Pairs with the server-side resume guard. The watcher now passes its current local transcript line count (`CountFileLines(transcriptPath)`) to `WatcherConnect` at both call sites (initial connect + reconnect) so the server can tell a genuine resume of an ended session (local transcript has lines past the server's resume point) from a watcher with nothing new — only the latter gets `session_already_ended`.

The reconnect path also covers the overnight reconnect-to-ended case (the `Server behind (2188 vs 2192)` loop), where a still-running watcher reconnects after the server has marked the session ended.

## Compatibility
- Older servers ignore the extra arg; older watchers send none and the new server falls back to its `-1` sentinel (preserving today's behaviour).

## Test plan
- CLI unit suite green; `dotnet publish -c Release` (AOT) clean, no IL2026/IL3050.

### DIFF
--- a/src/Capacitor.Cli/Commands/WatchCommand.cs
+++ b/src/Capacitor.Cli/Commands/WatchCommand.cs
@@ -167,7 +167,7 @@ static partial class WatchCommand {
 
             // Re-register with server and check if it's behind us (gap recovery)
             try {
-                var serverPosition = await hubConnection.InvokeAsync<int>("WatcherConnect", sessionId, agentId, cancellationToken: cts.Token);
+                var serverPosition = await hubConnection.InvokeAsync<int>("WatcherConnect", sessionId, agentId, CountFileLines(transcriptPath), cancellationToken: cts.Token);
 
                 if (serverPosition < state.LinesProcessed) {
                     Log($"Server behind ({serverPosition} vs {state.LinesProcessed}), rewinding to resend gap");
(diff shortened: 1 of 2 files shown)


alexeyzimarev:
Good catch @qodo-code-review — valid bug. `CountFileLines` returns `0` for both a missing file and an IO error, so an unreadable transcript would be reported as "0 lines", which the server's resume guard reads as "nothing new" and would use to stop a live resume — re-introducing the very freeze this fixes.

Fixed across both halves:

- **CLI (8980b533c, this PR):** new `LocalLineCountForConnect(path)` returns **-1 ("unknown")** when the transcript is missing or unreadable (distinct from a genuine, readable `0`), and is used at both `WatcherConnect` call sites. `CountFileLines` keeps its `0`-on-missing contract (WatcherManager + its test depend on it); the shared read loop is extracted to `CountLines`. Added tests for the `-1` path.
- **Server (kurrent-io/kcap-server#733, e60b1e63):** `WatcherResumeDecision.ShouldStopEndedWatcher` now stops only on a **confirmed** count (`localLineCount >= 0 && <= resumePoint`); an unknown/negative count **fails open** (don't stop). Without a confirmed count we can't rule out a resume, and freezing a live resume is worse than a genuinely-idle ended watcher lingering (still cleaned up by the explicit session-end drain + parent-PID exit).

Nice side effect: failing open on the `-1` default also means deploying the server alone fixes the freeze for already-deployed watchers (which send no count).

alexeyzimarev:
Closing in favor of a **server-only** fix. Per the SignalR constraint (hub method signatures must match exactly on both sides — optional parameters don't provide version compatibility), adding `localLineCount` to `WatcherConnect` would be a breaking protocol change that fails under any server↔CLI version skew. 

kurrent-io/kcap-server#733 now fixes the freeze entirely server-side: it simply stops pre-emptively killing a watcher on an already-`Ended` session, so a resume streams its new lines and reactivates via `SessionSummaryFold`. No CLI change is needed, and it fixes the freeze for already-deployed watchers on deploy.

Reopen if we later want the "promptly stop genuinely-idle ended watchers" optimization — that would need a lockstep signature change or a new additive hub method with client fallback.

PR #152, opened 2026-06-13
Retry hosted-agent permission requests across SignalR reconnects

alexeyzimarev:
## Summary

A transient SignalR WebSocket blip during a hosted-agent permission wait was being turned into an automatic **deny**, breaking the user's conversation with the agent. SignalR cancels in-flight `InvokeAsync` calls when the transport drops (surfacing as `TaskCanceledException`), and `LocalPermissionBridge` caught that and fell back to deny — even though the daemon's auto-reconnect recovers the connection seconds later (the recurring cloudflared instability pattern).

This change adds reconnect-aware retry **inside** `ServerConnection.RequestPermissionAsync` so a permission request survives a blip and only resolves to deny on a genuinely non-recoverable outcome.

- **`RegistrationGate`** — tracks "connected **and** re-registered" readiness. Bracketed around every `RegisterDaemon()` (cleared at start, set on success) plus reset on `Reconnecting`/`Closed`, so it also covers the heartbeat slot-displacement re-register path (which fires no transport event).
- **`ConnectionRetry`** — retry loop over injected delegates. `OperationCanceledException` (transport killed an in-flight invoke) and `InvalidOperationException`-while-not-ready are transient → wait for readiness, retry. `InvalidOperationException`-while-ready and `HubException` (and anything else) propagate → the bridge's existing `catch → deny` is now the *final* fallback only. Bounded solely by the daemon-shutdown token; end-to-end still capped by the hook client's ~10h timeout.
- **`ServerConnection`** — holds the gate, exposes `IsReady`, drives it from the connection lifecycle, and routes `RequestPermission` through the retry helper. New `LogPermissionRetry` (info) replaces the per-blip warning-with-stack-trace.

No user-facing CLI surface change, so no README/help update required.

Design + plan: `docs/superpowers/specs/2026-06-13-permission-bridge-reconnect-retry-design.md`, `docs/superpowers/plans/2026-06-13-permission-bridge-reconnect-retry.md`.

## Test Plan

- [x] `RegistrationGateTests` (5) — readiness transitions incl. slot-displacement without transport loss
- [x] `ConnectionRetryTests` (6) — recovery, cancellation, `HubException`/`InvalidOperationException` classification, already-recovered race
- [x] `LocalPermissionBridgeTests` (19) — unchanged, still green (incl. `ServerFailureFallsBackToDeny`)
- [x] Full unit suite: 1467 passed / 0 failed
- [x] `dotnet publish src/Capacitor.Cli.Daemon/Capacitor.Cli.Daemon.csproj -c Release` — 0 IL2026/IL3050 warnings

### DIFF
--- a/docs/superpowers/plans/2026-06-13-permission-bridge-reconnect-retry.md
+++ b/docs/superpowers/plans/2026-06-13-permission-bridge-reconnect-retry.md
@@ -1,0 +1,655 @@
+# Resilient Hosted-Agent Permission Requests Across SignalR Reconnects — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** A transient SignalR disconnect during a hosted-agent permission wait no longer falls back to `deny`; the daemon waits for the connection to recover (connected *and* re-registered) and retries, only denying on daemon shutdown or a genuinely non-recoverable failure.
+
+**Architecture:** Add reconnect-aware retry inside `ServerConnection.RequestPermissionAsync` via two small, isolated, unit-testable units: a `RegistrationGate` (tracks "connected and registered" readiness) and a `ConnectionRetry` helper (the retry loop over injected delegates). `LocalPermissionBridge` is unchanged — its `catch → deny` becomes the final fallback that now only fires on non-recoverable outcomes.
+
+**Tech Stack:** .NET 10 (NativeAOT), `Microsoft.AspNetCore.SignalR.Client`, TUnit + `Assert.That(...)` on Microsoft Testing Platform.
+
+**Design doc:** `docs/superpowers/specs/2026-06-13-permission-bridge-reconnect-retry-design.md`
+
+---
+
+## File Structure
+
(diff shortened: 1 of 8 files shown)


alexeyzimarev:
Addressed all three Qodo findings in 9469a83.

**Bug 1 (invoke before readiness wait) — fixed.** Moved the readiness wait to the top of every attempt in `ConnectionRetry`, so the *first* invoke is gated on connected-and-registered too, not just post-failure retries. Added a `Waits_for_readiness_before_the_first_invoke` test.

**Bug 3 (IOE race) — fixed, but not via the suggested snapshot.** The snapshot-before-invoke approach is incoherent once Bug 1 is fixed: with the readiness wait at the top, `readyBeforeInvoke` is always `true`, so an IOE would *always* be classified final → never retried (worse). The correct resolution falls out of Bug 1 — since we only invoke after observing readiness, an `InvalidOperationException` ("connection is not active") can only mean the transport dropped in the gap between the check and the call, i.e. it's always transient. So `IsTransientDisconnect` no longer consults `isReady()` at all, which removes the race entirely. This supersedes the old "IOE-while-ready is final" rule; its anti-spin purpose is moot because `InvokeAsync`'s IOE is purely connection-state, and genuine permanent faults surface as `HubException`/other and still propagate → deny.

**Bug 2 (stale comment) — fixed.** Updated the `LocalPermissionBridge` comment: the wait is no longer bounded by a single connection's lifetime but by daemon shutdown (`ct`) and the hook client's ~10h timeout, and retries span reconnects.

Design spec updated to match. Full unit suite 1468/0, daemon AOT publish 0 IL2026/IL3050.

PR #161, opened 2026-06-17
[AI-884] Report hosted-agent PTY dimensions to the server

alexeyzimarev:
Linear: [AI-884](https://linear.app/kurrent/issue/AI-884)

_Supersedes #160 (closed by a branch rename)._

## Problem

The hosted-agent **Terminal** tab in the Capacitor UI shows garbled output — wrapped/overlapping text with truncated left edges. Classic PTY-width vs. display-width mismatch.

**Root cause:** hosted-agent PTYs spawn at a fixed **120×40** and never resize, but the daemon only streamed PTY *output* to the hub — it never reported the dimensions. So the server's `GetTerminalDimensions` returned `null`, the browser never received a `TerminalDimensions` event, and the read-only xterm stayed in **auto-fit mode** (~80 cols), rendering a 120-col TUI at the wrong width. Box-drawing and absolute cursor moves landed in the wrong cells.

The full dimension-sync machinery (`SetTerminalDimensions` → broadcast → `setFixedSizeReadOnly`) already existed — it was built for the MAUI rendered-agent path and was simply never wired for hosted daemon agents.

## Change

- Add `HostedPtyCols`/`HostedPtyRows` constants (120/40) and pass them explicitly to `Spawn`, kept in sync with the `IPtyProcessFactory` defaults.
- Fire `SendTerminalDimensionsAsync` right after the agent registers so read-only viewers lock their xterm to the source size. Fire-and-forget — late delivery is harmless, the viewer reflows when dims arrive.

The matching server/UI side (lock + font-scale the locked grid to fit the panel) ships in the kcap-server PR that bumps this submodule.

This is the **interim lock** fix; a server-side virtual terminal is tracked as a separate follow-up (AI-885).

### DIFF
--- a/src/Capacitor.Cli.Daemon/Pty/IPtyProcessFactory.cs
+++ b/src/Capacitor.Cli.Daemon/Pty/IPtyProcessFactory.cs
@@ -1,7 +1,20 @@
 namespace Capacitor.Cli.Daemon.Pty;
 
+/// <summary>
+/// Single source of truth for the fixed size hosted-agent PTYs are spawned at.
+/// Hosted PTYs are never resized; the daemon reports these dims to the server so
+/// read-only viewers lock their xterm to exactly the width Claude drew for (AI-884).
+/// Referenced by both <see cref="IPtyProcessFactory.Spawn"/>'s defaults and the
+/// orchestrator's spawn + dimension-report calls so the two can never drift.
+/// </summary>
+public static class PtyDefaults
+{
+    public const ushort Cols = 120;
+    public const ushort Rows = 40;
+}
+
 public interface IPtyProcessFactory
 {
     IPtyProcess Spawn(string command, string[] args, string cwd,
-        Dictionary<string, string>? extraEnv = null, ushort cols = 120, ushort rows = 40);
+        Dictionary<string, string>? extraEnv = null, ushort cols = PtyDefaults.Cols, ushort rows = PtyDefaults.Rows);
 }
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
Thanks @qodo-code-review — addressed all three findings in `3384496`:

**1. Dims not resent on reconnect (Bug / Reliability)** — Valid, fixed. The server stores dims in memory, so a server restart (not just a daemon blip) wipes them, and `ReRegisterAgentsAsync` re-registered agents without resending dims — reintroducing the garble. Now resends `SendTerminalDimensionsAsync` after the status change in the re-register loop, best-effort with its own catch so a dims-send failure can't trigger a re-registration retry or withhold daemon readiness.

**2. Unobserved send exceptions (Bug / Reliability)** — Valid, fixed. The launch-path send is now `await`ed inside a try/catch that logs with `agentId` (`LogTerminalDimsSendFailed`) instead of fire-and-forget. It stays best-effort (a failed send never fails the launch — late delivery is harmless), but the fault is observed at the call site with agent context rather than landing in the global unobserved-task handler. The reconnect-path resend uses the same logged best-effort pattern.

**3. PTY defaults duplicated (Bug / Maintainability)** — Valid, fixed. Introduced `PtyDefaults` (`Cols`/`Rows`) in the `Capacitor.Cli.Daemon.Pty` namespace as the single source of truth; both `IPtyProcessFactory.Spawn`'s default parameters and the orchestrator's `HostedPtyCols`/`HostedPtyRows` now derive from it, so they can no longer drift.

Daemon builds clean (0 errors). The matching server/UI side is in kcap-server#757, whose submodule pointer was bumped to include these fixes.

PR #162, opened 2026-06-17
[AI-886] Add Pi (badlogic/pi-mono) CLI vendor support

realtonyyoung:
Linear: [AI-886](https://linear.app/kurrent/issue/AI-886/add-pi-badlogicpi-mono-as-a-5th-ingest-vendor)

Adds **Pi** (badlogic/pi-mono, the `pi` CLI) as a CLI ingest vendor. Pi has **no shell hooks**, so live capture ships as a Pi TypeScript extension rather than a hooks file.

## Changes
- **Core:** `PiPaths` (the `~/.pi/agent` layout) and `PiExtensionInstaller`, which embeds `kcap.ts` and writes it to `~/.pi/agent/extensions/` (version-markered, like the Copilot installer).
- **Live:** `kcap hook --pi` (`PiHookCommand`) reads the session JSONL header for the id, POSTs `/hooks/session-{start,end}/pi` (repo/PR-enriched), and spawns the existing detached watcher with `vendor=pi`. The shipped `kcap.ts` extension calls it from `pi.on("session_start"/"session_shutdown")`.
- **Batch:** `kcap import --pi` (`PiImportSource`) walks `~/.pi/agent/sessions/**/*.jsonl`, sharing the import classify/watermark machinery.
- Wired `--pi` into `VendorSelection`, the import source list, the hook dispatcher, `kcap plugin install/remove --pi`, and `kcap setup` auto-detection (`--skip-pi-hooks`).
- Tests: `PiImportSourceTests` + `PiExtensionInstallerTests` (20 pass). README updated.

Builds + AOT-publishes clean (0 IL warnings).

Server-side ingest is in the kurrent-io/kcap-server PR on the same branch.


src/Capacitor.Cli/Commands/PiHookCommand.cs
```diff
@@ -0,0 +1,213 @@
...
+        }
+
+        var file = GetArg(args, "--file");
+        if (string.IsNullOrWhiteSpace(file)) return 0; // ephemeral / no session file — nothing to record
+
+        var header          = await TryReadHeaderAsync(file);
+        var dashedSessionId = header?.SessionUuid ?? Path.GetFileNameWithoutExtension(file);
+
+        // Pi session ids are UUIDs; anything else (a stray .jsonl) isn't a session.
+        if (string.IsNullOrEmpty(dashedSessionId) || !Guid.TryParse(dashedSessionId, out _)) return 0;
```
realtonyyoung: This makes live Pi capture drop the normal `session_start` path. Upstream Pi creates session filenames as `<timestamp>_<uuid>.jsonl`, and `ctx.sessionManager.getSessionFile()` can return that path before the header has been flushed. In that state `TryReadHeaderAsync(file)` returns null, `Path.GetFileNameWithoutExtension(file)` is the whole `2026-..._<uuid>` name, and `Guid.TryParse` fails, so we return before posting `/session-start/pi` or starting the watcher. The later `session-end` inline drain starts at EOF when there is no server watermark, so it does not recover the missed transcript. Please parse the UUID suffix from Pi filenames (and add a test for a missing-header `session-start` file path) or pass the session id from the extension if Pi exposes it.

src/Capacitor.Cli/Commands/PiHookCommand.cs
```diff
@@ -0,0 +1,213 @@
...
+            return 0;
+        }
+
+        var exit = await PostHookAsync(baseUrl, "session-start/pi", enriched);
+        if (exit != 0) return exit;
+
+        await WatcherManager.EnsureWatcherRunning(
+            baseUrl, sessionId, file,
+            agentId: null, sessionIdOverride: null, cwd: cwd,
+            skipTitle: false, vendor: "pi"
```
realtonyyoung, line 121: Since this starts the shared watcher with `vendor: "pi"`, the parent-exit fallback also needs to accept that vendor. `WatchCommand.PostSessionEndOnParentExitAsync` currently gates on `KnownVendors = { "claude", "codex", "copilot" }`, so a Pi watcher that sees the parent process disappear will drain and then skip the `/hooks/session-end/pi` POST as an unknown vendor. That leaves crashed/closed Pi sessions active until a manual import. Please add `pi` to that allow-list and cover the parent-exit path with a small test.

src/Capacitor.Cli/Commands/PiHookCommand.cs
realtonyyoung: Fixed in e165768 — `ExtractSessionId` now prefers the header uuid and falls back to the uuid **suffix** of the `<timestamp>_<uuid>.jsonl` filename, so a `session_start` that fires before the header line is flushed no longer derives a non-uuid id and drop the session + watcher. Added `PiHookCommandTests` covering the missing-header / `<timestamp>_<uuid>` filename path.

src/Capacitor.Cli/Commands/PiHookCommand.cs
realtonyyoung: Fixed in e165768 — added `pi` to `WatchCommand.KnownVendors`, so a Pi watcher whose parent process exits drains and POSTs `/hooks/session-end/pi` instead of skipping as an unknown vendor. Added a `ParentExit_PostsToVendorSpecificRoute_ForPi` case to `WatcherParentExitPostTests`.

realtonyyoung:
New re-review note:

`src/Capacitor.Cli/Commands/WatchCommand.cs` still needs Pi-aware title/event helpers now that `PiHookCommand` starts the watcher with `vendor: "pi"`. The transcript upload itself still works after the raw-line threshold, but the watcher title path falls through to the Claude JSONL shape: `TryExtractUserText(..., "pi")`, `TryExtractAssistantText(..., "pi")`, and `IsEvent(..., "pi")` all read top-level `type: "user"|"assistant"`, while Pi emits `type: "message"` with `message.role: "user"|"assistant"`. That means live Pi sessions never get the initial title or LLM title path, even though historical Pi import opts into title generation. Please add Pi-specific extractor/event branches mirroring the server normalizer user/assistant mapping, plus watcher helper tests for Pi lines.

realtonyyoung:
Addressed in 4186407 — added Pi-aware branches to `WatchCommand.TryExtractUserText` / `TryExtractAssistantText` / `IsEvent`. They handle Pi's `type:"message"` + `message.role` (string or text-block content), mirroring the server `PiTranscriptNormalizer` user/assistant mapping, so live Pi sessions now reach the initial + LLM title path (matching historical import). Added `PiTitleHelperTests` covering user/assistant extraction and the `IsEvent` message-role threshold. CLI builds + AOT-clean (0 IL warnings); 35 Pi unit tests pass.

src/Capacitor.Cli/Commands/WatchCommand.cs
```diff
@@ -768,6 +769,15 @@ internal static bool IsEvent(string line, string vendor = "claude") {
                 };
             }
 
+            if (vendor == "pi") {
+                // Pi conversational envelopes are type:"message" with
+                // message.role user/assistant (Pi has no top-level user/assistant
+                // type). Everything else (model_change, compaction, tool results,
+                // labels, …) is plumbing and must not count toward the threshold.
+                return root.Str("type") == "message"
+                    && root.Obj("message")?.Str("role") is "user" or "assistant";
```
realtonyyoung: This Pi branch should probably use the same “contentful message” predicate as the server normalizer/import relevance, not just the role. `PiTranscriptNormalizer.NormalizeUser` returns null when `content` is empty, and `NormalizeAssistant` returns null when there are no text/thinking/toolCall parts; `PiImportSource.IsImportRelevantLine` similarly requires `HasContent` for user/assistant. As written, empty `message.role:"user"` / `"assistant"` envelopes still increment `state.EventCount`, so a live Pi session can hit the 5-event title threshold on lines that never produce canonical events. Please gate user/assistant here on extractable text or assistant parts, and add false-case tests for empty Pi user/assistant messages.

realtonyyoung: Fixed in 30ef94b — `IsEvent(…, "pi")` now gates user/assistant on **content** via `PiUserHasContent` / `PiAssistantHasContent`, mirroring the server `NormalizeUser`/`NormalizeAssistant` emit condition and `PiImportSource.IsImportRelevantLine.HasContent`: a user needs a non-empty string or a text block, an assistant needs at least one text/thinking/toolCall part. Empty Pi envelopes no longer increment the title-event threshold. Added empty-user/assistant false-cases (and a tool-only-assistant true-case) to `PiTitleHelperTests`.

src/Capacitor.Cli/Commands/PiImportSource.cs
```diff
@@ -0,0 +1,450 @@
...
+            using var doc  = JsonDocument.Parse(line);
+            var       root = doc.RootElement;
+
+            switch (root.Str("type")) {
+                case "session": return true;
+                case "message":
+                    if (root.Obj("message") is not { } msg) return false;
+                    return msg.Str("role") switch {
+                        "user"          => HasContent(msg),
+                        "assistant"     => HasContent(msg),
```
realtonyyoung: This assistant relevance check is still broader than the server normalizer and the watcher predicate added in the latest fix. `HasContent` treats any non-empty assistant content array as import-relevant, but `PiTranscriptNormalizer.NormalizeAssistant` only emits for `thinking`, `text`, or `toolCall` blocks with the required fields; unsupported arrays such as image-only or empty blocks will not advance `/last-line`. If one of those lines sits at the end of the transcript, `lastRelevantIndex` can stay ahead of the server watermark and keep classifying the import as Partial forever. Please mirror the role-specific assistant predicate here and add false cases for unsupported assistant content arrays.

realtonyyoung: Fixed in 0dcef57 — extracted the predicate to `Capacitor.Cli.Core.Pi.PiContent` (`HasUserContent` / `HasAssistantContent`) mirroring `NormalizeUser`/`NormalizeAssistant`, and `IsImportRelevantLine` now uses it: assistant requires a `thinking`/`text`/`toolCall` part (an array of only unsupported blocks like images is no longer relevant), user requires a non-empty string or text block. `WatchCommand.IsEvent` calls the same shared predicate now, so the importer, the watcher, and the server normalizer stay in lockstep. Added assistant-unsupported-block + empty-user import-relevance test cases.

src/Capacitor.Cli.Core/Pi/PiContent.cs
```diff
@@ -0,0 +1,48 @@
...
+/// title-event threshold (<c>WatchCommand.IsEvent</c>) — when it actually
+/// carries content. Shared so those two never drift from the normalizer (or
+/// each other): a role check alone over-counts empty/unsupported envelopes that
+/// the normalizer skips, which would re-classify a complete session as Partial
+/// forever / trip the title threshold on lines that emit nothing.
+/// </summary>
+public static class PiContent {
+    /// <summary>User message has content: a non-empty string, or a text block with non-empty text.</summary>
+    public static bool HasUserContent(JsonElement msg) {
+        if (msg.Str("content") is { Length: > 0 }) return true;
```
realtonyyoung, line 20: This new helper is in `Capacitor.Cli.Core.Pi`, but the `Str`/`Arr` extension methods it calls are declared in the sibling namespace `Capacitor.Cli.Core` (`JsonElementExtensions`). Child namespaces do not see parent-namespace extension methods automatically, and there is no global using for `Capacitor.Cli.Core`, so this should fail to compile with `JsonElement` not containing `Str`/`Arr`. Please add `using Capacitor.Cli.Core;` here (or move the helper into the core namespace) and keep the build test covering the extracted helper.

realtonyyoung: Added `using Capacitor.Cli.Core;` to PiContent in e5f5a41. For what it's worth it does compile as-is — a clean `dotnet build -t:Rebuild` of the CLI is green, and the 45 Pi unit tests + AOT publish on this branch already exercise PiContent via `PiImportSource`/`WatchCommand` (C# extension-method lookup walks enclosing namespaces, and `JsonElementExtensions` lives in the enclosing `Capacitor.Cli.Core`). But the explicit using removes that reliance and keeps the file correct if it's ever moved — so it's in now.

src/Capacitor.Cli/Commands/VendorSelection.cs
```diff
@@ -12,7 +12,7 @@ public sealed record Result(IReadOnlySet<string> Vendors, string? Error) {
         public bool HasError => Error is not null;
     }
 
-    static readonly string[] KnownVendorFlags = ["--claude", "--codex", "--cursor", "--copilot"];
+    static readonly string[] KnownVendorFlags = ["--claude", "--codex", "--cursor", "--copilot", "--pi"];
```
realtonyyoung: Adding `--pi` here also needs to extend the unknown-option guards below to `--pi-`. As written, a typo like `kcap import --pi-session ...` or any future-looking `--pi-*` flag is not rejected by the vendor-prefixed check, and if it is more than edit distance 2 from `--pi` it falls through as an unrelated flag. In the import command that can silently leave `Vendors` empty and import every detected source instead of only Pi. Please include `a.StartsWith("--pi-")` in both vendor-prefixed checks and add a parser test for an unknown `--pi-*` option.

realtonyyoung: Fixed in c219e2c — added `--pi-` to both vendor-prefixed guards (the unknown-option rejection loop and the typo-detection skip), so `--pi-session` / any `--pi-*` is now rejected as an unknown source option instead of silently falling through (and leaving `Vendors` empty = import-all). Tests: `--pi` selects only pi; `--pi-session` is rejected.

README.md
```diff
@@ -75,7 +75,7 @@ For non-interactive environments:
...
 
 > **Need hooks for an agent installed after setup, or scoped to a single repo?**
 > Run `kcap plugin install [--codex|--cursor|--copilot]` (omit the flag for the Claude Code plugin), or pair Codex with `--project` for a per-repo install. Use `--skills` instead of `--codex` if you only want the agent skills without Codex hooks. Cursor uses user-scope only — `--project` has no effect with `--cursor`. After installing Codex hooks, the next `codex` launch prompts to trust the new hooks — accept once to trust them all (run `/hooks` inside Codex if you'd rather trust each entry individually). After a `--project` install, also run `codex` once in the repo and accept the workspace trust prompt. Re-running after a kcap upgrade is rarely needed for user-scope installs — the npm postinstall hook auto-refreshes them on every `npm install -g @kurrent/kcap`, and `kcap update` refreshes them too (npm 11+ blocks install scripts by default — `kcap update` works regardless, or add `allow-scripts[]=@kurrent/kcap` to `~/.npmrc` to opt the postinstall in once).
```
realtonyyoung: This summary now needs to include Pi as a plugin target too. The PR adds `kcap plugin install --pi` and the Pi note below mentions it, but this quick-reference line still advertises only `--codex|--cursor|--copilot`, so users scanning the setup section can miss the supported Pi install flag. Please add `--pi` here (and consider updating the import-filter sentence below that still says pass `--claude`, `--codex`, `--cursor`, or `--copilot`).

src/Capacitor.Cli/Commands/PiImportSource.cs
```diff
@@ -0,0 +1,446 @@
...
+        var result = new List<DiscoveredSession>();
+        var seen   = new HashSet<string>(StringComparer.Ordinal);
+
+        foreach (var jsonl in Directory.EnumerateFiles(_sessionsDir, "*.jsonl", SearchOption.AllDirectories)) {
+            ct.ThrowIfCancellationRequested();
+
+            var header = await TryReadHeaderAsync(jsonl, ct);
+            if (header is not { SessionUuid: { Length: > 0 } uuid }) continue; // not a Pi session file
+
+            var dashless = uuid.Replace("-", "");
```
realtonyyoung: Historical discovery should validate the header id the same way the live hook path does. Right now any non-empty `id` in a `type:"session"` header is accepted and normalized with `Replace("-", "")`, so a malformed/corrupt Pi file under `~/.pi/agent/sessions` can create/import an arbitrary non-GUID session id. `PiHookCommand.ExtractSessionId` already rejects non-GUID headers/filenames for this reason. Please gate this with `Guid.TryParse(uuid, out _)` (or share the extraction helper) and add a discovery test that skips an invalid session header id.

src/Capacitor.Cli/Program.cs
```diff
@@ -604,8 +605,11 @@
         if (args.Contains("--copilot")) {
             return await CopilotHookCommand.Handle(baseUrl!, Console.In, args);
         }
+        if (args.Contains("--pi")) {
```
realtonyyoung, line 625: Now that `watch` accepts `vendor: "pi"` through `PiHookCommand`, the `kcap watch` usage string above still advertises only `claude|codex|copilot`. This is the only help users get when they invoke `kcap watch` incorrectly, and it now hides a supported value that is required for manual Pi watcher/debug runs. Please include `pi` in that usage text and add/adjust a CLI usage test if this command has one.

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -35,15 +36,15 @@ public static async Task<int> HandleAsync(string[] args, PluginEnvironment? env
         };
     }
 
-    static readonly string[] ExclusiveTargetFlags = ["--codex", "--cursor", "--copilot", "--skills"];
+    static readonly string[] ExclusiveTargetFlags = ["--codex", "--cursor", "--copilot", "--pi", "--skills"];
```
realtonyyoung: The command now accepts `--pi` as a plugin target, but `PrintUsage()` still prints `Usage: kcap plugin <install|remove> [--project] [--codex|--cursor|--skills] [--if-installed]`. That means `kcap plugin` / invalid subcommands hide the new Pi install/remove path (and also still omit Copilot). Please update the usage string to include all current targets, including `--pi`, and cover it with a usage-output test if available.

realtonyyoung:
New full-pass finding:

`kcap plugin install/remove --pi` is wired, but the existing lifecycle-management callers were not updated for the new extension. `src/Capacitor.Cli/Commands/UninstallCommand.cs` removes Claude, Codex, Cursor, Copilot, and skills, but never calls `PluginCommand.HandleAsync(["plugin", "remove", "--pi"])`, so `kcap uninstall` can leave `~/.pi/agent/extensions/kcap.ts` auto-loading after the user removed kcap. Likewise `npm/kcap/bin/refresh.js` refreshes skills/Codex/Cursor/Copilot/Claude with `--if-installed`, but never runs `plugin install --pi --if-installed`, so an opted-in Pi extension will not be refreshed by npm postinstall or `kcap update`. Please add Pi to both cleanup and refresh paths, plus tests covering uninstall invoking `--pi` and refresh including the Pi command.

README.md
realtonyyoung: Done in `40feae0` — added `--pi` to the plugin quick-reference line (`[--codex|--cursor|--copilot|--pi]`) and to the import-filter sentence (`--claude`, `--codex`, `--cursor`, `--copilot`, or `--pi`).

src/Capacitor.Cli/Commands/PiImportSource.cs
realtonyyoung: Done in `40feae0` — discovery now reuses `PiHookCommand.ExtractSessionId(jsonl, header.SessionUuid)`, so a corrupt/non-GUID `{"type":"session"}` header is rejected instead of minting an arbitrary id (and it picks up the live path's `<timestamp>_<uuid>.jsonl` filename fallback for an unflushed header). Added two discovery tests: `discovery_skips_session_with_non_guid_header_id` and `discovery_recovers_session_id_from_filename_when_header_id_missing`.

src/Capacitor.Cli/Program.cs
realtonyyoung: Done in `40feae0` — `kcap watch` usage now reads `--vendor claude|codex|copilot|pi`, matching `WatchCommand.KnownVendors`.

src/Capacitor.Cli/Commands/PluginCommand.cs
realtonyyoung: Done in `40feae0` — `PrintUsage()` now lists all current targets: `[--codex|--cursor|--copilot|--pi|--skills]` (it previously omitted both `--copilot` and `--pi`).

realtonyyoung:
New broader-pass finding: the embedded CLI help resources were not updated with the new Pi surface, even though some direct usage strings and the root README were.

Concrete stale spots at `40feae0`:
- `src/Capacitor.Cli.Core/Resources/help-import.txt` still says import discovers Claude/Codex/Cursor/Copilot, says kcap walks all four, and lists vendor filters only through `--copilot`.
- `src/Capacitor.Cli.Core/Resources/help-plugin.txt` has no `--pi` option, no Pi install/remove notes, and no `kcap plugin install --pi` / `remove --pi` examples.
- `src/Capacitor.Cli.Core/Resources/help-setup.txt` omits `--skip-pi-hooks`, omits Pi from the detected-agent description, and its post-setup install example lists only `--codex`, `--cursor`, and `--copilot`.
- `src/Capacitor.Cli.Core/Resources/help-hook.txt` still documents only stdin JSON hook dispatch and only `--claude|--codex|--cursor|--copilot`, but Pi is arg-driven and now supported via `kcap hook --pi --event ... --file ...`.

These back `kcap <command> --help`, so users relying on CLI help will not discover the Pi integration or its skip/install/import flags. Please update the embedded help resources and add/update help-output snapshot tests if the project has them.

realtonyyoung:
New routed-import finding: `PiImportSource.SupportsTitleGeneration` returns `true`, but Pi sessions never actually reach the import title-generation path.

Pi classifications set `FilePath = ""`, so they go through the routed-source phase (`src.ImportSessionAsync`) instead of `ImportChainsAsync`. The only place that queues `GenerateTitleForImportAsync` is the chain worker `OnTitleTaskReady` callback; the routed phase only calls `src.ImportSessionAsync`, updates counts, and never schedules title/background work. Even if a Pi title task were queued, `GenerateTitleForImportAsync` currently only special-cases `vendor == "codex"`; every other vendor uses the Claude-shaped `TitleGenerator.ExtractTitleContext`, which looks for top-level `type:"user"/"assistant"` and would skip Pi `type:"message"` + `message.role` lines.

Net effect: historical `kcap import --pi` advertises/supports title generation in the source contract, but imported Pi sessions will not get LLM-generated titles from the import pipeline. Please either set `SupportsTitleGeneration = false` and document that the server fallback title is the intended behavior, or add a routed-source title hook plus a Pi-shaped title extractor/tests so `--pi` imports match live Pi title behavior.

realtonyyoung:
New setup UX finding: the end-of-setup restart reminder is still Claude-specific after adding Pi to the setup flow.

`CodingAgentsStep` can now install only the Pi extension, and `AnyHooksInstalled` correctly includes `PiExtensionInstalled`, so `SetupCommand.LiveRecordingRestartTip` will run for that case. But the text still says `Restart your agent (or run claude --continue)`, which is misleading for Pi, Cursor, and Copilot installs, and especially odd when the only installed integration was Pi. Please make the reminder generic or build it from the installed agents; for Pi it should tell users to restart Pi so the extension is loaded, matching the per-agent note printed during the Pi install step.

realtonyyoung:
Additional docs/help drift from the exhaustive sweep, separate from the command-specific help resources already noted:

The root README still has several old “four agents” / Claude-Codex-only sections after the targeted edits: the npm install-script warning and manual refresh command omit `--pi`; setup step 4 omits Pi; the restart note still suggests only `claude --continue`; the “Need at least one agent” / “What it records” / import prose still describes only the older agent set in places; and the upgrade/uninstall/finer-grained removal sections omit the Pi extension. `npm/kcap/README.md` is also still Claude-only, and `help-usage.txt` still advertises import/plugin/hook support without Pi (and plugin help there is missing Cursor/Copilot/skills too).

Please do one docs sweep across the root README, npm package README, and `help-usage.txt` so the published/package help matches the new Pi support and the existing Cursor/Copilot surfaces.

realtonyyoung:
`kcap status` still only reports Claude and Codex hook state, so a user who installs the new Pi extension has no status feedback that Pi recording is installed. `StatusCommand.HandleAsync` builds the Hooks line from only `IsClaudePluginInstalled(ClaudePaths.UserSettings)` and `IsCodexHooksInstalled(CodexPaths.UserHooksJson)`, and `help-status.txt` documents the same older two-agent view. Please extend status/help to include the new Pi extension state (and ideally the already-supported Cursor/Copilot states too) so setup/plugin install can be verified from `kcap status`.

realtonyyoung:
Coverage gap worth closing before this lands: the new Pi support has helper-level tests (`PiExtensionInstallerTests`, discovery/import-relevance in `PiImportSourceTests`, and Pi title parsing), but I could not find tests that exercise the user-facing Pi branches. In particular, `PluginCommand.HandleAsync(["plugin", "install", "--pi" ...])` / `remove --pi` / `--if-installed`, `CodingAgentsStep` setup/no-prompt behavior for detected Pi, and `PiImportSource.ClassifyAsync` + `ImportSessionAsync` posting `/hooks/session-start/pi`, the transcript payload with vendor `pi`, and `/hooks/session-end/pi` appear uncovered. Those are the integration points most likely to drift from the lower-level helpers, so please add focused tests around the command/routed-import flows rather than only the file writer and line parser.

realtonyyoung:
All four broader-pass findings addressed in `79dbf7a`:

**Lifecycle callers (uninstall + refresh).** `UninstallCommand` now calls `plugin remove --pi` (and prints the `~/.pi/agent/extensions/kcap.ts` removal line), and `refresh.js` now runs `["plugin","install","--pi","--if-installed"]`. `InstallPi` already honours `--if-installed` (no-ops unless a marker/extension is present), so the npm postinstall / `kcap update` refresh path never force-installs Pi onto users who never opted in. Tests: `PluginCommandPiTests` (`--if-installed` no-op when not installed, refresh-in-place, remove deletes extension+marker) and `UninstallCommandTests.User_level_uninstall_removes_pi_extension` (removes kcap.ts + marker, preserves a user-authored sibling extension).

**Routed-import title contract.** Set `PiImportSource.SupportsTitleGeneration => false`. You're right that a routed source (`FilePath=""`) never reaches the chain worker's `OnTitleTaskReady`, and `GenerateTitleForImportAsync` is Claude-shaped anyway — so I took option 1: imported Pi sessions get the server-side fallback title from `PiHookHandlers` (matching Copilot/Cursor), while **live** Pi still gets an LLM title via the watcher's Pi-aware extractors. Updated the docstring to say so; added `does_not_support_title_generation`.

**Setup restart reminder.** `LiveRecordingRestartTip` now builds the hint from the installed agents: Claude keeps `claude --continue`; a Pi-only install reads "Restart `pi` so the kcap extension loads"; anything else gets a generic "Restart your agent". Added `LiveRecordingRestartTip_pi_only_tells_user_to_restart_pi` (asserts it mentions pi and not `claude --continue`).

**Embedded help resources.** Updated `help-import.txt` (Pi in the agent list, transcript path, `--pi` filter, "all five"), `help-plugin.txt` (`--pi` option + notes + examples), `help-setup.txt` (`--skip-pi-hooks`, Pi detection, install example), `help-hook.txt` (arg-driven Pi dispatch, since Pi has no stdin payload), and also `help-usage.txt` (top-level import line + hook list) which had the same gap.

No JS test harness exists in the repo (no `*.test.js`, no `package.json` test script), so the refresh.js change is covered indirectly via the C# `InstallPi --if-installed` no-op test rather than a new JS suite.

1573/1573 unit tests pass; AOT publish clean (0 IL warnings).

realtonyyoung:
Both addressed in `d61719d`:

**`kcap status` now reports Pi.** The Hooks line was Claude+Codex only; it now reports all five agents — `Claude ✓  Codex ✗  Cursor ✗  Copilot ✗  Pi ✓` — using the existing per-agent detectors (`CursorHooksInstaller.IsInstalled`, `CopilotHooksInstaller.IsInstalled`, `PiExtensionInstaller.IsInstalled`). I extracted the line formatting into a pure `StatusCommand.BuildHooksStatusLine(...)` so it's unit-testable without the HTTP/daemon/global-state in `HandleAsync`, and added two tests for it. `help-status.txt` updated to describe the per-agent state.

**Command + routed-import coverage gap closed.** Added the user-facing branch tests that were missing:
- **`CodingAgentsStepTests`** — Pi detected/accepted, declined, not-detected, skipped-by-flag, **`--no-prompt` auto-install**, kcap-not-on-PATH abort, and installer-failure warning. Mirrors the Copilot/Cursor branch coverage.
- **`PiImportSourceImportTests`** (integration, WireMock) — drives the full `DiscoverAsync → ClassifyAsync → ImportSessionAsync` path and asserts the wire contract: `POST /hooks/session-start/pi`, then the transcript batch carrying `"vendor":"pi"`, then `POST /hooks/session-end/pi` with `reason: pi-import` — in receipt order. A second case covers the Partial/resume-from-watermark path (server reports `last_line_number: 0` → resume from line 1 → `Resumed`).
- The `plugin install/remove --pi` / `--if-installed` branches you noted were covered by `PluginCommandPiTests` in `79dbf7a`.

1582/1582 unit + 2 new integration tests pass; AOT publish clean.

realtonyyoung:
Remaining docs/help drift on the current head after the Pi follow-ups:

- `src/Capacitor.Cli.Core/Resources/help-usage.txt` updated import + hooks for Pi, but the top-level **Plugin** section still advertises only `plugin install/remove [--project] [--codex]` and says “Claude or --codex for Codex”. Since this is what `kcap --help` shows, users discovering integrations there still cannot see `--cursor`, `--copilot`, `--pi`, or `--skills`. Please broaden those two lines to match `help-plugin.txt`.
- `README.md` still has several older-agent passages despite nearby Pi edits: the install-script warning/manual refresh list omits `--pi`, setup step 4 omits Pi detection, the historical import section still says “All four agents” and omits the Pi session root, the upgrade paragraph omits Cursor/Copilot/Pi refresh targets, and uninstall/finer-grained removal still omits Copilot/Pi.
- `npm/kcap/README.md` is still Claude-only (“records and visualizes Claude Code sessions”, setup walks through Claude plugin installation) even though the package now installs/refreshes multi-agent integrations including Pi.

This is user-visible packaging/help text, so it is worth one final docs sweep before merging; otherwise the shipped CLI/npm docs understate the new Pi support and some already-supported Cursor/Copilot paths.

realtonyyoung:
Done in `9674e84` (the help-usage Plugin lines were already broadened in the `origin/main` merge `38c6046`):

- **`help-usage.txt` Plugin section** — now `plugin install/remove [--project] [--codex|--cursor|--copilot|--pi|--skills]` with "Register/Remove hooks/skills (Claude default; flag picks the vendor)" (landed via the #163 merge + the `--pi` add).
- **`README.md`** — postinstall refresh list, the manual `--if-installed` example, setup step 4 detection, both import backfill source lists ("all four" → "all five" + `~/.pi/agent/sessions/`), the uninstall coverage paragraph (+ Copilot hooks, + the Pi `kcap.ts` extension), and the `plugin remove` finer-grained-cleanup flags now all include Pi (and the Cursor/Copilot paths that were missing).
- **`npm/kcap/README.md`** — no longer Claude-only: the tagline and setup step now describe multi-agent recording across Claude/Codex/Cursor/Copilot/Pi.

Verified no residual `all four` / Claude-only / `--pi`-less plugin lists remain in either README.

realtonyyoung:
Fresh re-review findings on current head (`7daa29b`), after the Gemini/main merge:

- `kcap status` now says it renders every supported agent, but the merged branch supports Gemini and `StatusCommand.BuildHooksStatusLine` still only accepts/prints Claude, Codex, Cursor, Copilot, and Pi. `help-status.txt` has the same omission. A user who installs `kcap plugin install --gemini` cannot verify it from `kcap status`, while all the other current vendors can. Please add Gemini to the status detector/line/helper tests (using `GeminiHooksInstaller.IsInstalled(GeminiPaths.SettingsJson())`) and update the help text from “Codex / Cursor / Copilot hooks, and Pi...” to include Gemini hooks.
- `npm/kcap/bin/refresh.js` still does not run `plugin install --gemini --if-installed`, even though this branch’s README says npm postinstall / `kcap update` refreshes Claude/Codex/Cursor/Copilot/Gemini/Pi. That means opted-in Gemini users keep stale `~/.gemini/settings.json` hook command strings after upgrade, while the docs promise refresh. Please add the Gemini refresh entry and update the header comment accordingly.
- `src/Capacitor.Cli.Core/Resources/help-uninstall.txt` mentions Pi only in the finer-grained `plugin remove [...]` note, but the main description still says uninstall removes Claude/Codex/Cursor/Gemini hooks, Copilot hooks, skills, and config. Since `UninstallCommand` now also removes `~/.pi/agent/extensions/kcap.ts`, the command-specific help should say that in the primary removal list too.
- `npm/kcap/README.md` was updated from Claude-only to multi-agent, but it lists Claude/Codex/Cursor/Copilot/Pi and still omits Gemini in both the tagline and setup sentence. Since the current package includes Gemini hooks/import/setup, please include Gemini there as well.

realtonyyoung:
New functional finding from the latest re-review:

`GeminiHookCommand.EnsureWatcher` starts the shared watcher with `vendor: "gemini"` and only skips titles on `source` resume/clear, so a fresh live Gemini session is expected to use the watcher title path. But `WatchCommand` has no Gemini branches in `TryExtractUserText`, `TryExtractAssistantText`, or `IsEvent`: it handles Codex/Copilot/Pi specially, then falls back to the Claude shape. Gemini transcript lines are shaped like `{"type":"user","content":[{"text":"..."}]}` and `{"type":"gemini","content":"..."}` (as used by `GeminiImportSource.IsImportRelevantLine`), while the Claude fallback looks for user text under `message.content` and only counts `type:"user"|"assistant"` events.

Net effect: live Gemini sessions still stream transcript lines, but they never capture `FirstUserText`, never send the initial title, and never reach the LLM title threshold because assistant turns are `type:"gemini"`, not `type:"assistant"`. Please add Gemini-specific watcher title/event helpers mirroring the Gemini normalizer/import predicate (skip `<session_context>` user bootstrap and use `type:"gemini"` as assistant context), plus `WatchCommand` tests for Gemini user/gemini lines.

PR #164, opened 2026-06-17
[AI-887] Google Gemini CLI ingest (CLI side)

realtonyyoung:
## [AI-887] Google Gemini CLI ingest — CLI side

Companion to the server PR (kurrent-io/kcap-server#759). Adds Google Gemini CLI as the 6th ingest vendor, mirroring the Copilot pattern. **Must merge after the server PR deploys** (this PR's hooks POST to `/hooks/session-{start,end}/gemini` + `/hooks/transcript?vendor=gemini`, which the server PR adds).

### Done in this PR (builds clean)
- `Core/Gemini/GeminiPaths` — `~/.gemini` layout (`$GEMINI_HOME` honoured; `settings.json`; `tmp/<project>/chats/`).
- `Core/Gemini/GeminiHooksParser` + `GeminiHooksInstaller` — helpers for the SHARED `~/.gemini/settings.json` `hooks` block (nested `{hooks:[{command}]}` shape; merge/detect/marker — must preserve user-authored hooks, like the Cursor installer).
- `Commands/GeminiHookCommand` — self-routes on `hook_event_name` (Gemini provides it, unlike Copilot); POSTs lifecycle to `/hooks/session-{start,end}/gemini`; spawns the watcher on the payload's `transcript_path` with `vendor=gemini`; capped inline drain on `SessionEnd`; `Notification` → `/hooks/notification`.
- Wiring: `kcap hook --gemini` dispatch + `KCAP_SKIP` gate + usage in `Program.cs`; `gemini` added to `WatchCommand.KnownVendors`.

### Follow-ups (this branch, before marking ready)
- `PluginCommand` `InstallGeminiHooks`/`RemoveGeminiHooks` (settings.json merge) + `--gemini` / `--skip-gemini-hooks` dispatch.
- `SetupCommand` / `CodingAgentsStep` Gemini detection + install offer.
- `GeminiImportSource : IImportSource` (`kcap import --gemini`, walking `~/.gemini/tmp/*/chats/*.jsonl`) + `VendorSelection` `--gemini`.
- Unit tests (hooks-merge + import discovery), AOT publish check, and `README.md` (new `--gemini` surface).

Spec + design: AI-887 (and server PR #759).

### DIFF
--- a/src/Capacitor.Cli/Commands/PluginCommand.cs
+++ b/src/Capacitor.Cli/Commands/PluginCommand.cs
@@ -943,7 +943,9 @@ public static class PluginCommand {
                     }
                 }
 
-                preserved.Add(GeminiHooksParser.BuildKcapEntry());
+                // Cast to JsonNode so this binds to JsonArray.Add(JsonNode?), not
+                // the generic Add<T>(T) — the generic trips IL2026/IL3050 under AOT.
+                preserved.Add((JsonNode)GeminiHooksParser.BuildKcapEntry());
                 hooks[evt] = preserved;
             }
 
--- a/src/Capacitor.Cli/Commands/PluginCommand.cs
+++ b/src/Capacitor.Cli/Commands/PluginCommand.cs
@@ -3,6 +3,7 @@ using System.Text.Json.Nodes;
 using Capacitor.Cli.Core;
 using Capacitor.Cli.Core.Copilot;
 using Capacitor.Cli.Core.Cursor;
+using Capacitor.Cli.Core.Gemini;
 
 namespace Capacitor.Cli.Commands;
 
@@ -35,7 +36,7 @@ public static class PluginCommand {
         };
     }
 
-    static readonly string[] ExclusiveTargetFlags = ["--codex", "--cursor", "--copilot", "--skills"];
+    static readonly string[] ExclusiveTargetFlags = ["--codex", "--cursor", "--copilot", "--gemini", "--skills"];
 
     static bool HasConflictingTargets(string[] args) =>
         ExclusiveTargetFlags.Count(args.Contains) > 1;
@@ -43,7 +44,7 @@ public static class PluginCommand {
     static async Task<int> Install(string[] args, PluginEnvironment env) {
         if (HasConflictingTargets(args)) {
             await env.Stderr.WriteLineAsync(
-                "--cursor, --codex, --copilot, and --skills are mutually exclusive."
+                "--cursor, --codex, --copilot, --gemini, and --skills are mutually exclusive."
             );
 
             return 1;
@@ -53,6 +54,7 @@ public static class PluginCommand {
         if (args.Contains("--codex")) return await InstallCodex(args, env);
         if (args.Contains("--cursor")) return await InstallCursor(args, env);
         if (args.Contains("--copilot")) return await InstallCopilot(args, env);
+        if (args.Contains("--gemini")) return await InstallGemini(args, env);
 
         return await InstallClaude(args, env);
     }
@@ -60,7 +62,7 @@ public static class PluginCommand {
     static async Task<int> Remove(string[] args, PluginEnvironment env) {
         if (HasConflictingTargets(args)) {
             await env.Stderr.WriteLineAsync(
-                "--cursor, --codex, --copilot, and --skills are mutually exclusive."
+                "--cursor, --codex, --copilot, --gemini, and --skills are mutually exclusive."
             );
 
             return 1;
@@ -70,6 +72,7 @@ public static class PluginCommand {
         if (args.Contains("--codex")) return await RemoveCodex(args, env);
         if (args.Contains("--cursor")) return await RemoveCursor(args, env);
         if (args.Contains("--copilot")) return await RemoveCopilot(args, env);
+        if (args.Contains("--gemini")) return await RemoveGemini(args, env);
 
         return await RemoveClaude(args, env);
     }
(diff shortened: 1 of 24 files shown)


src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -835,6 +838,158 @@ public static bool RemoveCursorHooks(string hooksPath) {
...
+    /// </summary>
+    public static bool InstallGeminiHooks(string settingsPath) {
+        try {
+            JsonObject root = [];
+
+            if (File.Exists(settingsPath)) {
+                try {
+                    if (JsonNode.Parse(File.ReadAllText(settingsPath)) is JsonObject obj) root = obj;
+                } catch {
+                    /* Malformed — start fresh */
```
realtonyyoung: This is risky for Gemini because `settings.json` is shared user config, not a kcap-owned hook file. If the existing file is malformed or temporarily half-written, this catch falls through with `root = []` and the later `File.WriteAllText` replaces the whole file with only kcap hooks, dropping unrelated Gemini settings. For a shared config merge, parse failure should fail the install/refresh and leave the file untouched (ideally with an error telling the user to fix `settings.json`), rather than 'start fresh'. A small regression test with malformed JSON would lock this down.

realtonyyoung: Fixed in 400ab72. `InstallGeminiHooks` now fails closed: when `settings.json` exists but won't parse into a JSON object (malformed/empty/half-written) it returns false and leaves the file untouched, instead of starting fresh and overwriting it — matching the `RemoveClaudePlugin` 'Malformed' precedent. The install command + setup wizard now tell the user to make settings.json valid JSON and that it was left untouched. Cursor/Copilot are unaffected (they own their dedicated hooks file). Regression: `install_leaves_malformed_settings_untouched` + `install_leaves_non_object_settings_untouched` assert false + a byte-for-byte unchanged file.

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -35,15 +36,15 @@ public static async Task<int> HandleAsync(string[] args, PluginEnvironment? env
         };
     }
 
-    static readonly string[] ExclusiveTargetFlags = ["--codex", "--cursor", "--copilot", "--skills"];
+    static readonly string[] ExclusiveTargetFlags = ["--codex", "--cursor", "--copilot", "--gemini", "--skills"];
```
realtonyyoung, line 39: Adding `--gemini` here makes the command work, but the embedded CLI help still does not advertise it. Please update the resource help text alongside this flag: `src/Capacitor.Cli.Core/Resources/help-plugin.txt` has install/remove/refresh examples but no Gemini entries; `help-hook.txt` lists supported hook vendors without Gemini; `help-setup.txt` says setup detects Claude/Codex/Cursor/Copilot only and shows install examples without `--gemini`; `help-import.txt` lists only Claude/Codex/Cursor/Copilot transcript roots and vendor filters; and `help-usage.txt` still describes import/plugin/hook support without Gemini. Otherwise users running `kcap help`, `kcap import --help`, `kcap plugin --help`, `kcap hook --help`, or `kcap setup --help` will not discover the new Gemini support even though README mentions it.

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -70,6 +72,7 @@ await env.Stderr.WriteLineAsync(
         if (args.Contains("--codex")) return await RemoveCodex(args, env);
         if (args.Contains("--cursor")) return await RemoveCursor(args, env);
         if (args.Contains("--copilot")) return await RemoveCopilot(args, env);
+        if (args.Contains("--gemini")) return await RemoveGemini(args, env);
```
realtonyyoung, line 75: Now that `plugin remove --gemini` exists, `kcap uninstall` needs to invoke it as part of the all-agent cleanup. `UninstallCommand.HandleAsync` currently removes Claude, Codex, Cursor, Copilot, and skills, and then deletes only the Claude/Codex/Cursor markers; it never calls `PluginCommand.HandleAsync(["plugin", "remove", "--gemini"])` or deletes the Gemini marker. A user who installed Gemini hooks and then runs `kcap uninstall` would still have `kcap hook --gemini` merged into `~/.gemini/settings.json`, so Gemini would keep launching a removed/unconfigured kcap command after uninstall. Please wire Gemini into uninstall and update `help-uninstall.txt` at the same time.

src/Capacitor.Cli/Commands/PluginCommand.cs
realtonyyoung: Fixed in 44fc402. Added Gemini to all the embedded help you listed: help-plugin.txt (`--gemini` option + notes + install/remove examples), help-hook.txt (`--gemini` dispatch + the install-flag list), help-setup.txt (`--skip-gemini-hooks` + the detection paragraph now names Gemini + the `kcap plugin install` example includes `--gemini`), help-import.txt (`~/.gemini/tmp/.../chats` transcript root + `--gemini` vendor filter + "all five"), and help-usage.txt (import summary + a `hook --gemini` line in the hooks section). So `kcap help`, `plugin --help`, `hook --help`, `setup --help`, and `import --help` all surface Gemini now.

src/Capacitor.Cli/Commands/PluginCommand.cs
realtonyyoung: Fixed in 44fc402. `UninstallCommand.HandleAsync` now runs `plugin remove --gemini` alongside the other agents and deletes the Gemini `.kcap-hooks-version` marker (belt-and-braces, same as Cursor), and prints the Gemini removal line; help-uninstall.txt documents it (and `plugin remove` now lists `--gemini`). So a user who installed Gemini hooks no longer keeps `kcap hook --gemini` merged in ~/.gemini/settings.json after uninstall. Regression: the comprehensive uninstall test now seeds a Gemini settings.json (kcap hook + user hook + unrelated key + marker) and asserts the kcap hook is gone, the user hook + key survive, and `GeminiHooksInstaller.IsInstalled` is false.

PR #165, opened 2026-06-17
[AI-888] AWS Kiro CLI ingest (hooks, JSONL watcher, import, transparent live install)

realtonyyoung:
## What

Phase 2 (CLI) for **AWS Kiro CLI** ingest — the fifth vendor (`kiro`). Pairs with the server side, [kcap-server#760](https://github.com/kurrent-io/kcap-server/pull/760).

Linear: [AI-888](https://linear.app/kurrent/issue/AI-888/aws-kiro-cli-session-ingest)

## What Kiro needs that's new

Kiro CLI (AWS's terminal coding agent; a bun/TUI rewrite, not the older Rust Amazon Q Developer CLI) writes an append-only **JSONL** log per session — `~/.kiro/sessions/cli/{id}.jsonl` (one `{version, kind, data}` object per line; honours `KIRO_HOME`), with a sibling `{id}.json` for cwd/title/timestamps/model. So it tails like the other file-vendors; what's actually new is the **lifecycle**: there's no session-start line in the log, so start metadata comes from Kiro's `agentSpawn` hook (fires every prompt → deduped server-side), and Kiro has **no session-end trigger**, so the watcher synthesizes it on `kiro-cli` exit. (Kiro *also* writes a `data.sqlite3`, but the JSONL is written for every session, so kcap ignores the DB.)

## Changes

- **`kcap hook --kiro`** (`KiroHookCommand`) — `agentSpawn` dispatcher: POSTs `/hooks/session-start/kiro` (enriched with the model read from the sibling `{id}.json`), ensures the watcher, always exits 0 with empty stdout (a stdout-writing Kiro hook re-injects and loops the agent). `agentSpawn`-only — it carries the session id and fires every prompt, covering session start + idempotent watcher liveness.
- **JSONL watcher** — the watcher tails `~/.kiro/sessions/cli/{id}.jsonl` directly through the existing file-tail drain (`WatchCommand`), streaming its lines to `/hooks/transcript?vendor=kiro` — so all the title/threshold/resume/parent-exit machinery is reused unchanged, with **no DB read, copy step, or flatten**. Session-end is synthesized via the existing parent-exit path (now that `kiro` is in `KnownVendors`).
- **`kcap import --kiro`** (`KiroImportSource`) — walks `~/.kiro/sessions/cli`, replaying historical sessions through session-start → transcript → session-end. Idempotent (deterministic server-side event ids).
- **Installer + setup** — `kcap plugin install/remove --kiro` writes `~/.kiro/agents/kcap.json`; the setup wizard detects Kiro (`~/.kiro/` or `kiro`/`kiro-cli` on PATH) and offers install. `--skip-kiro-hooks` opts out.
- **Wiring** — `--kiro` in `VendorSelection`, `Program.cs` hook/import dispatch + `KCAP_SKIP`, watcher `KnownVendors` + Kiro title extractors.

## No new dependencies

Kiro is pure JSONL tailing, so it reuses the existing file-tail watcher/drain — **no `Microsoft.Data.Sqlite` or other native dependency was added**, and nothing new to vet under NativeAOT publish.

## Tests

- 21 new unit tests across `KiroHooksTests`, `KiroImportSourceTests`, `KiroSettingsTests`, and `KiroUsageTests`, plus the Kiro setup branch of `CodingAgentsStepTests`. Full unit suite green (the one unrelated `GetCodingAgentPid` flake passes in isolation).
- 1 new integration test: `WatcherParentExitPostTests.ParentExit_PostsToVendorSpecificRoute_ForKiro` — the synthesized session-end routes to `/hooks/session-end/kiro`.
- Docs: `README.md` + `help-{hook,import,plugin,setup}.txt` updated.

## Notes / follow-ups

- The `~/.kiro/agents/kcap.json` owned-agent model assumes Kiro applies `agents/*.json` hooks globally; final confirmation belongs to a manual E2E against a live `kiro-cli` (the JSONL line schema is officially undocumented — kirodotdev/Kiro #5094 — so it's treated as version-fragile/fail-open throughout).
- Kiro persists no token counts, so Kiro sessions show no token usage by design.


src/Capacitor.Cli/Commands/KiroHookCommand.cs
```diff
@@ -0,0 +1,202 @@
...
+        var enriched = await RepositoryDetection.EnrichWithRepositoryInfo(forwarded.ToJsonString());
+
+        if (activeProfile?.ExcludedRepos is { Length: > 0 } excludedRepos
+         && await RepoExclusion.IsExcludedAsync(enriched, excludedRepos)) {
+            DisabledSessions.Mark(sessionId);
+            return 0;
+        }
+
+        var exit = await PostHookAsync(baseUrl, "session-start/kiro", enriched);
+        if (exit != 0) return exit;
```
realtonyyoung: The Kiro hook contract in this PR says the dispatcher must always exit 0 with empty stdout, but this path returns `1` whenever `/hooks/session-start/kiro` is temporarily unreachable or returns a non-2xx. That can surface as a failed `agentSpawn` hook inside `kiro-cli` for a server/auth blip, which is exactly the failure mode the Kiro-specific wrapper is trying to avoid. For this vendor, please log the failure to stderr but still return 0 (and decide separately whether `EnsureWatcherRunning` should still be attempted after a failed start POST).

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -798,6 +803,198 @@ public static bool RemoveCopilotHooks(string hooksPath) {
...
+            // overwrite it with "kcap").
+            var recordedDefault = alreadyKcap
+                ? KiroHooksInstaller.ReadPreviousDefault(agentJsonPath) ?? "kiro_default"
+                : currentDefault;
+
+            // Clone the current default into the kcap agent (kiro-cli writes it to
+            // the global agents dir, preserving tools/prompt). Skipped if kcap exists.
+            if (!File.Exists(agentJsonPath)) {
+                if (!AgentDetector.IsInstalled(KiroBinary)) return false;
+                if (RunKiroCli($"agent create {KiroAgentName} --from {recordedDefault}") != 0 || !File.Exists(agentJsonPath))
```
realtonyyoung: This builds the `kiro-cli` argument string by concatenating the saved default-agent name. `ProcessStartInfo(fileName, arguments)` does not preserve argument boundaries, so a valid Kiro agent name containing whitespace (or quotes) is split before it reaches `kiro-cli` (`--from My Agent` becomes two trailing args). That makes transparent install fail for users whose current default agent is not a single shell token. Please switch this helper to `ProcessStartInfo.ArgumentList` (for example `agent`, `create`, `kcap`, `--from`, `recordedDefault`) so the recorded default is passed as one argument.

src/Capacitor.Cli/Commands/KiroImportSource.cs
```diff
@@ -0,0 +1,476 @@
...
+        var result = new List<DiscoveredSession>();
+
+        if (!Directory.Exists(_sessionsDir))
+            return Task.FromResult<IReadOnlyList<DiscoveredSession>>(result);
+
+        foreach (var jsonl in Directory.EnumerateFiles(_sessionsDir, "*.jsonl")) {
+            ct.ThrowIfCancellationRequested();
+
+            // Filename stem is the dashed session UUID Kiro uses for both files.
+            var dashed   = Path.GetFileNameWithoutExtension(jsonl);
```
realtonyyoung: Kiro session files are UUID-named, and the live hook path already rejects non-GUID session ids. The importer should apply the same `Guid.TryParse` guard here before deriving `dashless`; otherwise any stray `*.jsonl` in this directory (backup/export/debug file, partially renamed file, etc.) is discovered as a session and then starts hitting `/api/sessions/{name}/last-line` / lifecycle routes with a non-session id. That will at best show probe/import errors for files that are not Kiro sessions, and at worst create inconsistent server state if the route accepts the string. Please skip stems that are not valid UUIDs.

src/Capacitor.Cli/Commands/KiroHookCommand.cs
realtonyyoung: Fixed in a69453e. `HandleAgentSpawn` now returns `0` when the session-start POST fails (`PostHookAsync` already logs the failure to stderr), so a transient server/auth blip no longer surfaces as a failed `agentSpawn` hook inside `kiro-cli`. On a failed POST it also skips `EnsureWatcherRunning` for that firing — `agentSpawn` fires again on the next prompt and retries both the POST and the watcher.

src/Capacitor.Cli/Commands/PluginCommand.cs
realtonyyoung: Fixed in a69453e. `RunKiroCli` now takes `params string[]` and uses `ProcessStartInfo.ArgumentList` (`"agent", "create", "kcap", "--from", recordedDefault`), so a default-agent name containing whitespace or quotes reaches `kiro-cli` as a single argument.

src/Capacitor.Cli/Commands/KiroImportSource.cs
realtonyyoung: Fixed in a69453e. `DiscoverAsync` now applies `Guid.TryParse` to the filename stem and skips any non-UUID `*.jsonl` (backup/export/debug/partial-rename) before deriving `dashless` or probing `/api/sessions/{id}/...`, mirroring the live hook path.

src/Capacitor.Cli.Core/Kiro/KiroSettings.cs
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /// <summary>
+    /// Sets <c>chat.defaultAgent</c> to <paramref name="agentName"/>, preserving
+    /// every other key, creating the file/dir if needed. Returns false on I/O
+    /// failure (caller surfaces it).
+    /// </summary>
+    public static bool SetDefaultAgent(string settingsPath, string agentName) {
+        try {
+            var root = (File.Exists(settingsPath) ? JsonNode.Parse(File.ReadAllText(settingsPath)) : null) as JsonObject
```
realtonyyoung: This is Kiro’s shared settings file, but when it already exists and is malformed or is not a JSON object this expression falls back to `new JsonObject()` and `File.WriteAllText` replaces the whole file with only `chat.defaultAgent`. That can silently drop unrelated Kiro settings (`chat.defaultModel`, etc.) during install/remove. The Gemini installer handles the same shared-settings case by failing closed on malformed JSON; this should do the same here: if the file exists but cannot be parsed as a `JsonObject`, return `false` and leave it untouched.

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -798,6 +803,202 @@ public static bool RemoveCopilotHooks(string hooksPath) {
...
+
+    static async Task<int> RemoveKiro(string[] args, PluginEnvironment env) {
+        var agentPath    = GetArg(args, "--kiro-agent-path")    ?? env.KiroKcapAgentJson;
+        var settingsPath = GetArg(args, "--kiro-settings-path") ?? KiroSettingsPathFor(agentPath);
+
+        try {
+            // Restore the default agent kcap replaced (recorded at install time).
+            var previousDefault = KiroHooksInstaller.ReadPreviousDefault(agentPath) ?? "kiro_default";
+            if (KiroSettings.ReadDefaultAgent(settingsPath) == KiroAgentName)
+                KiroSettings.SetDefaultAgent(settingsPath, previousDefault);
```
realtonyyoung: `SetDefaultAgent` returns `false` on I/O/parse failure, but removal ignores that and then deletes `~/.kiro/agents/kcap.json` plus the marker. If the settings write fails here, the user is left with `chat.defaultAgent` still set to `kcap` while the `kcap` agent file has been removed, and the recorded previous default is gone so a retry cannot restore it. Please fail removal before deleting the agent/marker when the default is currently `kcap` and restoring `previousDefault` fails (and report the settings path/error to the user).

src/Capacitor.Cli.Core/Kiro/KiroSettings.cs
realtonyyoung: Fixed in e3aeeba. `SetDefaultAgent` now fails closed: if the settings file exists but does not parse as a `JsonObject` (a JSON array/scalar, or malformed JSON which throws) it returns `false` and leaves the file untouched, instead of replacing it with a fresh `{chat.defaultAgent}` and dropping `chat.defaultModel` etc. Mirrors the Gemini installer. Added `set_default_fails_closed_on_valid_json_that_is_not_an_object` and `set_default_fails_closed_on_malformed_file`.

src/Capacitor.Cli/Commands/PluginCommand.cs
realtonyyoung: Fixed in e3aeeba. `RemoveKiro` now checks `SetDefaultAgent`'s return: when kcap is the current default and restoring `previousDefault` fails, it aborts BEFORE deleting `kcap.json`/the marker, prints the settings path + retry hint to stderr, and returns 1. The agent file and the recorded previous default stay in place so `kcap plugin remove --kiro` is retryable after the settings file is fixed.

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -798,6 +803,215 @@ public static bool RemoveCopilotHooks(string hooksPath) {
...
+            // the global agents dir, preserving tools/prompt). Skipped if kcap exists.
+            if (!File.Exists(agentJsonPath)) {
+                if (!AgentDetector.IsInstalled(KiroBinary)) return false;
+                if (RunKiroCli("agent", "create", KiroAgentName, "--from", recordedDefault) != 0 || !File.Exists(agentJsonPath))
+                    return false;
+            }
+
+            if (!InjectKiroHooksIntoAgent(agentJsonPath)) return false;
+
+            KiroHooksInstaller.WriteMarker(agentJsonPath, recordedDefault);
```
realtonyyoung: `WriteMarker` happens before the final `SetDefaultAgent` write. If the settings flip fails (for example the existing shared settings file is malformed and `SetDefaultAgent` now correctly returns `false`), `InstallKiroHooks` reports failure but leaves `.kcap-hooks-version` behind. That makes the failed install look installed to `IsInstalled`/`--if-installed` refreshes, even though `chat.defaultAgent` was never changed to `kcap`. Please only write the marker after `SetDefaultAgent` succeeds, or delete it on this failure path so failed installs do not poison future refresh/detection state.

realtonyyoung: Fixed in 7674626. `InstallKiroHooks` now flips the default (`SetDefaultAgent`) FIRST and stamps `.kcap-hooks-version` only after it succeeds — a failed settings flip returns `false` with no marker left behind, so `IsInstalled`/`--if-installed` no longer treat the broken install as done and the next `--if-installed` refresh retries it.

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -798,6 +803,223 @@ public static bool RemoveCopilotHooks(string hooksPath) {
...
+
+            // Flip the default FIRST; only stamp the marker once it succeeds.
+            // Otherwise a failed settings flip (e.g. a malformed shared settings
+            // file, which SetDefaultAgent now fails closed on) would leave a marker
+            // that makes IsInstalled / --if-installed treat the broken install as
+            // done — so the next refresh would skip it and kcap would never become
+            // the default. With no marker, the next --if-installed refresh retries.
+            if (!KiroSettings.SetDefaultAgent(settingsPath, KiroAgentName)) return false;
+
+            KiroHooksInstaller.WriteMarker(agentJsonPath, recordedDefault);
```
realtonyyoung: The ordering is fixed now, but `WriteMarker` still swallows all write failures while `InstallKiroHooks` returns success. That marker is not just a refresh optimization for Kiro: line 2 is the only record of the user’s previous default agent. If writing `.kcap-hooks-version` fails after `SetDefaultAgent` succeeds, `plugin install --kiro` reports success with `chat.defaultAgent=kcap` but `plugin remove --kiro` later falls back to `kiro_default` and can restore the wrong agent; a subsequent `--if-installed` refresh also sees the hook file and can re-stamp the wrong previous default. Please make the marker write part of the atomic install contract (return `false`/throw from `WriteMarker`, and either fail before reporting success or roll the default back if the marker cannot be written).

realtonyyoung: Fixed in c8d3e5a. `WriteMarker` now returns `bool` (false on I/O failure instead of swallowing), and `InstallKiroHooks` treats it as part of the atomic install: if the marker write fails after the default flip, it rolls `chat.defaultAgent` back to the recorded previous default and returns failure — so install never reports success in a state where `remove --kiro` would restore the wrong agent or `--if-installed` could re-stamp a bogus previous default.

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -861,6 +866,231 @@ public static bool RemoveCopilotHooks(string hooksPath) {
...
+            // the default. With no marker, the next --if-installed refresh retries.
+            if (!KiroSettings.SetDefaultAgent(settingsPath, KiroAgentName)) return false;
+
+            // The marker's line 2 is the ONLY record of the replaced default, so a
+            // silent failure here would let `remove --kiro` restore the wrong agent
+            // (and a later --if-installed refresh re-stamp a bogus previous default).
+            // Treat it as part of the atomic install: on failure roll the default
+            // back and report failure rather than a success we can't undo.
+            if (!KiroHooksInstaller.WriteMarker(agentJsonPath, recordedDefault)) {
+                KiroSettings.SetDefaultAgent(settingsPath, recordedDefault);
```
realtonyyoung: This rollback is still best-effort because the return value is ignored. If the marker write fails and the settings rollback also fails (for example the shared settings file is replaced, locked, or becomes malformed between the two writes), `InstallKiroHooks` returns failure but leaves `chat.defaultAgent=kcap` with no marker recording the previous default. A later `remove --kiro` then falls back to `kiro_default`, so the user can still lose their real prior default. Please check the rollback result and, on rollback failure, either leave enough recovery state for `remove --kiro` to restore the recorded default or surface a distinct failure path instead of silently returning the same false state.

realtonyyoung: Fixed in 7607018. The rollback now checks its own result: if `SetDefaultAgent(recordedDefault)` also fails, `InstallKiroHooks` writes a distinct, actionable message to stderr naming the previous default agent (`recordedDefault`) plus the settings + agent-file paths, so the user can restore `chat.defaultAgent` by hand — instead of returning the same opaque `false` as "kiro-cli not installed" and silently losing the prior default. (A richer return type would ripple through the `Func<string,bool>` wiring in SetupCommand/CodingAgentsStep, so the recovery info is surfaced via stderr on this rare double-failure path.)

PR #168, opened 2026-06-18
[AI-897] Deliver Copilot post-sessionEnd session.shutdown tail (live finalize drain)

realtonyyoung:
## Problem

In the **live hook path**, Copilot session token stats show output-only — input/cache stay 0. Copilot writes `session.shutdown` (the per-model input/cache token aggregate) to `events.jsonl` **after** the `sessionEnd` hook returns:

```
line 29  hook.start  sessionEnd
line 30  hook.end    sessionEnd
line 31  session.shutdown      ← last line, written AFTER the hook
```

The hook's inline-drain runs *before* that line exists and then kills the live watcher, so the live path never delivers it. (Batch `kcap import --copilot` reads the complete file and is unaffected.) Secondary risk: a session that ends promptly can also drop its **final assistant turn**, since the watcher is killed before draining the tail.

## Fix

Spawn a detached `kcap copilot-finalize <sid> <path>` process from the Copilot `sessionEnd` hook, **after** the session-end POST (so its poll budget isn't consumed by the pre-drain + POST). It outlives the hook, polls until `session.shutdown` is the terminal transcript line (or a 10s budget), then performs **one idempotent inline-drain**.

- **No deadlock / no stall.** Fully decoupled from the hook's return and from the server's `StopAndDrainAsync` (which sends `StopWatcher` and waits up to 10s on session-end). The watermark + deterministic event ids make the late delivery safe.
- **Timeout fallback** rescues a dropped final assistant turn when `session.shutdown` never lands (e.g. Copilot crash).
- Also redact `WatcherManager.InlineDrainAsync` output via `SecretRedactor` — it previously skipped redaction unlike the watcher's drain, a latent leak now exercised by the finalize tail.

Rejected the alternative "keep the watcher alive to observe shutdown": the server actively stops the watcher on session-end and waits on drain-complete, so lingering would deadlock or force a ~10s stall on every Copilot session-end.

## Dependency

Server-side ingest of `session.shutdown` → `CopilotUsageBackfilled` ships separately in **kurrent-io/kcap-server#763**. Until that merges + deploys, this PR reliably *delivers* the line but token totals are unchanged. Real-`copilot` live E2E should run against a #763-inclusive server.

## Testing

- 8 new tests: terminal-line detection (incl. resume-safety + malformed) and timing via WireMock (already-present, **slow-hook delayed write**, **timeout fallback**).
- Full Unit suite **1544/1544** green (confirms the shared `InlineDrainAsync` redaction change is safe); Integration **30/30** green.
- AOT publish clean — no IL2026/IL3050 warnings; native arm64 binary builds.

`copilot-finalize` is an internal command (spawned by the hook, like `watch` / `generate-whats-done`) — not user-facing, so no README change.

Linear: https://linear.app/kurrent/issue/AI-897

### DIFF
--- a/src/Capacitor.Cli/Commands/CopilotFinalizeDrainCommand.cs
+++ b/src/Capacitor.Cli/Commands/CopilotFinalizeDrainCommand.cs
@@ -1,0 +1,142 @@
+using System.Text.Json;
+using Capacitor.Cli.Core;
+
+namespace Capacitor.Cli.Commands;
+
+/// <summary>
+/// Detached post-<c>sessionEnd</c> "finalize drain" for GitHub Copilot CLI
+/// sessions (AI-897).
+/// </summary>
+/// <remarks>
+/// Copilot appends <c>session.shutdown</c> — the per-model input/cache token
+/// aggregate the server consumes as <c>CopilotUsageBackfilled</c> — and
+/// occasionally the final assistant turn, to <c>events.jsonl</c> only AFTER the
+/// <c>sessionEnd</c> hook returns (verbatim live tail: <c>hook.start</c>,
+/// <c>hook.end</c>, <c>session.shutdown</c>). The hook's own inline-drain runs
+/// before that line exists and then kills the live watcher, so nothing is left
+/// to deliver it — input/cache totals stay 0 for live sessions. (Batch
+/// <c>kcap import --copilot</c> is unaffected; it reads the complete file.)
+///
+/// <see cref="WatcherManager.SpawnCopilotFinalizeDrain"/> launches this command
+/// detached AFTER the session-end POST, so it outlives the hook. It polls until
+/// <c>session.shutdown</c> is the terminal transcript line (or a short budget
+/// elapses), then performs one inline-drain. It is fully decoupled from the
+/// hook's return and the server's StopAndDrain, so it cannot deadlock or stall
+/// them; the server watermark + deterministic-id dedup make the late delivery
+/// idempotent. The timeout fallback also rescues a dropped final assistant turn
+/// when <c>session.shutdown</c> never lands (e.g. Copilot crash).
+/// </remarks>
+static class CopilotFinalizeDrainCommand {
+    // Measured from spawn, which the hook does AFTER its pre-drain + session-end
+    // POST — so this only has to cover the gap between the hook returning and
+    // Copilot flushing session.shutdown (sub-second to ~2s observed). Generous on
+    // purpose; the process is idle while polling and exits as soon as it drains.
+    static readonly TimeSpan DefaultPollBudget   = TimeSpan.FromSeconds(10);
+    static readonly TimeSpan DefaultPollInterval = TimeSpan.FromMilliseconds(300);
+
+    /// <summary>
+    /// Process entry point: <c>kcap copilot-finalize &lt;sessionId&gt; &lt;transcriptPath&gt;</c>.
+    /// </summary>
+    public static async Task<int> Run(string baseUrl, string sessionId, string transcriptPath) {
+        // The spawning hook closes our redirected std streams the instant it
+        // starts us; writing to the dead pipe would fault. Redirect to a
+        // per-session log file, mirroring WatchCommand.
+        try {
+            var logDir = PathHelpers.ConfigPath("logs");
+            Directory.CreateDirectory(logDir);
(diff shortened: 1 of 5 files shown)


src/Capacitor.Cli/Commands/CopilotHookCommand.cs
```diff
@@ -219,7 +219,22 @@ await Console.Error.WriteLineAsync(
             forwarded["agent_host_id"] = agentHostId;
         }
 
-        return await PostHookAsync(baseUrl, "session-end/copilot", forwarded.ToJsonString());
+        var exit = await PostHookAsync(baseUrl, "session-end/copilot", forwarded.ToJsonString());
```
realtonyyoung: This still leaves a hole in the exact failure mode the finalizer is meant to make robust: `PostHookAsync` uses the default `PostWithRetryAsync` budget (30s), and Copilot hook timeouts are commonly 30s too. If the server is slow/unreachable here, the hook can be killed before line 235 ever runs, so no detached finalizer is spawned and the post-sessionEnd tail is lost. Can we either spawn the finalizer before this uncapped lifecycle POST, or wrap this POST in a shorter budget so finalize creation is guaranteed to happen before Copilot can terminate the hook?

realtonyyoung: Fixed in 829fd52 — moved `SpawnCopilotFinalizeDrain` to the top of `HandleSessionEnd` (before the capped pre-drain and the POST), so the detached finalizer is created even if the POST hangs and Copilot SIGKILLs the hook; being detached (setsid) it survives the kill and still delivers the tail. Bumped its poll budget 10s→45s to outlast the worst-case hook lifetime (PreHookDrainCap + Copilot ~30s timeout). Left the lifecycle POST uncapped to avoid the session sticking Active.

src/Capacitor.Cli/Commands/CopilotFinalizeDrainCommand.cs
```diff
@@ -0,0 +1,145 @@
...
+/// Copilot appends <c>session.shutdown</c> — the per-model input/cache token
+/// aggregate the server consumes as <c>CopilotUsageBackfilled</c> — and
+/// occasionally the final assistant turn, to <c>events.jsonl</c> only AFTER the
+/// <c>sessionEnd</c> hook returns (verbatim live tail: <c>hook.start</c>,
+/// <c>hook.end</c>, <c>session.shutdown</c>). The hook's own inline-drain runs
+/// before that line exists and then kills the live watcher, so nothing is left
+/// to deliver it — input/cache totals stay 0 for live sessions. (Batch
+/// <c>kcap import --copilot</c> is unaffected; it reads the complete file.)
+///
+/// <see cref="WatcherManager.SpawnCopilotFinalizeDrain"/> launches this command
```
realtonyyoung: This remark is stale after 829fd52: the finalizer is now spawned before the pre-drain/session-end POST, not after the session-end POST. Since the timing is the core correctness property here, can we update this XML doc (and any matching comments such as `SpawnCopilotFinalizeDrain`) so future readers do not reason from the old ordering?

realtonyyoung: Done in the latest commit — updated the `CopilotFinalizeDrainCommand` `<remarks>` and the `SpawnCopilotFinalizeDrain` summary to describe the spawn-first ordering (before the pre-drain + POST). Grepped for other stale "after the POST" wording; the only remaining hit is an unrelated pre-existing comment in `TimeBudget.cs`.

realtonyyoung:
One more stale-doc nit from the rerun: the file-level `CopilotHookCommand` wire-contract summary still says `sessionEnd → kill watcher + capped inline drain, then POST /hooks/session-end/copilot` (around lines 26-27), but the implementation now has a correctness-critical first step: spawn the detached `copilot-finalize` drainer before the pre-drain and POST. Can we update that summary too so it reflects the final ordering?

realtonyyoung:
Done in the latest commit — updated the `CopilotHookCommand` wire-contract `<remarks>` so the `sessionEnd` entry leads with "spawn the detached copilot-finalize drainer FIRST" before the watcher kill + capped inline drain + POST. Also re-swept all finalize/shutdown-ordering comments across `src/Capacitor.Cli`; the rest are either already spawn-first or are correct statements that Copilot writes `session.shutdown` after the hook returns.

realtonyyoung:
Approved to merge.

PR #170, opened 2026-06-22
[AI-900] CLI: ingest Gemini subagents (import + live child watchers)

realtonyyoung:
## What

Ingest **Gemini CLI subagents** — the nested transcripts Gemini records at
`~/.gemini/tmp/<project>/chats/<parentSessionId>/<subId>.jsonl` when a session
spawns a subagent via the `invoke_agent` tool. Previously the CLI captured the
main Gemini session only (the nested files were ignored). Now they're captured
both **on historical import** and **live**, and nest under the parent in the
trace.

Pairs with server PR kurrent-io/kcap-server#774 (correlation + idempotent
lifecycle). Closes the CLI half of **AI-900**.

## How

The parent's `invoke_agent` tool call persists `agentId == subId` plus
`args.agent_name`, which is the subagent's only on-disk type signal. A shared
`GeminiSubagentDiscovery` helper (`Capacitor.Cli.Core/Gemini/`) is the single
source of truth for both paths:

- **Discover** nested `chats/<dashedParent>/*.jsonl` files for a given main transcript.
- **Resolve** each subId → `agent_name` from the parent's `invoke_agent` call.
- **Canonicalize** the (dashed) subId to the dashless form the server routes/correlates on.
- **Build** the vendor-agnostic `/hooks/subagent-start|stop` payloads.

**Import** (`GeminiImportSource.ImportSubagentsAsync`): after the main batches and
before session-end, enumerate subagents → fail-closed `subagent-start` →
`/hooks/transcript` batches tagged `agent_id` + `vendor=gemini` → `subagent-stop`.

**Live** (`WatchCommand.ScanGeminiSubagents`): the parent watcher scans for new
nested files after each drain; for each unseen subagent it posts `subagent-start`
(fail-closed) and spawns a **child watcher** (`WatcherManager.EnsureWatcherRunning`,
key `{session}-{agentId}`) that streams the subagent transcript live.

**Teardown** (`GeminiHookCommand.DrainGeminiSubagentsAsync`): on `SessionEnd`, for
each subagent: kill the child watcher → inline-drain the tail → post `subagent-stop`.

All subagent lifecycle/content carries the **canonical (dashless) `agent_id`** and
`vendor=gemini`, so the server routes it to `AgentSubsession-*` and correlates the
`SubagentStarted`/`SubagentCompleted` lifecycle (deterministic ids → idempotent).

## Tests

- `GeminiSubagentDiscoveryTests` (unit) — enumerate / resolve-type / canonicalize.
- `GeminiSubagentImportTests` (integration) — full discover → classify → import;
  asserts the `subagent-start`/`-stop` ordering, canonical `agent_id`,
  `agent_type` from the parent `agent_name`, and `vendor=gemini` on the subagent batch.
- **Full Unit suite: 1650/1650.** AOT publish clean (0 IL2026/IL3050).

> Note: the Integration suite has one **pre-existing, unrelated** flake
> (`ClaudeHookStdoutTests.Emits_nudge_envelope_when_server_returns_newer_version_only`)
> — a process-global `Console.SetOut` capture race that fails identically on
> clean `main` without this branch, and passes in isolation. Not introduced here.

## Live E2E (real Gemini run)

Drove a real `gemini` session (API-key auth) that spawned a `codebase_investigator`
subagent against a fresh local server + KurrentDB:

- Parent watcher discovered the nested subagent file → posted `subagent-start`
  (`SubagentStarted` on the parent `AgentSession-` stream).
- Spawned child watcher streamed the subagent live (22 lines) → `AgentSubsession-`
  populated with the subagent's turns (thinking ×2, tool calls ×4, results ×4, user ×1).
- `SessionEnd` teardown killed the child watcher and posted `subagent-stop` →
  `SubagentCompleted` + `AgentSummaryComputed` on **both** streams, each exactly once
  (server idempotency holds).

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,6 +417,8 @@ kcap plugin remove --gemini                 # removes only kcap's entries
 
 Live sessions stream from the chat-recording JSONL Gemini names in each hook's `transcript_path` (`~/.gemini/tmp/<project>/chats/session-*.jsonl`); historical sessions import via `kcap import --gemini`. Sessions resumed with `gemini --resume` reattach to the same recorded session. (Historical import leaves working-directory / repo enrichment empty — Gemini doesn't record the cwd in a machine-readable header; live capture gets it from the hook payload.)
 
+Spawned subagents are captured too: Gemini records each in a nested `chats/<session>/<subId>.jsonl`, and both live capture (a child watcher per subagent) and `kcap import --gemini` discover and stream them, so they nest under the parent session in the trace.
+
 #### AWS Kiro CLI hooks
 
 AWS Kiro CLI (the rebranded Amazon Q Developer CLI) is detected via `~/.kiro/` or the `kiro` / `kiro-cli` binary on `PATH`. Kiro hooks fire only for the **active** agent — there is no global hook — so to capture every session transparently, `install --kiro` **clones your current default agent** into `~/.kiro/agents/kcap.json` (preserving its tools; a minimal agent would lose tool access), adds kcap's `agentSpawn` hook, and makes it your default agent (`chat.defaultAgent` in `~/.kiro/settings/cli.json`). This needs `kiro-cli` on `PATH` to perform the clone. Restart any running `kiro` session after installing. `remove --kiro` restores your previous default agent and deletes `kcap.json`.
--- a/src/Capacitor.Cli.Core/Gemini/GeminiPaths.cs
+++ b/src/Capacitor.Cli.Core/Gemini/GeminiPaths.cs
@@ -42,4 +42,15 @@ public static class GeminiPaths {
     /// <summary>Chat-recording directory for a project tmp dir: <c>&lt;tmp&gt;/&lt;project&gt;/chats</c>.</summary>
     public static string ChatsDir(string projectTmpDir)
         => Path.Combine(projectTmpDir, "chats");
+
+    /// <summary>
+    /// Nested subagent-recording directory for a parent session:
+    /// <c>&lt;chats&gt;/&lt;parentSessionId&gt;/</c>. Gemini records each subagent's transcript
+    /// here as <c>&lt;subId&gt;.jsonl</c> (subId = a fresh dashed UUID — the executor's agent
+    /// id; nested subagents at any depth stay flat under the top-level session).
+    /// <paramref name="parentSessionId"/> is the DASHED form from the parent transcript's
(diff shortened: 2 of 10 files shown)


realtonyyoung:
BLOCKING: `src/Capacitor.Cli/Commands/WatchCommand.cs:354` only posts the parent `/hooks/session-end/gemini` when the watcher takes over after the Gemini process dies, but it never runs the new Gemini subagent teardown path (`DrainGeminiSubagentsAsync`) that the normal `GeminiHookCommand` session-end path runs. The child watchers spawned at `WatchCommand.cs:407` monitor the same parent process and, because `RunWatch` explicitly skips parent-exit lifecycle for `agentId != null`, they can final-drain their files and exit without ever posting `subagent-stop`. In the crash/no-hook path this leaves the server with `SubagentStarted` and subagent content but no `SubagentCompleted`/summary for those live Gemini subagents; historical import can repair it later, but live recording of an ended session is wrong until manual recovery. The parent-exit Gemini session-end path needs to enumerate/kill/drain/finalize nested subagents before or as part of posting the synthetic session-end.

realtonyyoung:
Fixed in 626bd52.

Confirmed exactly as described: the watcher's parent-exit fallback (`PostSessionEndOnParentExitAsync`) posted the synthetic `session-end` but never ran the subagent teardown, and the child watchers spawned in `ScanGeminiSubagents` carry no parent-pid watchdog — so on a crash that bypasses the session-end hook, subagents were left with `SubagentStarted` + content but no `SubagentCompleted` until a manual re-import.

**Fix:** extracted the teardown into a shared `GeminiSubagentTeardown.DrainAsync` (single source of truth, parallel to `GeminiSubagentDiscovery`) and call it from **both** session-end paths — the `GeminiHookCommand` hook path and the watcher parent-exit fallback. On the parent-exit path it runs **before** the session-end POST (so `SubagentCompleted` lands ahead of `SessionEnded`) and is wrapped in the existing `TimeBudget.RunCappedAsync` cap so a slow drain can't block the watchdog's self-termination. Each step (kill → drain → subagent-stop) is now independently best-effort, so `subagent-stop` is always attempted even if kill/drain hiccups.

**Regression test:** `ParentExit_FinalizesGeminiSubagents_BeforeSessionEnd` — asserts `subagent-stop` posts with the canonical `agent_id` and precedes `session-end`. Unit 1650/1650, Integration 36/36, AOT clean.

PR #172, opened 2026-06-23
Treat Pi branch summaries as import-relevant

realtonyyoung:
## Summary
- Treat non-empty Pi `branch_summary` entries as import-relevant so partial imports resume them instead of leaving sessions stuck behind the watermark.
- Keep empty branch summaries ignored.
- Extend Pi import unit and routed import integration tests.

## Tests
- `/Users/tony/.dotnet/dotnet run --project test/Capacitor.Cli.Tests.Unit/Capacitor.Cli.Tests.Unit.csproj -- --treenode-filter "/*/*/PiImportSourceTests/*"`
- `/Users/tony/.dotnet/dotnet run --project test/Capacitor.Cli.Tests.Integration/Capacitor.Cli.Tests.Integration.csproj -- --treenode-filter "/*/*/PiImportSourceImportTests/*"`

### DIFF
--- a/src/Capacitor.Cli/Commands/PiImportSource.cs
+++ b/src/Capacitor.Cli/Commands/PiImportSource.cs
@@ -368,11 +368,13 @@ internal sealed class PiImportSource : IImportSource {
     /// True when the line maps to at least one canonical event under the
     /// server's <c>PiTranscriptNormalizer</c>: the <c>session</c> header always
     /// emits; <c>compaction</c> emits a <c>ContextCompacted</c> (AI-892);
+    /// <c>branch_summary</c> emits an <c>AssistantTextGenerated</c> with Pi metadata
+    /// (AI-892);
     /// <c>message</c> emits for roles user/assistant (non-empty content),
     /// toolResult (has toolCallId), bashExecution (has command). Everything else
-    /// (model_change / thinking_level_change / branch_summary / label /
-    /// session_info / custom*) is skipped server-side and can never advance the
-    /// watermark. Keep in sync with the normalizer.
+    /// (model_change / thinking_level_change / label / session_info / custom*) is
+    /// skipped server-side and can never advance the watermark. Keep in sync with
+    /// the normalizer.
     /// </summary>
     internal static bool IsImportRelevantLine(string line) {
         try {
@@ -380,8 +382,9 @@ internal sealed class PiImportSource : IImportSource {
             var       root = doc.RootElement;
 
             switch (root.Str("type")) {
-                case "session":    return true;
-                case "compaction": return true;
+                case "session":        return true;
+                case "compaction":     return true;
+                case "branch_summary": return root.Str("summary") is { Length: > 0 };
                 case "message":
                     if (root.Obj("message") is not { } msg) return false;
                     return msg.Str("role") switch {
--- a/src/Capacitor.Cli/Commands/PiImportSource.cs
+++ b/src/Capacitor.Cli/Commands/PiImportSource.cs
@@ -384,7 +384,7 @@ internal sealed class PiImportSource : IImportSource {
             switch (root.Str("type")) {
                 case "session":        return true;
                 case "compaction":     return true;
-                case "branch_summary": return root.Str("summary") is { Length: > 0 };
+                case "branch_summary": return !string.IsNullOrWhiteSpace(root.Str("summary"));
                 case "message":
                     if (root.Obj("message") is not { } msg) return false;
                     return msg.Str("role") switch {
--- a/test/Capacitor.Cli.Tests.Integration/PiImportSourceImportTests.cs
+++ b/test/Capacitor.Cli.Tests.Integration/PiImportSourceImportTests.cs
@@ -33,7 +33,6 @@ public class PiImportSourceImportTests : IDisposable {
         File.WriteAllLines(path, new[] {
             $$"""{"type":"session","version":3,"id":"{{DashedSid}}","timestamp":"2026-06-12T10:00:00.000Z","cwd":"/work/a"}""",
             """{"type":"message","id":"a1","parentId":null,"timestamp":"2026-06-12T10:00:01.000Z","message":{"role":"user","content":"hello"}}""",
-            """{"type":"branch_summary","id":"b1","parentId":"a1","timestamp":"2026-06-12T10:00:01.500Z","fromId":"a1","summary":"branch summary kept for import"}""",
             """{"type":"message","id":"a2","parentId":"a1","timestamp":"2026-06-12T10:00:02.000Z","message":{"role":"assistant","content":[{"type":"text","text":"hi there"}],"model":"gpt-5","usage":{"input":10,"output":3}}}"""
         });
         return sessionsDir;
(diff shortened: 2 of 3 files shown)


src/Capacitor.Cli/Commands/PiImportSource.cs
```diff
@@ -368,20 +368,23 @@ internal sealed record PiHeader(string? SessionUuid, string? Cwd, DateTimeOffset
...
         try {
             using var doc  = JsonDocument.Parse(line);
             var       root = doc.RootElement;
 
             switch (root.Str("type")) {
-                case "session":    return true;
-                case "compaction": return true;
+                case "session":        return true;
+                case "compaction":     return true;
+                case "branch_summary": return root.Str("summary") is { Length: > 0 };
```
realtonyyoung: **Empty-check diverges from the server — narrow, but it produces a re-import loop.**

The server's `NormalizeBranchSummary` (kcap-server#781) gates with `string.IsNullOrWhiteSpace(summary)` and returns `null` (emits no event, so the watermark never advances) for null / empty / **whitespace-only** summaries. This predicate uses `Length > 0`, so a whitespace-only `summary` (e.g. `" "` or a lone newline) counts as import-relevant here while the server emits nothing.

That's the dangerous direction: `lastImportable` points at a line the watermark can never reach, so the session classifies `Partial` and re-imports on every run — exactly the "Partial forever" failure this function's docstring warns about. The `""` case already agrees; only whitespace-only diverges. Mirror the server's check:

```suggestion
                case "branch_summary": return !string.IsNullOrWhiteSpace(root.Str("summary"));
```

test/Capacitor.Cli.Tests.Unit/PiImportSourceTests.cs
```diff
@@ -140,6 +140,8 @@ public async Task does_not_support_title_generation() {
     [Arguments("""{"type":"message","id":"a","message":{"role":"bashExecution","command":"ls"}}""", true)]
     [Arguments("""{"type":"model_change","id":"a","modelId":"gpt-5"}""", false)]
     [Arguments("""{"type":"compaction","id":"a","summary":"x"}""", true)]  // AI-892: compaction → ContextCompacted
+    [Arguments("""{"type":"branch_summary","id":"a","summary":"branch","fromId":"b"}""", true)] // AI-892: branch_summary → AssistantTextGenerated
+    [Arguments("""{"type":"branch_summary","id":"a","summary":""}""", false)]
```
realtonyyoung, line 144: Add the whitespace-only case — that's the input that actually diverges from the server's `IsNullOrWhiteSpace` gate, and it's currently untested. It fails against the present `Length > 0` predicate, so it pairs with the one-liner above:

```suggestion
    [Arguments("""{"type":"branch_summary","id":"a","summary":""}""", false)]
    [Arguments("""{"type":"branch_summary","id":"a","summary":"   "}""", false)] // whitespace-only → server skips (IsNullOrWhiteSpace)
```

test/Capacitor.Cli.Tests.Integration/PiImportSourceImportTests.cs
```diff
@@ -96,6 +97,8 @@ await Assert.That(posts).IsEquivalentTo(new[] {
             .Single(e => e.RequestMessage.Path == "/hooks/transcript")
             .RequestMessage.Body;
         await Assert.That(transcriptBody!).Contains("\"vendor\":\"pi\"");
+        await Assert.That(transcriptBody!).Contains("branch_summary");
+        await Assert.That(transcriptBody!).Contains("branch summary kept for import");
```
realtonyyoung: These two assertions pass with or without this PR's change. `SendTranscriptBatches` forwards every non-blank line from `startLine` regardless of `IsImportRelevantLine`, and this is a fresh import (`startLine = 0`), so the `branch_summary` line lands in the body either way. They verify forwarding, not the watermark/resume behavior the PR is about — revert the one-line change in `PiImportSource.cs` and this test stays green.

To actually guard the classifier path, seed a server `last-line` watermark just before a trailing `branch_summary` and assert the session comes back `Partial` (resumes) rather than `AlreadyLoaded`. kcap-server#781 has this end-to-end in `PiAi892ImportE2ETests` (asserts `ResumeFromLine == initialLines.Length`); the CLI classifier is unguarded here.

realtonyyoung (commented):
Reviewed against the server normalizer in kcap-server#781, which this PR's `IsImportRelevantLine` predicate has to mirror (the docstring's "Keep in sync with the normalizer"). The `branch_summary` → import-relevant change is the right call and closes the documented "stuck behind the watermark" resume gap. One should-fix on the empty-check semantics, plus two test notes — inline.

realtonyyoung:
Updated for the review comments:

- Changed Pi branch_summary relevance to use IsNullOrWhiteSpace parity with the server.
- Added the whitespace-only branch_summary unit case.
- Replaced the weak fresh-import body assertions with a resume/classification guard where the only remaining relevant line is a trailing branch_summary.

Tests run:
- /Users/tony/.dotnet/dotnet run --project test/Capacitor.Cli.Tests.Unit/Capacitor.Cli.Tests.Unit.csproj -- --no-progress (1655 passed)
- /Users/tony/.dotnet/dotnet run --project test/Capacitor.Cli.Tests.Integration/Capacitor.Cli.Tests.Integration.csproj -- --no-progress (37 passed)

test/Capacitor.Cli.Tests.Integration/PiImportSourceImportTests.cs
```diff
@@ -33,11 +33,24 @@ string WriteSessionFile() {
         File.WriteAllLines(path, new[] {
             $$"""{"type":"session","version":3,"id":"{{DashedSid}}","timestamp":"2026-06-12T10:00:00.000Z","cwd":"/work/a"}""",
             """{"type":"message","id":"a1","parentId":null,"timestamp":"2026-06-12T10:00:01.000Z","message":{"role":"user","content":"hello"}}""",
+            """{"type":"branch_summary","id":"b1","parentId":"a1","timestamp":"2026-06-12T10:00:01.500Z","fromId":"a1","summary":"branch summary kept for import"}""",
```
realtonyyoung: Minor / optional: this `branch_summary` line was added in the first commit to back the `Contains("branch_summary")` assertions in `ImportSession_posts_lifecycle_and_transcript_with_pi_vendor`. Those assertions are gone now (good call — they didn't exercise the predicate), and the real coverage moved to `WriteSessionFileEndingWithBranchSummary` + `Classify_treats_trailing_branch_summary_as_import_relevant_when_resuming`.

So this line is now unasserted fixture data in the shared helper. Fine to keep as incidental "a branch_summary doesn't break the New-import / resume flows" coverage — but if that wasn't the intent, it can be dropped to keep `WriteSessionFile()` minimal.

realtonyyoung (commented):
Re-reviewed at `f09bea3`. All three points from the previous pass are resolved:

- empty-check now mirrors the server's `IsNullOrWhiteSpace` gate (verbatim) — predicate is in sync;
- whitespace-only unit case added;
- the forwarding-only `Contains(...)` assertions were dropped in favor of `Classify_treats_trailing_branch_summary_as_import_relevant_when_resuming`, which seeds a watermark and asserts `Partial` / `ResumeFromLine == 2` — i.e. it actually fails if the change is reverted. 

One minor/optional leftover, inline. Otherwise this looks good.

realtonyyoung:
Follow-up pushed in `f746da3`.

Review action taken:
- Removed the now-unused `branch_summary` fixture line from `PiImportSourceImportTests.WriteSessionFile`.

Verification:
- `/Users/tony/.dotnet/dotnet run --project test/Capacitor.Cli.Tests.Integration/Capacitor.Cli.Tests.Integration.csproj -- --no-progress` -> 37 passed.

PR #173, opened 2026-06-24
[AI-921] OpenCode live-ingest plugin + subagent discovery (7th vendor)

realtonyyoung:
Adds SST OpenCode as a live-ingest vendor on the CLI side (AI-919 / AI-921). OpenCode has no shell hooks, so kcap ships a dependency-free TypeScript plugin (`~/.config/opencode/plugins/kcap.ts`) that OpenCode auto-loads.

## What it does
- **Plugin** (`OpenCodeExtensionInstaller`): on `session.created` shells `kcap hook --opencode --event session-start` (POSTs lifecycle + spawns the transcript watcher); on `session.idle` fetches the session's messages via the in-process SDK `client` and appends `{info,parts}` JSONL the watcher tails. Dependency-free (`node:` builtins only) + fail-safe.
- **Dispatcher** (`OpenCodeHookCommand`): `kcap hook --opencode` → POST `/hooks/session-start/opencode` + `WatcherManager.EnsureWatcherRunning(vendor:opencode)`. The watcher owns session-end via the parent-pid watchdog (Kiro pattern).
- **Setup / plugin / uninstall wiring**: `kcap setup` detection (`--skip-opencode-hooks`), `kcap plugin install|remove --opencode` (+ `--if-installed` npm refresh), `kcap uninstall`.
- **Subagents (phase 2)**: on the parent's `session.idle` the plugin discovers child sessions via `client.session.children` and writes each child's `{info,parts}` into a nested dir (`<cache>/<parent>/<child>.jsonl`); the parent watcher (`ScanOpenCodeSubagents`) registers each subagent (`/hooks/subagent-start`) + spawns a child watcher streaming it with the child's `agent_id` → `AgentSubsession-*`, then tears them down on parent exit. Mirrors the Gemini AI-900 machinery.
- **Fix**: added `opencode` to `WatchCommand.KnownVendors` — without it the watcher's parent-exit `session-end` POST was rejected for opencode (so the main-session session-end-on-exit never fired).

## Testing
- New `OpenCodeSubagentDiscoveryTests` + existing `OpenCodeExtensionInstallerTests` / `PluginCommandOpenCodeTests` / `CodingAgentsStepTests` cases.
- Full CLI unit suite green (1673); AOT publish clean (no IL3050/IL2026).
- The server-side wire is proven by `OpenCodeSubagentImportTests` (kcap-server) plus a live E2E against a dev server.

Pairs with the kcap-server PR (server normalizer / routes / subagent correlation); the server picks this up via a submodule bump after merge.


realtonyyoung:
## Review findings

### BLOCKING

- `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:100` / `:141` / `:143` - OpenCode child flushing can permanently record a partial child message and/or duplicate a child as a top-level session. The plugin treats a message as written once `id:parts.length` has been seen, so if `flushSubagents(parent)` observes a still-streaming child and the text inside an existing part later changes without increasing `parts.length`, the final content is never appended on a later idle. The same race gets worse if `session.created` lacks `info.parentID` or the child `session.idle` fires before the parent goes idle: the child goes through `start(sid)`/`flushTo(sid, file(sid))` as a top-level OpenCode session, and the later parent flush can also stream it as a subagent. Concrete fix: do not rely solely on the parent idle/task-blocking assumption or `info.parentID`; track child ids from `client.session.children(parent)` before deciding top-level ingestion, and make `flushTo` safe for in-place part updates (for example key by a digest/version of `parts`, or only mark writes after a session is known complete). If OpenCode exposes child completion/parent metadata on the message/session object, gate child-file writes on that instead of first sighting.

### Non-blocking

- `src/Capacitor.Cli/Commands/OpenCodeSubagentTeardown.cs:37` / `:38` / `:39` and `src/Capacitor.Cli/Commands/WatchCommand.cs:554` - the OpenCode parent-exit teardown has one 8s cap for all subagents, but each subagent is processed serially and `KillWatcher` alone can spend up to 5s before the inline drain and `subagent-stop` POST. A slow first child can consume the cap and prevent later children from ever getting `SubagentCompleted`; unlike Gemini/Pi, the README says there is no `kcap import --opencode` recovery path. Concrete fix: make the shutdown path deadline-aware per step, avoid the 5s kill wait here, and ensure every discovered child gets a best-effort `subagent-stop` even when drains time out (drain-before-stop for each child, but don't let one child starve the rest).

- `src/Capacitor.Cli/Commands/WatchCommand.cs:879` / `:934` / `:1002` - `vendor == "opencode"` is missing from the title/event extraction gates. OpenCode transcript lines are `{ info, parts }`, but `TryExtractUserText`, `TryExtractAssistantText`, and `IsEvent` fall through to the Claude shape (`type: "user"|"assistant"`). Live capture still streams, but OpenCode sessions will not get initial titles or LLM-generated titles from the watcher. Concrete fix: add OpenCode-specific extractors that read `info.role` and text parts from `parts`, and add an `IsEvent` branch for meaningful user/assistant OpenCode messages.

### Nit

- `src/Capacitor.Cli/Program.cs:513` - the `kcap watch` usage string still lists `--vendor claude|codex|copilot|kiro|pi`, omitting both `gemini` and the newly-supported `opencode`. Concrete fix: include all accepted watcher vendors in the usage text.

### Checked while reviewing

- `WatchCommand.ScanOpenCodeSubagents` does post `subagent-start` before spawning the child watcher and removes from `seen` on POST failure, so that path is fail-closed.
- The child watcher starts with a fresh local `WatchState`; provided the server watermark for the new `(sessionId, agentId)` is `0`, lines already written before the watcher connects are read from the beginning.
- `OpenCodeSubagentDiscovery.BuildStartPayload` and `BuildStopPayload` include the required HookBase/SubagentStopHook fields called out in the review request.
- I did not see new C# reflection/dynamic patterns that should trip NativeAOT IL3050/IL2026; the embedded TypeScript `any` is just string content.

Verification note: per the latest instruction, I did not run tests or publish checks.

realtonyyoung:
Thanks — all four addressed in `eff3365`:

- **BLOCKING (child dedup/classification + partial content):** the plugin now resolves a session's `parentID` authoritatively via `client.session.get` (new `isChild`) instead of relying only on `session.created`'s `info.parentID`, so a child can't be misfiled as a top-level session even if its `session.idle` fires before the parent is discovered. Child flushing is gated on message finality (`finalOnly` → `isFinal`: assistant messages only once `finish`/`time.completed` is set), so a still-streaming child can't write partial content as the first — and, given the server keeps-first-by-`prt_`-id, permanent — append.
- **Titles (non-blocking):** added `opencode` branches to `TryExtractUserText` / `TryExtractAssistantText` / `IsEvent` that read `info.role` + non-hidden text parts (mirrors the server normalizer). OpenCode sessions now get watcher titles; tool-only turns don't count toward the threshold. New `OpenCodeExtractorTests`.
- **Teardown cap (non-blocking):** `OpenCodeSubagentTeardown` now caps each step per child, so `KillWatcher`'s up-to-5s graceful-exit wait can't consume the shared budget and starve later children of their `SubagentCompleted`.
- **Nit:** the `kcap watch --vendor` usage string now lists `gemini` + `opencode`.

Full CLI unit suite 1677 green; AOT publish clean (no IL3050/IL2026). Your "checked while reviewing" notes all hold (child watcher reads from server-watermark 0; `subagent-start` is fail-closed; payloads carry the required HookBase fields).

realtonyyoung:
## Re-review findings for eff3365

### BLOCKING

- `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:81` / `:175` / `:178` - the child-classification fix still fails open in the exact ambiguity case we need to make safe. If `session.created` lacks `info.parentID` and `client.session.get` has a transient failure, `isChild` returns `false`, then `start(sid)` registers a top-level OpenCode session and starts a watcher on `<dir>/<child>.jsonl`. Later, when the parent idles, `flushSubagents` adds the same child id to `children` and writes `<dir>/<parent>/<child>.jsonl`, so the child is duplicated as both a top-level session and a subagent. The outer event handler catches throws, so this won't disrupt OpenCode, but the fail-open result corrupts capture. Concrete fix: make `isChild` tri-state (`child` / `top-level` / `unknown`) and do not start top-level ingestion on `unknown`; retry on the next idle, or only fail open after a successful `session.get` proves there is no parent. For `session.created` with missing parentID, deferring start until the first idle is safer than permanently misfiling a child.

- `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:100` / `:132` / `:158` - `finalOnly` can permanently drop child assistant messages. Child flush now writes user messages immediately but skips assistant messages unless `info.finish` or `info.time.completed` is present. If OpenCode's SDK omits those fields on completed child messages, uses a different completion field, or returns assistant messages whose completion is only session-level, every parent idle will keep skipping that assistant message forever. The parent-exit teardown cannot recover it: `OpenCodeSubagentTeardown` only enumerates and drains the already-written child JSONL files, and never has access to the OpenCode SDK/SQLite child messages. A child file with only user messages will be finalized incomplete; a child with no written final-looking messages may never be discovered at all. Concrete fix: gate on child/session completion rather than per-message marker, or add a safe fallback that flushes all child messages once the parent task has returned / parent is idle / process is exiting. At minimum, if final markers are absent after a child is discovered from `client.session.children(parent)`, retry for a bounded interval and then write the full child transcript instead of silently skipping it forever.

### Non-blocking

- `src/Capacitor.Cli/Commands/WatchCommand.cs:556` and `src/Capacitor.Cli/Commands/OpenCodeSubagentTeardown.cs:42` - per-step caps improve the previous starvation issue, but the outer `OpenCodeSubagentTeardown.DrainCap` is still 8s while one child can consume up to ~6.5s (`KillWatcher` 1.5s + drain 2.5s + stop 2.5s). With multiple children, the outer `TimeBudget.RunCappedAsync` can still abandon the teardown before later children get `subagent-stop`, and the log says "until a re-run" even though OpenCode has no historical import path. Concrete fix: make the teardown deadline-aware across all children and prioritize sending `subagent-stop` for every discovered child, or raise/remove the outer cap for OpenCode while keeping individual network/drain calls capped.

### Nit

- None found.

### Checked

- `isChild` itself is wrapped defensively and cannot throw through the plugin event handler; the top-level `event` handler also catches any unexpected error.
- The `children` set is populated on `isChild(...parentID...)` and `flushSubagents`, so the known-child path is consistent once classification succeeds.
- The OpenCode title extractors now match the stated normalizer contract in this repo: joined non-hidden `type:"text"` parts, with `synthetic`/`ignored` skipped, and `IsEvent` only counts user/assistant turns with titleable text.
- I did not see new C# reflection/dynamic usage in the eff3365 changes that should introduce NativeAOT IL3050/IL2026 risk.
- Per instruction, I did not run tests, builds, or publishes.

_🤖 Reviewed by Codex_

realtonyyoung:
Thanks — both BLOCKING + the non-blocking addressed in `e79929c`:

- **BLOCKING (isChild fail-open → duplicate):** replaced `isChild` with a tri-state `classify` (child / top / unknown). Top-level ingestion starts ONLY on a confirmed "top" (a successful `client.session.get` with no parent); "unknown" (a transient get failure, or a `session.created` without parentID we couldn't confirm) DEFERS to the next idle instead of misfiling a child as both top-level and subagent.
- **BLOCKING (finalOnly drops child messages):** removed the per-message `finalOnly`/`isFinal` gate. The full child transcript is flushed at parent-idle (the task tool blocks the parent → child complete). To stay safe without relying on that invariant, the per-message dedup key now includes a **terminal-tool discriminator** — a tool transitioning running→completed changes the key and is re-appended, and the server's terminal-gate + keep-first ingest the final snapshot. No more permanent skip.
- **Non-blocking (teardown starvation):** `OpenCodeSubagentTeardown` is now deadline-aware across ALL children (a shared cleanup budget for kill+drain; `subagent-stop` ALWAYS attempted per child) and self-bounding, so the outer cap that could abandon later children is removed and the misleading "until a re-run" log is gone.

Full CLI unit suite 1677 green; AOT publish clean. The dedup discriminator lives in the embedded plugin TS (not directly unit-testable here); the server-side terminal-gate + dashed-id paths are covered by tests in #784.

realtonyyoung:
## Re-review findings for e79929c

### BLOCKING

- `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:96` / `:135` / `:149` - the new dedupe key fixes terminal tool snapshots, but partial text/reasoning content is still protected only by the comment-level invariant that "parent-idle == child complete". `dedupeKey` changes when `parts.length` or terminal tool count changes, but not when an existing `text`/`reasoning` part keeps growing. Since `flushSubagents` now writes the full child transcript immediately on the parent's idle, a child observed while assistant text is still streaming can still append the partial text first; later appends with the same part ids will be ignored by the server's keep-first normalizer. I don't see code in this PR that proves the child is complete before `flushTo(cid, childFile(...))`; it is assumed from OpenCode's task behavior. Concrete fix: either gate child flushing on a durable child/session completion signal from OpenCode before writing any assistant text, or add a test/source-linked assertion that the `task` tool cannot let the parent idle until the child is fully complete. Without that guarantee, defer child writes rather than appending partial text snapshots.

### Non-blocking

- `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:83` / `:181` - tri-state classification no longer misfiles `unknown` as top-level, and already-started top-level sessions do avoid repeated `session.get` on later idles. The tradeoff is that a real top-level session is never ingested if `client.session.get` keeps failing while `client.session.messages` would have worked: `session.created` defers on `unknown`, every idle retries `classify`, and `start`/`flushTo` never run. This is safer than duplicating children, but it creates a permanent capture hole for any endpoint-specific `session.get` failure. Concrete fix: keep a small `unknownCounts`/timestamp map and fall back to top-level after a bounded number of failed classifications or elapsed time, ideally logging/marking the fallback so duplicate risk is explicit. A child whose `session.created` was missed is otherwise handled correctly on first idle: it is unstarted, calls `classify(sid)`, and skips top-level if `session.get` returns a parent.

- `src/Capacitor.Cli/Commands/OpenCodeSubagentTeardown.cs:31` / `:48` / `:56` - removing the outer cap avoids cutting off later children, but `DrainAsync` is now bounded only as `~CleanupBudget + 2.5s * childCount` (plus file enumeration/type reads). That is fine for realistic small N, but pathological or polluted nested dirs can delay the parent-exit `session-end` POST for a long time because every discovered `*.jsonl` gets a serial capped `subagent-stop`. `CappedAsync` observes faults on timed-out tasks, so unobserved exceptions are covered, but timed-out HTTP/kill tasks are intentionally left running and can overlap the later session-end path. Concrete fix: add a hard overall shutdown deadline or a max-child guard, and prioritize `subagent-stop` with bounded concurrency so all real children get a chance without making parent session-end scale linearly with arbitrary file count.

### Nit

- `src/Capacitor.Cli/Commands/OpenCodeSubagentTeardown.cs:23` - the old one-line `<summary>` remains immediately before the new multi-line `<summary>` for `CleanupBudget`, producing duplicate XML doc summaries. Concrete fix: delete the stale line 23 summary.

### Checked

- The `started` and `children` fast paths mean a confirmed top-level session does not re-run `session.get` on every idle, and known children short-circuit before any handler work.
- Plugin handlers remain fail-safe: `runKcap`, `flushTo`, `flushSubagents`, and the outer `event` handler all swallow errors, so no throw should escape into OpenCode.
- `CappedAsync` attaches an `OnlyOnFaulted` continuation to timed-out tasks, so later faults are observed; cancellation is not an unobserved-exception risk.
- I did not see new C# reflection/dynamic usage in e79929c that would introduce NativeAOT IL3050/IL2026 risk.
- Per instruction, I did not run tests, builds, or publishes.

_🤖 Reviewed by Codex_

realtonyyoung:
Thanks — both BLOCKING + both non-blocking + the nit addressed in `4ec65bc`:

- **BLOCKING (partial child text/reasoning freeze):** right that the dedup discriminator only re-sends tool parts, leaving text/reasoning on the assumed invariant. Child flushing is now gated on a **durable completion signal** — the parent's `task` tool part being terminal (`state.metadata.sessionId` present + `state.status` completed/error), via a new `completedChildIds(parent)`. A child whose task isn't terminal yet is deferred (still recorded in `children` so its own events skip top-level), so no partial child transcript is ever written first.
- **Non-blocking (capture hole on persistent get failure):** `classify` now falls back to top-level after 3 consecutive `session.get` failures (`unknownCounts`), so a persistent endpoint-specific failure can't permanently drop a real top-level session — while still deferring on transient ambiguity.
- **Non-blocking (teardown unbounded by child count):** added a hard `OverallBudget` (20s) ceiling to `DrainAsync` so a polluted/huge nested dir can't make parent-exit session-end scale linearly.
- **Nit:** removed the duplicate `<summary>` on `CleanupBudget`.

Full CLI unit suite 1677 green; AOT publish clean.

realtonyyoung:
Static re-review of `4ec65bc` only. I did not run tests, builds, or publishes.

## BLOCKING

1. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:173` / `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:194` - The new child flush gate can permanently lose a child transcript when the parent has no terminal `task` part with `state.metadata.sessionId == childId`. `flushSubagents` records every child from `client.session.children(parent)` into the global `children` set, then skips flushing unless `completedChildIds(parent)` contains that child. Once that happens, the child's own top-level event path is suppressed by `children.has(sid)`, but no nested JSONL is created. Parent-exit teardown only drains files that already exist under the subagent directory, so it cannot recover a child that was never written. This looks worse than the previous partial-text risk because the failure mode is total loss for any child path where OpenCode omits `state.metadata.sessionId`, uses a different metadata shape/casing, or creates a child session through a non-`task` path. Fix: keep the terminal-task gate as the preferred signal, but add a bounded fallback for children seen in `client.session.children(parent)` that remain pending across N parent idles or for an elapsed-time budget, and flush those children even without a matching terminal task part. If OpenCode exposes a child/session completion signal, use that as an additional gate. Also avoid adding a child to the global skip set until it either has a reliable parent classification or has been flushed/fallback-queued.

## Non-blocking

1. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:167` / `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:219` - Each parent idle now fetches the full parent transcript twice: `flushTo(sid, file(sid))` calls `client.session.messages(parent)`, then `completedChildIds(parent)` calls `client.session.messages(parent)` again. For large sessions this doubles the cost of the idle path and increases the chance that one fetch observes a different snapshot than the other. Fix: have `flushTo` return the fetched messages, or split message fetching from appending and pass the already-fetched parent messages into `completedChildIds`/`flushSubagents`.

2. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:93` / `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:217` - The classify fallback is per-session and does reset on a successful `session.get`, and already-started top-level sessions do not re-`get` on later idles. But a real child whose `session.get` fails three times transiently is now classified as top-level and ingested on the third idle; if the parent later reports it in `client.session.children`, it can be ingested again as a subagent. Three idle events can happen much faster than three meaningful retry intervals, so the count threshold is a weak proxy for confidence. Fix: make the fallback time-based/backoff-based, and before falling back to top-level, check whether any known parent task/children list references the session. At minimum, use a longer elapsed-time threshold for sessions first seen through `session.idle` rather than `session.created`.

3. `src/Capacitor.Cli/Commands/OpenCodeSubagentTeardown.cs:52` / `src/Capacitor.Cli/Commands/WatchCommand.cs:555` - `OverallBudget` does bound the loop in practice, and `CappedAsync` observes faults for detached operations, but the new `break` means teardown no longer “ALWAYS attempts subagent-stop for every child” as the watcher comment says. In pathological child counts, remaining child files are simply skipped once the budget is exhausted; that is bounded, but it is also silent. Fix: update the comment to match the budgeted behavior and log/track how many children were skipped. If the server needs a hard “every discovered child gets a stop attempt” guarantee, use bounded concurrency or a per-parent cap that still schedules a stop attempt for each child inside the overall budget.

## Nit

1. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:173` - The terminal gate assumes `task.state.metadata.sessionId` and `client.session.children(parent)[].id` are exactly the same `ses_...` identifier form. That is probably correct for current OpenCode, but the plugin has no normalization or alternate key support (`sessionID`, `session_id`, etc.), so a small upstream shape change turns into the blocking loss case above. Fix: centralize extraction/normalization for task child IDs and accept any documented aliases if OpenCode has them.

Additional checks: the event handlers are still fail-safe from a throw-escaping perspective (`classify`, `flushTo`, `completedChildIds`, `flushSubagents`, `runKcap`, and the outer event handler all catch). I did not see new C# reflection/dynamic usage in the round-3 changes that would create an AOT warning on its own.

_🤖 Reviewed by Codex_

realtonyyoung:
Thanks — the BLOCKING + all 3 non-blocking + the nit addressed in `f54e976`:

- **BLOCKING (task-gate could totally lose a child):** added a **bounded fallback** — a child stays pending across parent idles and is flushed after 3 idles even without a terminal `task` match (by then the task has long returned, so it's complete). The durable terminal-task signal stays the preferred trigger; `flushedChildren`/`pendingChildren` track state. A missing/renamed `state.metadata.sessionId` or a non-`task` spawn can no longer permanently lose the transcript.
- **Non-blocking (double parent fetch):** split into `fetchMessages`/`writeMessages`; the parent transcript is fetched ONCE per idle and threaded into both the write and `completedChildIds` (no second round-trip / snapshot skew).
- **Non-blocking (classify count threshold weak):** the get-failure fallback is now **time-based** (60s elapsed) instead of a 3-idle count, so a burst of fast idles can't prematurely misfile a real child.
- **Non-blocking (silent teardown skip):** `DrainAsync` returns the count of children left unfinalized by the overall-budget ceiling; `WatchCommand` logs it, and the "always stop every child" comment is corrected to "within the overall budget."
- **Nit:** task child-id extraction is now alias-tolerant (`sessionId` / `sessionID` / `session_id`).

Full CLI unit suite 1677 green; AOT publish clean.

realtonyyoung:
BLOCKING:
- `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:214`: the 3-idle fallback can still flush a child while its transcript is non-terminal, which reopens the partial-text freeze this gate was meant to avoid. The fallback path sets `flush = n >= PENDING_IDLE_FALLBACK` and then writes the child's full transcript without inspecting the child itself (`lines 216-223`). Three parent idles is not a sound completeness signal for async/long-running children, or for parent idles caused by unrelated parent activity while the child is still streaming. Worst case: the first child write contains an assistant `text`/`reasoning` part with a stable `prt_` id and partial content; later fetches do not re-send it because the dedupe discriminator is only `messageId:parts.length:terminalToolCount`, and even if it were re-sent the server keeps the first event for that `prt_` id. Concrete fix: the fallback should only flush after a child-level terminal signal, e.g. fetch the child and require the last assistant message to be complete (`info.finish`/`time.completed` or equivalent, plus no non-terminal tool parts), or do not write text/reasoning until terminal. A bare idle count should only keep the child pending/logged, not produce the first persisted transcript.

- `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:96`: the 60s `classify()` fallback can still double-ingest a child. After persistent `session.get` failure it returns `"top"` (`line 102`), so `session.created`/`session.idle` can call `start(sid)` as a top-level session (`lines 242` and `245-247`). If the parent later succeeds in `client.session.children`, `flushSubagents` adds the same id to `children` and writes it under the parent (`lines 207-223`); nothing tears down the already-started top-level watcher, deletes the top-level file, or prevents the nested child watcher. `started` and `children` only affect future plugin event handling, so they make the race rarer but do not prevent top-level + subagent duplicate ingest. Concrete fix: do not promote `"unknown"` to top-level without a successful parent check, or add an explicit reconciliation path that, before nested ingest, detects `started.has(cid)`/top-level fallback and prevents one of the two ingests before any server events are written. Prefer fail-closed for ambiguous ids; a top-level retry can wait for `session.get`, but a misfiled child cannot be cleanly un-ingested.

- `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:144`: `writeMessages` is not actually fail-safe on filesystem errors. It adds each dedupe key to `seen` before `mkdirSync`/`appendFileSync` succeeds (`lines 150-158`), then catches append errors. A transient disk/permission/full-volume error therefore permanently suppresses those message snapshots for the rest of the plugin process, so the watcher never sees them. The child path also marks `flushedChildren.add(cid)` before `flushTo` has fetched or written anything (`lines 220-223`); if `fetchMessages` returns `[]` or the append fails near session exit, teardown enumerates no child file and the subagent may never get `SubagentStarted`/`SubagentCompleted`. Concrete fix: have `writeMessages` build pending `(key,line)` pairs, append first, and only add keys to `seen` after the append succeeds. Return a success/count from `writeMessages`/`flushTo`, and only mark a child flushed after a successful non-empty write (or at least after a verified file exists). `fetchMessages` returning `[]` should be a retryable no-op, not a state transition.

non-blocking:
- `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:214`: once `flushedChildren` contains a child id, every later parent idle sets `flush` immediately and calls `flushTo(cid, ...)` again (`lines 214` and `220-223`). That means a long parent session with `I` later idles and `C` completed children performs `I*C` full `client.session.messages(child)` fetches and full dedupe scans, even though terminal children cannot produce useful new lines. For example, 200 later idles and 10 children is 2,000 full child transcript fetches. Concrete fix: after the blocking completeness/write-success fixes, distinguish terminal-flushed children from provisional/pending children and skip refetching terminal-flushed children entirely; only retry children whose previous fetch/write failed or whose completion has not been proven.

nit:
- none.

Other checks:
- `fetchMessages` returning `[]` is safe for parent writes because it does not advance any plugin state there; it is not safe in the child flush path today because `flushedChildren` is updated before verifying a write.
- The new teardown unfinalized-count path is bounded and best-effort as documented; I did not see a new AOT issue.
- The plugin event handler still has an outer `try/catch`, and the helper paths catch their own failures, so I did not find a throw escaping into OpenCode.

Static review only per request; I did not run tests/builds/publishes.

_🤖 Reviewed by Codex_

realtonyyoung:
Thanks — all 3 BLOCKING + the non-blocking addressed in `f3fdbac`:

- **BLOCKING (3-idle fallback could flush a streaming child):** removed the idle-count trigger entirely. A child is now flushed only when proven COMPLETE — `childComplete` requires every tool part terminal AND the last assistant message carrying `info.finish`/`time.completed` — or when the parent's `task` part is terminal. Until then it's deferred; no partial transcript is ever written, so no partial-text freeze.
- **BLOCKING (60s classify fallback could double-ingest):** now FAIL-CLOSED — an ambiguous id (persistent `session.get` failure) stays `"unknown"` and defers, never promoted to top-level. A real top-level session waits for `session.get` to recover (if the SDK is down, fetching its transcript fails too — nothing lost); a misfiled child, which can't be cleanly un-ingested, is avoided.
- **BLOCKING (writeMessages not fail-safe):** it now builds the pending (key,line) set, appends FIRST, and marks keys `seen` only after the append succeeds (returns false on failure). A child is marked flushed only after a successful write, and an empty/failed `fetchMessages` is a retryable no-op (no state transition) — so a transient disk error or empty fetch no longer permanently suppresses snapshots or skips `SubagentStarted`/`SubagentCompleted`.
- **Non-blocking (re-fetch every idle):** `flushedChildren` short-circuits before the fetch, so a completed child isn't re-fetched on every later parent idle.

Full CLI unit suite 1677 green; AOT publish clean.

realtonyyoung:
Static re-review of `f3fdbac` only. I did not run tests, builds, or publishes.

## BLOCKING

1. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:192` / `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:228` - The new child-completeness gate still has a permanent-loss path unless OpenCode guarantees at least one of these two completion signals for every completed child: the child’s final assistant message has `info.finish` or `info.time.completed`, or the parent’s terminal `task` part has a recognized child session id. If a completed child lacks both, `flushSubagents` keeps deferring forever. Because the child id has already been put in `children`, the child’s own event path is suppressed, no nested child JSONL is ever written, and parent-exit teardown cannot recover it because it only enumerates files that already exist. I do not see this invariant proven in this PR. The server-side normalizer also accepts OpenCode assistant messages without `time.completed`, and has an explicit completed `task` fixture without child-session metadata, so the plugin gate is stricter than the ingest shape. Fix: either add a fixture/contract proving completed subagents always have one of these signals, or add a last-resort non-partial completion signal. Safer options: accept a terminal `step-finish` part on the child’s last assistant message; flush when the child disappears from `client.session.children(parent)`; or use an absolute elapsed-time bound only when the child’s latest message is assistant-role and all tool parts are terminal. The important bit is to fail closed while streaming, but not defer a completed child forever.

## Non-blocking

1. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:226` / `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:228` - Every parent idle fetches and scans the full transcript of every unflushed child until it completes. That is `O(parent idles * pending children * child transcript size)`, plus the SDK call cost, and it is unbounded for a child that remains pending because one completion marker is missing. Fix: cache a per-child scan key such as last message id + part count + terminal-tool count + final marker, and only rescan when that key changes; also short-circuit children once they have all tools terminal and only lack the final marker, using the bounded fallback above.

2. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:92` / `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:251` - The fail-closed classify behavior does avoid the earlier double-ingest risk, and a one-off `session.get` failure is not sticky: later `session.idle` calls re-run `classify` and can start the top-level session once `get` recovers. The remaining tradeoff is that an endpoint-specific `session.get` regression drops a top-level session even if `session.messages` would have worked. That may be acceptable, but it should be an explicit choice. Fix: if you want a safety valve without reopening child double-ingest, only consider a delayed top-level fallback after checking known parent `children`/terminal-task references and after a real elapsed window, not after a raw idle count.

## Nit

1. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:142` - The fail-safe write path is now in the right order: `written.set(targetFile, seen)` stores the set object, but failed keys are not added until after `appendFileSync` succeeds, and `flushedChildren` is set only when `writeMessages(...)` returns true. No change requested here; this is just calling out that the prior append-failure suppression is fixed.

Additional checks: I do not see a throw path escaping the OpenCode event handler in the changed plugin code (`classify`, `fetchMessages`, `writeMessages`, `flushSubagents`, `runKcap`, and the outer handler are all fail-safe). I also do not see new C# reflection/dynamic usage in `f3fdbac`; the NativeAOT risk from this commit looks unchanged.

_🤖 Reviewed by Codex_

realtonyyoung:
Thanks — the BLOCKING + non-blocking addressed in `67b6b25`:

- **BLOCKING (permanent-defer if no completion marker):** added a third, reliable signal — a `step-finish` part on the child's final assistant turn (OpenCode emits one to close a completed step, present even when `info.finish`/`time.completed` are absent) — plus a **last-resort** flush once a child has been **structurally at rest** (every tool terminal AND last message assistant) past `LAST_RESORT_MS` (2 min). So a completed child can never be deferred forever, while a mid-stream child (a non-terminal tool, or a non-assistant last message) is still never written. The new `structurallyAtRest` helper gates both the marker check and the last-resort window.
- **Non-blocking (unbounded re-fetch):** bounded by the last-resort window — a pending child is flushed within 2 min max, after which `flushedChildren` short-circuits before the re-fetch. (Skipped the scan-key cache: the SDK fetch dominates the in-memory scan, and the fetch count is now bounded.)
- **Non-blocking (classify fail-closed tradeoff):** confirmed intentional — a misfiled child can't be cleanly un-ingested, so we fail closed; a one-off `session.get` failure is non-sticky (a later idle re-runs classify once get recovers). An endpoint-specific get-only regression dropping a top-level session is the accepted cost vs. reopening child double-ingest.
- Nit: thanks — the write-order fix stands.

Full CLI unit suite 1677 green; AOT publish clean.

realtonyyoung:
Static re-review of `67b6b25` only. I did not run tests, builds, or publishes.

## BLOCKING

None. I do not see a concrete remaining correctness defect in this iteration, and I think this is ready to merge.

## Non-blocking

1. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:245` / `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:250` - The only residual edge I can still construct is the one called out in the prompt: if every reliable completion signal is absent (`info.finish`, `info.time.completed`, `step-finish`, and parent `task` metadata), a long pure-text child can satisfy `structurallyAtRest` as soon as its last message is assistant and it has no non-terminal tools. After 2 minutes, the last-resort path may flush that snapshot even if text is still streaming. I could not confirm from upstream schema/docs in this pass that `step-finish` is guaranteed on every completed SST OpenCode subagent, so I would not rely on that as a formally proven invariant; however, this now requires all normal completion signals to be absent simultaneously, so I do not think it should block the PR. Concrete hardening if desired: record a lightweight stability key for last-resort candidates, for example last message id + part count + terminal-tool count + final text/reasoning lengths, and only allow the 2-minute fallback after the key has been unchanged across two observations. That preserves the “never defer forever” guarantee without reopening the partial-text write for slow markerless text streams.

2. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:248` / `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:256` - `childFirstSeen` is cleaned on successful flush, and it is only populated for structurally-at-rest children, so normal completed children do not leak. The only way it can grow during one long parent session is if many children become structurally at rest but never successfully write or stop being returned by `client.session.children(parent)` before the fallback fires. That is bounded by pathological/unflushed children rather than ordinary idles, so I would not block on it. If you want belt-and-suspenders cleanup, delete map entries for child ids not returned by the latest `children(parent)` scan or after a failed write eventually succeeds.

## Nit

1. `src/Capacitor.Cli.Core/OpenCode/OpenCodeExtensionInstaller.cs:244` - `done.has(cid)` correctly skips the child-completeness gate, but it cannot short-circuit the child fetch because the plugin still needs the child messages to write the nested JSONL. No fix needed; the current ordering is reasonable.

Additional checks: the event/control flow is still fail-safe (`runKcap`, `classify`, `fetchMessages`, `writeMessages`, `flushSubagents`, and the outer event handler all swallow failures). I do not see new C# reflection/dynamic usage in this commit, so the AOT risk looks unchanged.

_🤖 Reviewed by Codex_

realtonyyoung:
Thanks — taking the "ready to merge" verdict, and folding in both non-blocking hardenings in `68d0271` so the last-resort path is provably safe rather than relying on an unproven invariant:

**Non-blocking #1 (last-resort vs. streaming text).** Implemented your suggested stability key. `childFirstSeen` now holds `{ key, since }` where `key = lastMessageId | totalPartCount | terminalToolCount | finalMessageTextLength` (`stabilityKey`). The 2-minute fallback only fires once the key has been **unchanged for the whole window** — any new part, a tool turning terminal, or growing text resets `since`. So a child streaming markerless text keeps resetting the window and is never flushed mid-stream, even with `info.finish` / `info.time.completed` / `step-finish` all absent. "Never defer forever" is preserved (a genuinely-at-rest markerless child still flushes after the window).

**Non-blocking #2 (`childFirstSeen` growth).** Added the belt-and-suspenders prune: after each `session.children` scan, timers for child ids no longer returned are dropped. A transient drop only resets a timer on reappearance (longer defer) — never an early flush — so the map stays bounded across a long parent session.

**Nit #1.** Agreed, no change — `done.has(cid)` correctly skips the completeness gate but the child fetch is still needed to write the nested JSONL; current ordering stands.

CLI unit 1677/1677, AOT clean (no IL3050/IL2026), control flow still fail-safe.

Issue #182, opened 2026-06-26
Proactively refresh auth tokens in the daemon to reduce forced re-logins


alexeyzimarev:
## Problem

Auth tokens are refreshed **lazily** — `TokenStore.GetValidTokensAsync()` (`src/Capacitor.Cli.Core/Auth/TokenStore.cs:123`) only refreshes when the access token is already expired, on the next call that needs it. Nothing keeps the refresh credential warm ahead of time.

This means that after an idle period, a user hits a 401 on the next hook and sees:

```
Stop hook error: Failed with non-blocking status code: Authentication token has expired. Run 'kcap login' to re-authenticate.
HTTP 401
```

WorkOS refresh tokens have both an **inactivity** window and an **absolute** session lifetime (configured server-side). When the inactivity window lapses, the refresh fails and the user must re-run `kcap login`.

## Proposal

Have the **daemon** refresh tokens **proactively** on a timer — refreshing *ahead* of expiry rather than reactively after it. For WorkOS's sliding inactivity window, periodic refresh keeps the session alive for as long as the daemon is running, pushing forced re-logins out to the absolute session max instead of the (shorter) inactivity timeout.

The daemon already loads tokens on demand in several places — `ServerConnection.cs:72,646`, `AgentOrchestrator.cs:840` — and already runs periodic loops we can model this on (`AgentOrchestrator.RunDaemonHeartbeatLoopAsync`, `_daemonHeartbeat = new(TimeSpan.FromSeconds(7))`). Add a low-frequency refresh tick that calls `GetValidTokensAsync()` when the token is within some window of expiry.

## Design notes / constraints

- **Go through the existing cross-process lock.** Refresh must use `RefreshWithCrossProcessLockAsync` (`TokenStore.cs:155`) so the daemon doesn't race hooks / watcher / MCP and clobber a rotated WorkOS refresh token. (The lock path is already what `GetValidTokensAsync` uses.)
- **Refresh within a window, don't hammer.** Only refresh when the token is within, say, N minutes of `ExpiresAt` — not on every tick.
- **WorkOS rotation reuse hazard.** WorkOS rotates the refresh token on each successful use; a refresh whose response is lost after the server rotated leaves the persisted token stale. This hazard already exists; just don't make it worse (the file lock + re-read-under-lock is the mitigation).
[...]

PR #183, opened 2026-06-26
Improve auth-expiry experience: retry refresh + non-nagging hooks

alexeyzimarev:
## Why

A user hit this on a Stop hook after their auth lapsed:

```
Stop hook error: Failed with non-blocking status code: Authentication token has expired. Run 'kcap login' to re-authenticate.
HTTP 401
```

Two separate problems were behind it, addressed in two commits.

## 1. Retry token refresh on transient failures

`TokenStore` refreshes tokens silently before every authenticated call, but the two refresh HTTP calls used a bare single-shot `PostAsync` and swallowed every exception to `null`. So a momentary transport blip (DNS stutter, connection reset, brief server slowness) degraded **instantly** to "Authentication token has expired. Run 'kcap login'" — even when the refresh credential was still perfectly valid.

Both calls now go through the existing `PostWithRetryAsync` helper with a short **5s** budget:
- `RefreshGitHubAsync` → `/auth/refresh`
- `RefreshWorkOSAsync` → WorkOS `/user_management/authenticate`

`PostWithRetryAsync` retries **only transport failures, never non-success HTTP responses**, so a genuinely-expired refresh token still returns fast (`400`/`401` → `null`, no pointless retries). The short budget keeps a hook from blocking on the default 30s when the server is truly down. Adding retry introduces no new failure mode for WorkOS's rotate-on-use refresh tokens — that reuse window already existed across separate refresh calls, and the cross-process lock + re-read-under-lock remains the mitigation.

## 2. Make hook auth-expiry non-nagging and non-alarming

When auth has genuinely lapsed (refresh credential dead → re-login truly required), the Claude hook used to POST a request it knew would `401`, print `HTTP 401`, and exit `1` — which Claude Code renders as a red **"Stop hook error"** banner **on every turn** until the user re-authenticates.

Recording is best-effort, so this now:
- **Skips the POST** when auth is expired/missing (no point — it would 401).
- **Exits 0** — no red banner.
- **Nudges the user to re-login only on `session-start`** (once per session), staying silent on the high-frequency `stop` / `notification` / `subagent-*` events.

Mechanism: `HttpClientExtensions.CreateClientWithAuthStatusAsync` now reports an `AuthStatus` (`Ok` / `NoAuthRequired` / `Expired` / `NotAuthenticated`) instead of swallowing it. The existing `CreateAuthenticatedClientAsync` is preserved as a thin wrapper (same stderr behaviour for interactive commands), so non-hook callers are unaffected.

The session-start nudge is emitted as a **`systemMessage` JSON object** on stdout, not stderr. This matters: on exit 0 Claude Code hides hook stderr entirely, and plain `session-start` stdout is injected into the *model's* context — `systemMessage` is the one channel that shows the notice to the **user** without leaking into context. (Verified against the Claude Code hooks docs.)

## Scope

Hook change applied to the **Claude Code** path (the one reported). The other agent hooks (Codex, Cursor, Gemini, …) share the same `CreateAuthenticatedClientAsync` shape and can adopt `CreateClientWithAuthStatusAsync` as a fast follow-up.

## Follow-up

Proactive background token refresh in the daemon is tracked separately in #182.

## Testing

- `dotnet build` — clean.
- AOT publish — no IL3050/IL2026 warnings.
- Unit tests — pass.
- README: no change (no user-facing CLI surface change — internal behaviour only).

### DIFF
--- a/src/Capacitor.Cli.Core/Auth/TokenStore.cs
+++ b/src/Capacitor.Cli.Core/Auth/TokenStore.cs
@@ -216,6 +216,14 @@ public static class TokenStore {
         return DateTimeOffset.UtcNow.AddMinutes(5);
     }
 
+    // Short retry budget for the refresh HTTP call. A bare single-shot POST turned any
+    // transient blip (DNS stutter, connection reset, brief server slowness) into a hard
+    // "token expired — run kcap login", even though the refresh credential was still valid.
+    // PostWithRetryAsync retries only transport failures, never non-success responses — so a
+    // genuinely-expired refresh token still returns fast (400/401 → null, no pointless retries).
+    // Kept short so a hook never blocks for the default 30s budget when the server is down.
+    static readonly TimeSpan RefreshRetryBudget = TimeSpan.FromSeconds(5);
+
     static async Task<StoredTokens?> RefreshGitHubAsync(StoredTokens tokens) {
         var       baseUrl = AppConfig.ResolvedServerUrl ?? Environment.GetEnvironmentVariable("KCAP_URL") ?? "http://localhost:5108";
         using var http    = new HttpClient();
@@ -227,7 +235,7 @@ public static class TokenStore {
         var payload = new StringContent(requestBody, System.Text.Encoding.UTF8, "application/json");
 
         try {
-            var response = await http.PostAsync($"{baseUrl}/auth/refresh", payload);
+            var response = await http.PostWithRetryAsync($"{baseUrl}/auth/refresh", payload, RefreshRetryBudget);
 
             if (!response.IsSuccessStatusCode) {
                 return null;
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
Addressed the Qodo finding ("Refresh can exit process") in 75c711f.

`RefreshGitHubAsync` now validates the URL with `IsAcceptableUrl` before calling `PostWithRetryAsync`, returning `null` on a scheme-less `server_url` instead of letting `EnsureAbsolute` → `Environment.Exit(2)` terminate the process. This preserves the graceful-failure contract for the daemon/background callers that reach refresh via `GetValidTokensAsync` (`ServerConnection`, `StatusCommand`). The hook *entry* paths still `EnsureAbsolute`-and-exit by design; WorkOS refresh was never affected (hardcoded absolute URL).

Verified: build clean, no IL3050/IL2026 warnings, 1759/1759 unit tests pass.

PR #192, opened 2026-06-26
kcap import --opencode: historical OpenCode import from SQLite (AI-1020)

realtonyyoung:
## What

Adds `kcap import --opencode` — historical import of OpenCode sessions (including subagents) from OpenCode's SQLite database, closing [AI-1020](https://linear.app/kurrent/issue/AI-1020). OpenCode previously reached the live surfaces (status, setup wizard, plugin, watcher) but not import.

## Why this shape

OpenCode's only durable history is `~/.local/share/opencode/opencode.db` (no per-session JSONL). The importer reconstructs the exact `{info,parts}` lines the live plugin streams by merging each row's key columns back onto its `data` JSON, so the server's existing `opencode` normalizer ingests import output identically — **no server changes**.

Three constraints shaped the design (all verified against `kcap-server@main` before coding — see the spec's Task 0):

- **Live vs import line spaces are incompatible** (live emits many snapshot lines per message; import emits one final-state line). So there is no line-by-line resume.
- **The server HWM advances on transcript ingest, not `session-end`**, and exposes **no completeness signal**. So a naive "any watermark → skip" can't tell complete from partial.
- **Transcript ingest dedupes by canonical `prt_`-id, independent of line number.**

The result:

- **Completeness is tracked client-side** in a per-machine, per-server import ledger (`~/.cache/kcap/opencode-imported.json`), written only after `session-end` succeeds.
- **Classification:** ledger hit → `AlreadyLoaded` (skip); else `New` (no watermark) or `Partial`-repair (watermark → replay the full transcript with line numbers offset **above** the HWM, deduped by `prt_`-id).
- **Strict send:** an opt-in `failOnError` on `SessionImporter.SendTranscriptBatches` (defaulted; peers unchanged) aborts before any terminal lifecycle POST, so a partial import stays unrecorded and is repaired on re-run.
- **Subagents:** child sessions import as subagents of their parent (`/hooks/subagent-*`) before the parent `session-end`, matching live nesting.

## Surface

- New: `OpenCodeDb` (read-only SQLite reader + line reconstruction + server-parity importable predicate), `OpenCodeImportLedger`, `OpenCodeImportSource`.
- `Microsoft.Data.Sqlite` + `SQLitePCLRaw.bundle_e_sqlite3` added to the **CLI project only** (kept out of Core so the AOT daemon never gets the native bundle).
- `--opencode` wired into `VendorSelection` + the import source list.
- Docs: `help-import.txt`, `help-usage.txt`, `README.md` (corrected the prior "OpenCode capture is live-only / no import" claim).

## Verification

- Unit **1716/1716**, integration **43/43** — 0 failures. New tests cover reconstruction, ordering, the importable predicate (server-parity), ledger round-trip, classification states, lifecycle order, strict-send abort, subagent routing, repair-above-HWM, batch-2-failure→rerun-repair, and ledger skip-on-rerun.
- AOT publish clean (zero IL2026/IL3050) for **both** CLI and daemon; daemon carries no SQLite reference. Binary ~18M.

## Known caveat

The ledger trusts local state: a session deleted server-side after being recorded would be skipped on re-run. Mitigated by keying on server URL + recording only after `session-end`. A `--reimport`/`--force` bypass is the natural follow-up.

Design spec + implementation plan: `docs/superpowers/specs/2026-06-26-opencode-import-design.md`, `docs/superpowers/plans/2026-06-26-opencode-import.md` (also attached to AI-1020).


realtonyyoung:
## Code-review disposition (Codex)

**Fixed inline (CLI):**
- **Discovery crash guard** — a corrupt/schema-drifted `opencode.db` now warns and skips OpenCode instead of throwing out of `DiscoverAsync` and killing the whole `import` run; `QuerySessions` skips malformed rows.
- **Cancellation** — `OperationCanceledException` now propagates out of the import catches instead of being swallowed as `Failed`.
- **Overflow** — `checked()` on the repair line-number offsets (parent + child).
- **Ledger key strengthened (was a Major)** — the ledger value is now a **SHA-256 content fingerprint** over the parent transcript **+** children, replacing the bare parent line-count. A same-line-count mutation (tool part completing, in-place text edit, changed/added child) now invalidates the skip and re-imports.

**Deferred — server-repo follow-ups (need kcap-server changes; this PR is CLI-only):**
- **#1 — `/hooks/transcript` returns `200` on a swallowed per-event write failure.** The strict sender's HTTP-status check can't see it, so the ledger may record a session whose tail didn't persist. The *live* watcher recovers (re-sends from the un-advanced HWM); the ledger makes import's skip permanent. `processed` isn't a reliable client-side signal (non-importable lines emit zero events). **Fix:** server-side strict transcript response (non-2xx on any write failure).
- **#3 — `/hooks/subagent-start|stop` return `200` even if the lifecycle append fails.** Shared exposure with the live path; the ledger makes the parent skip permanent. **Fix:** server returns 500 on subagent-lifecycle write failure.

Both are documented in the design spec's Risks section. A Linear follow-up for the kcap-server work should be filed (the Linear connector is currently disconnected on my side).

**Verified clean** by Codex: `SynthesizeLines` (LEFT JOIN, empty-part messages, reader lifetime, key merge/order), `IsImportRelevantLine` vs the server normalizer, wiring, and CLI-only SQLite placement.

realtonyyoung:
Final pre-merge review pass, reading current code only. No blockers found; I would not hold merge on correctness/AOT grounds.

Findings:

- Minor: `src/Capacitor.Cli/Commands/OpenCodeImportSource.cs` contains a literal NUL byte in the child fingerprint separator (`Feed("\0child:" + child.Id)` appears as an actual NUL in the source). Runtime hashing is fine, but tools like `rg` now treat the `.cs` file as binary. Concrete fix: replace the literal byte with an escaped source representation, e.g. `"\0child:"` or `"\u0000child:"`.

- Minor: cancellation is still not fully propagated through every helper path. The parent/children import catches now rethrow `OperationCanceledException`, but `PostHookAsync` catches all and returns `false`, the watermark-probe catch in `ClassifyAsync` would turn cancellation into `ProbeError`, and `SendTranscriptBatches` still has no `CancellationToken`. This is mostly pre-existing because routed imports currently pass `CancellationToken.None`, but the clean fix is to rethrow OCE in helper catches and eventually thread a token through the shared sender.

- Nit: the accepted server limitations are documented honestly in the spec, but the risk section does not name `AI-1023`. Add the issue id there if that traceability matters pre-merge.

Verified:

- Discovery guard landed correctly: DB open/scan is inside try/catch, OCE is rethrown, corrupt/schema-drift failures warn and return `[]`; malformed session rows are skipped.
- Overflow no longer wraps: parent and child offsets use `checked`. Parent overflow bubbles to the routed import wrapper as `Failed`; child overflow is caught by the child import wrapper before parent `session-end`.
- Ledger fingerprint is deterministic enough: parent lines are drained before child queries, children are ordered by `QueryChildren`, child content is hashed, `IncrementalHash` is disposed, and `WithFingerprint` copies `SourceMeta` rather than mutating shared state.
- The two server-side limitations are now explicitly accepted in the spec, and code comments describe strict send as HTTP-failure abort-before-end rather than a full persistence guarantee.

realtonyyoung:
## Final-review disposition (commit 3898ebe)

All three items from the pre-merge review actioned — none were blockers:

- **NUL byte in source (Minor)** — confirmed: line 294 had a literal `0x00` (so `rg`/grep treated the `.cs` as binary). Replaced with a `\u0000` escape sequence — the runtime fingerprint separator is unchanged (still a NUL, collision-proof against JSON text), but the source is now plain text.
- **Cancellation in helpers (Minor)** — `PostHookAsync` and the `ClassifyAsync` watermark-probe catch now rethrow `OperationCanceledException` before their generic catches, so cancellation isn't masked as a hook-failure / `ProbeError`. (Threading a `CancellationToken` through the shared `SendTranscriptBatches` is left as pre-existing — routed sources pass `CancellationToken.None` today.)
- **AI-1023 traceability (Nit)** — the spec's two server-side limitation notes now name **AI-1023**.

OpenCode unit (32) + integration (6) green; AOT build clean. CI re-running on the new commit.

PR #193, opened 2026-06-27
[AI-1030] Codex rollout idle-timeout session-end

realtonyyoung:
## Summary

Codex has no session-end hook, so a Codex session ends only when the `kcap watch` parent-exit watchdog sees the parent `codex` process exit. The **Codex desktop app** runs one long-lived `codex app-server` process that is the `--parent-pid` for every conversation's watcher, so that PID never exits per-conversation and Codex sessions stay "active" forever.

This adds a **rollout-file idle-timeout fallback** to the watcher:

- `ResolveCodexIdleTimeout` reads `KCAP_CODEX_IDLE_MINUTES` (default **60**; invalid/non-positive → default).
- `WatchState.LastActivityAt` tracks the last time new lines were appended; updated in `DrainNewLines`.
- `ShouldEndOnIdle` (pure, unit-tested) — gated to **vendor codex, session watchers only, threshold-reached only**.
- The main watch loop ends the session on idle, reusing `PostSessionEndOnParentExitAsync` with a new `reason` parameter so it POSTs `/hooks/session-end/codex` with `reason: idle_timeout` (parent-exit path still posts `parent_exited`).

Self-correcting: Codex fires a `Stop` hook per turn that re-spawns the watcher (idempotent `EnsureWatcherRunning`), and the server reactivates a session when new transcript events arrive, so resuming a conversation brings it back.

**Pairs with kcap-server PR for AI-1030** (adds the `idle_timeout` `SessionEndReason` + a reactivation-aware session-end idempotency guard). Deploy **server before CLI**; older servers degrade the unknown reason to `Other` harmlessly.

## Test Plan

- [x] Unit: `ResolveCodexIdleTimeout` (env parsing incl. non-positive) + `ShouldEndOnIdle` (codex/non-codex, subagent, below-threshold, strict `>` boundary) — full unit suite 1850/1850.
- [x] Integration: `WatcherParentExitPostTests` asserts the POST to `/hooks/session-end/codex` carries `reason: idle_timeout` — 8/8.
- [x] AOT publish clean (no IL2026/IL3050).
- [ ] Real Codex desktop E2E (set `KCAP_CODEX_IDLE_MINUTES=2`: idle-end fires, then resume re-spawns watcher + reactivates).

### DIFF
--- a/src/Capacitor.Cli/Commands/WatchCommand.cs
+++ b/src/Capacitor.Cli/Commands/WatchCommand.cs
@@ -534,6 +534,28 @@ static partial class WatchCommand {
             ? TimeSpan.FromMinutes(minutes)
             : DefaultCodexIdleTimeout;
 
+    /// <summary>
+    /// Whether the watcher should self-terminate and POST session-end because the
+    /// Codex rollout file has gone idle. Pure so the policy is unit-testable.
+    /// Gated to: vendor codex (only vendor whose parent-exit watchdog can't fire
+    /// per-conversation — the desktop app's shared app-server never exits per
+    /// session), session watchers (not subagents), and threshold-reached sessions
+    /// (below-threshold short-lived sessions have no server session to end). Uses
+    /// strictly-greater-than so the boundary tick is not yet considered idle.
+    /// </summary>
+    internal static bool ShouldEndOnIdle(
+            string         vendor,
+            bool           isSessionWatcher,
+            bool           thresholdReached,
+            DateTimeOffset lastActivityAt,
+            DateTimeOffset now,
+            TimeSpan       idleTimeout
+        ) =>
+        vendor == "codex"
+        && isSessionWatcher
+        && thresholdReached
+        && now - lastActivityAt > idleTimeout;
+
     internal static async Task PostSessionEndOnParentExitAsync(
             string             baseUrl,
             string             sessionId,
--- a/src/Capacitor.Cli/Commands/WatchCommand.cs
+++ b/src/Capacitor.Cli/Commands/WatchCommand.cs
@@ -512,6 +512,28 @@ static partial class WatchCommand {
     /// </summary>
     static readonly TimeSpan ParentExitPostBudget = TimeSpan.FromSeconds(10);
 
+    /// <summary>
+    /// Codex has no session-end hook and the desktop app's single long-lived
+    /// `codex app-server` process is the watchdog's parent PID for EVERY
+    /// conversation, so the parent-exit watchdog never fires per-conversation —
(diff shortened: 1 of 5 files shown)


src/Capacitor.Cli/Commands/WatchCommand.cs
```diff
@@ -307,6 +310,20 @@ public static async Task<int> RunWatch(
                     await ScanOpenCodeSubagents(baseUrl, sessionId, transcriptPath, seenSubagents, cts.Token);
                 }
 
+                if (ShouldEndOnIdle(
```
realtonyyoung, line 313: [P1] This can still false-end an active Codex turn because the idle signal is only “no new rollout lines.” Codex does append durable response items such as reasoning/function-call/tool-output/token events, so ordinary turns keep moving, but a single long-running segment can legitimately be silent for longer than the default: for example, after a function_call line, a shell command with no output for >60m will not append the function_call_output line until it exits. In that window this cancels the watcher and POSTs `reason: "idle_timeout"` even though Codex is still working. Please gate the idle end on an explicit turn/tool lifecycle marker from the Codex hooks (for example mark active on UserPromptSubmit/PreToolUse and clear/refresh on Stop/PostToolUse), or otherwise require a known turn-ended state before applying the rollout-idle timeout.

realtonyyoung (commented):
Reviewed the Codex idle-timeout rollout path. I left one P1 inline on the idle check: a long active Codex turn/tool segment can be rollout-silent for more than 60 minutes and be ended as `idle_timeout`. I also checked the env parsing/docs, non-Codex and subagent gating, parent-exit precedence, cancellation/final-drain ordering, and the disconnected-SignalR placement; no other findings from this pass.

src/Capacitor.Cli/Commands/WatchCommand.cs
realtonyyoung: Fixed in 2f1af61. The watcher now tracks outstanding Codex tool calls (`function_call`/`custom_tool_call` add a `call_id`; the matching `*_output` removes it) via a pure `UpdateCodexPendingToolCalls` helper, and `ShouldEndOnIdle` now takes a `toolInFlight` argument so idle-end is suppressed while any tool call is outstanding — so a long-running silent shell exec (function_call with no output for >60m) no longer false-ends the session. No hard ceiling by design: if the process truly dies mid-tool the parent-exit watchdog ends the session (for the desktop app the shared app-server dying ends all sessions), and a hung-but-alive tool is a real in-flight turn the user may still cancel. Tracked watcher-side (no hook changes); validated against a real 432-event rollout (set returns to empty between turns) plus unit tests.

PR #196, opened 2026-06-29
[AI-997] kcap recap --per-turn/--get-turn + MCP get_turn tool

realtonyyoung:
Part of **AI-995** (per-turn summaries), phase 2 (AI-997) — the CLI/MCP surface for the new server-side per-turn endpoints (`GET /api/sessions/{id}/turns` and `/turns/{index}`, added in the companion server PR).

## Changes
- **`kcap recap --per-turn <sessionId>`** — compact per-turn index (turn #, prompt excerpt, tool names, file count, tokens, time range), one block per turn.
- **`kcap recap --get-turn <sessionId> <N>`** — full event transcript for a single turn.
- **MCP `get_turn(session_id, turn_index)`** tool on the `kcap mcp sessions` server — lets eval judges / agents drill into one turn.
- README + `help-recap.txt` updated for both flags and the MCP tool.

## Notes for review
- Turn JSON is parsed with `JsonDocument` (not a typed DTO): the CLI has no `TurnClosed` type and `JsonDocument` is AOT-safe. The server serializes snake_case (`turn_index`, `total_tokens`, trace entries use `is_error`), and the parsing keys match.
- `--get-turn` is declared as a value-flag in `ResolveSessionId` so the turn index isn't mistaken for the session id.
- AOT publish verified: **0 IL2026/IL3050 warnings**.

### DIFF
--- a/src/Capacitor.Cli/Commands/RecapCommand.cs
+++ b/src/Capacitor.Cli/Commands/RecapCommand.cs
@@ -265,4 +265,196 @@ static class RecapCommand {
             _             => ""
         };
     }
+
+    /// <summary>
+    /// `kcap recap --per-turn &lt;sessionId&gt;` — prints a compact per-turn index
+    /// (turn #, prompt excerpt, tool names, file count, token count, time range),
+    /// one block per turn. Parses the server's snake_case JSON with JsonDocument
+    /// (the CLI has no TurnClosed DTO and JsonDocument is AOT-safe).
+    /// </summary>
+    public static async Task<int> HandlePerTurnRecap(string baseUrl, string sessionId) {
+        using var httpClient = await HttpClientExtensions.CreateAuthenticatedClientAsync();
+
+        HttpResponseMessage resp;
+
+        try {
+            resp = await httpClient.GetWithRetryAsync($"{baseUrl}/api/sessions/{sessionId}/turns");
+        } catch (HttpRequestException ex) {
+            HttpClientExtensions.WriteUnreachableError(baseUrl, ex);
+
+            return 1;
+        }
+
+        if (await HttpClientExtensions.HandleUnauthorizedAsync(resp)) {
+            return 1;
+        }
+
+        if (resp.StatusCode == System.Net.HttpStatusCode.NotFound) {
+            Console.Error.WriteLine($"Session not found: {sessionId}");
+
+            return 1;
+        }
+
+        if (!resp.IsSuccessStatusCode) {
+            Console.Error.WriteLine($"HTTP {(int)resp.StatusCode}");
+
+            return 1;
+        }
+
+        var json = await resp.Content.ReadAsStringAsync();
+
+        using var doc = JsonDocument.Parse(json);
+
+        if (doc.RootElement.ValueKind != JsonValueKind.Array || doc.RootElement.GetArrayLength() == 0) {
+            await Console.Out.WriteLineAsync("No turns found. The session may still be active or has no recorded turns.");
+
+            return 0;
+        }
+
+        await Console.Out.WriteLineAsync($"# Turns for session {sessionId}");
+        await Console.Out.WriteLineAsync();
+
+        foreach (var turn in doc.RootElement.EnumerateArray()) {
+            var turnIndex  = turn.TryGetProperty("turn_index", out var ti) ? ti.GetInt32() : -1;
+            var userPrompt = turn.TryGetProperty("user_prompt", out var up) ? up.GetString() ?? "" : "";
+            var prompt     = userPrompt.Length == 0
+                ? $"(turn {turnIndex})"
+                : userPrompt.Length > 80 ? userPrompt[..80] + "…" : userPrompt;
+
+            var tools = "—";
+
+            if (turn.TryGetProperty("tools", out var toolsEl) && toolsEl.ValueKind == JsonValueKind.Array) {
+                var names = new List<string>();
+
+                foreach (var t in toolsEl.EnumerateArray()) {
+                    if (t.TryGetProperty("name", out var n) && n.GetString() is { Length: > 0 } nm) {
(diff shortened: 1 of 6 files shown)


src/Capacitor.Cli/Program.cs
```diff
@@ -101,24 +101,48 @@
...
+        var useGetTurn = args.Contains("--get-turn");
 
         if (useRepo) {
             return await RecapCommand.HandleRepoRecap(baseUrl!);
         }
 
-        var recapSessionId = ResolveSessionId(args);
+        // --get-turn takes a value (the turn index), so declare it as a value flag
+        // or ResolveSessionId would mistake the index for the sessionId.
+        var recapSessionId = ResolveSessionId(args, valueFlags: ["--get-turn"]);
```
realtonyyoung, line 116: [Important] This makes --get-turn treat the token immediately after the flag as the turn index, but the README and the per-turn hint document kcap recap --get-turn {sessionId} <N>. With args like [recap, --get-turn, abc123, 2], ResolveSessionId skips abc123 as the flag value and returns 2 as the session id, then int.TryParse(abc123) fails. Either document/enforce --get-turn <N> [sessionId] everywhere, or parse the documented --get-turn <sessionId> <N> form before ResolveSessionId.

src/Capacitor.Cli/Commands/RecapCommand.cs
```diff
@@ -265,4 +265,196 @@ static string GetLanguageHint(string filePath) {
...
+        if (!doc.RootElement.TryGetProperty("trace", out var trace) || trace.ValueKind != JsonValueKind.Array) {
+            await Console.Out.WriteLineAsync("No trace available for this turn.");
+
+            return 0;
+        }
+
+        await Console.Out.WriteLineAsync($"# Turn {turnIndex} — session {sessionId}");
+        await Console.Out.WriteLineAsync();
+
+        foreach (var entry in trace.EnumerateArray()) {
```
realtonyyoung, line 405: [Minor] The turn-detail trace includes agent_id/agent_type for subagent entries, but this renderer never displays or groups by those fields. For a root turn that spawned a subagent, the CLI flattens root and subagent messages/tools into one stream and loses the attribution that existing --full recap preserves with agent headers. Consider tracking the current agent_id while enumerating trace and emitting an agent header or prefix when it changes.

realtonyyoung:
Reviewed as one logical AI-995 change with kcap-server#864. Left 2 inline comments: [Important] the documented kcap recap --get-turn <sessionId> <N> order does not match the parser, and [Minor] single-turn CLI rendering drops subagent attribution from trace entries. I did not run tests or builds.

src/Capacitor.Cli/Program.cs
realtonyyoung: Fixed in 6863bc0 — aligned the docs with the parser rather than changing it: `--get-turn` takes the turn number as its value, and the session id is the usual positional (or env). README + the per-turn hint now show `kcap recap --get-turn <N> <sessionId>` (and note `kcap recap <sessionId> --get-turn <N>` works too). The parser already handles both orderings correctly; only the old `--get-turn <sessionId> <N>` doc form was wrong.

src/Capacitor.Cli/Commands/RecapCommand.cs
realtonyyoung: Fixed in 6863bc0 — HandleGetTurn now tracks the active agent while enumerating the trace and emits a `### Subagent: {type} ({id})` header when agent_id changes (plus a `### (main session)` marker on return), instead of flattening root + subagent into one stream. Pairs with the server-side fix that makes the trace's agent_id the bare id rather than the full stream name.

src/Capacitor.Cli/Commands/RecapCommand.cs
```diff
@@ -265,4 +265,225 @@ static string GetLanguageHint(string filePath) {
...
+            await Console.Out.WriteLineAsync();
+        }
+
+        Console.Error.WriteLine($"Use `kcap recap --get-turn <N> {sessionId}` for full turn detail.");
+
+        return 0;
+    }
+
+    /// <summary>
+    /// `kcap recap --get-turn &lt;sessionId&gt; &lt;turnIndex&gt;` — prints the formatted
```
realtonyyoung: [Minor] This XML summary still documents the old positional order (`--get-turn <sessionId> <turnIndex>`). The README, help text, usage error, parser, and per-turn hint now all use the corrected `--get-turn <N> [sessionId]` shape, so this last in-code doc should be realigned too.

realtonyyoung:
Re-review complete. Left 1 inline comment: [Minor] one XML summary still documents the old `--get-turn <sessionId> <turnIndex>` order. The README, help text, usage error, parser behavior, and subagent header grouping otherwise line up with the fixed server shape. I did not run tests or builds.

src/Capacitor.Cli/Commands/RecapCommand.cs
realtonyyoung: Fixed in a145a2b — realigned the HandleGetTurn XML summary to `--get-turn <N> [sessionId]`, so the in-code doc now matches the README, help text, usage error, parser, and per-turn hint.

realtonyyoung:
Also fixed the failing `Build and test` check: `Tools_list_returns_three_tools_with_correct_names` asserted exactly 3 MCP tools, but this PR adds `get_turn` (4th). Updated the test to expect 4 tools and assert `get_turn` is present (a145a2b).

realtonyyoung:
Targeted round-2 re-review complete. No findings. The HandleGetTurn XML summary now matches the README/help/usage/parser shape for --get-turn <N> [sessionId], including positional-before/after and env-session cases, and the MCP tools-list test now matches BuildToolsList with four tools including get_turn. I did not run tests, builds, or AOT publishes.

PR #197, opened 2026-06-29
Load SQLite native on demand for kcap import --opencode

realtonyyoung:
## Problem

`kcap import --opencode` (and `--all`) crashed with `DllNotFoundException: Unable to load shared library 'e_sqlite3'` at `OpenCodeDb` → `SqliteConnection`.

Root cause: the AOT `kcap` binary does **not** statically link SQLitePCLRaw's native engine (NativeAOT emits a dynamic `libe_sqlite3.dylib`/`.so`/`.dll` and `dlopen`s it), and — unlike `libpty_shim` — the npm platform packages never shipped it. SQLite is only needed to read OpenCode's `opencode.db` during import.

Two surfaces, one cause:
- `kcap import --opencode`: `DiscoverAsync` swallows the load failure → "Found 0 opencode sessions".
- `kcap import --all`: `ClassifyAsync` opened the db unconditionally (runs for every vendor) → unhandled crash.

## Approach — download on demand

Rather than bundle ~1.6 MB into every install for a feature only OpenCode import uses, fetch the native lazily.

`SqliteNativeResolver` installs a `DllImport` resolver for `e_sqlite3` (registered from `OpenCodeDb`'s static ctor, before the first SQLite P/Invoke) that:

1. **Tries default OS resolution first** — dev/standalone builds with the lib beside the binary never touch the network.
2. Else **downloads** the correct per-RID native from this release's GitHub assets, **verifies a pinned SHA-256**, caches it under `~/.cache/kcap/native/e_sqlite3/<engine-version>/<rid>/`, and loads it.
3. Supports **`KCAP_SQLITE_NATIVE_BASE_URL`** (HTTP mirror **or** local dir / `file://`) for airgapped/offline installs.
4. On failure prints an **actionable message** with the exact cache path, instead of a raw type-initializer crash.

`release.yml` publishes the **pristine** per-RID native (from the restored NuGet package — the macOS publish output is ad-hoc re-signed and hash-unstable) as release assets named to match the resolver's pin table (`libe_sqlite3-<rid>.<ext>` / `e_sqlite3-win-x64.dll`).

## Hardening
- `ClassifyAsync` early-returns on an empty slice → a single vendor's unreadable db skips OpenCode instead of crashing the whole `--all` run.
- `DiscoverAsync` surfaces the first `DllNotFoundException` message (download/mirror guidance) — `GetBaseException()` over-unwraps to the inner DNS error.

## Notes / trade-offs
- This path requires network at first OpenCode import unless `KCAP_SQLITE_NATIVE_BASE_URL` is set or the cache is pre-seeded — inherent to download-on-demand; mitigated by the env override + actionable error.
- Pins are for `SQLitePCLRaw.lib.e_sqlite3` **3.50.3** (what `bundle_e_sqlite3` 3.0.3 restores). `SqliteNativeResolverTests` has a drift guard that fails if the restored native no longer matches the pin, forcing a regen on bundle bumps.

## Testing
- 5 new resolver unit tests: pin/drift guard, download→verify→cache→load (asserts the cached lib is a real loadable SQLite), integrity rejection (corrupt download leaves nothing cached), unsupported RID, resolver↔release.yml naming contract.
- 35 existing OpenCode tests pass.
- Manual: AOT-published binary with **no** bundled lib fetched from a local mirror and read 6 opencode sessions; offline path printed the actionable message.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -156,8 +156,10 @@ jobs:
           # of truth: src/Capacitor.Cli/Commands/SqliteNativeResolver.cs). A mismatch (e.g.
           # publish post-processed the native) would ship an asset that every cold-cache
           # OpenCode import fails to verify — fail the release here instead.
+          # `|| true`: a zero-match grep returns non-zero, which under `set -euo pipefail`
+          # would abort the substitution before the actionable empty-check below can run.
           pin=$(grep -E "\[\"$rid\"\]" src/Capacitor.Cli/Commands/SqliteNativeResolver.cs \
-                | grep -oE '[0-9a-f]{64}' | head -1)
+                | grep -oE '[0-9a-f]{64}' | head -1 || true)
           if [ -z "$pin" ]; then
             echo "::error::no SqliteNativeResolver pin found for $rid"
             exit 1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,6 +125,29 @@ jobs:
             exit 1
           fi
 
+      # `kcap import --opencode` reads opencode.db via Microsoft.Data.Sqlite +
+      # SQLitePCLRaw.bundle_e_sqlite3. NativeAOT does NOT statically link the engine and we
+      # deliberately do NOT bundle it (it's only needed for OpenCode import) — instead the
+      # CLI downloads it on demand from this release's assets (SqliteNativeResolver).
+      # Publish the PRISTINE per-RID native (from the restored NuGet package, not the
+      # re-signed publish output) as a release asset, named to match the resolver's pin
+      # table: <basename>-<rid>.<ext> (e.g. libe_sqlite3-osx-arm64.dylib, e_sqlite3-win-x64.dll).
+      - name: Stage e_sqlite3 native release asset
+        shell: bash
+        run: |
+          set -euo pipefail
+          src=$(find "$HOME/.nuget/packages/sqlitepclraw.lib.e_sqlite3" \
+                  -path "*/runtimes/${{ matrix.rid }}/native/*e_sqlite3*" -type f | head -1)
+          if [ -z "$src" ]; then
+            echo "::error::pristine e_sqlite3 native not found in NuGet cache for ${{ matrix.rid }}"
+            find "$HOME/.nuget/packages/sqlitepclraw.lib.e_sqlite3" -maxdepth 3 -type d || true
+            exit 1
+          fi
+          mkdir -p sqlite-native
+          base=$(basename "$src"); name="${base%.*}"; ext="${base##*.}"
+          cp "$src" "sqlite-native/${name}-${{ matrix.rid }}.${ext}"
+          ls -la sqlite-native/
+
       - name: Copy binaries to npm package
         shell: bash
         run: |
@@ -162,6 +185,14 @@ jobs:
         with:
           name: release-${{ matrix.rid }}
           path: kcap-${{ matrix.rid }}.*
+
+      # Per-RID SQLite native, downloaded on demand by SqliteNativeResolver. Named
+      # release-* so the github-release job's `pattern: release-*` publishes it.
+      - name: Upload e_sqlite3 native asset
+        uses: actions/upload-artifact@v7
+        with:
+          name: release-sqlite-native-${{ matrix.rid }}
+          path: sqlite-native/
 
   publish-npm:
     needs: build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -129,18 +129,19 @@ jobs:
       # SQLitePCLRaw.bundle_e_sqlite3. NativeAOT does NOT statically link the engine and we
       # deliberately do NOT bundle it (it's only needed for OpenCode import) — instead the
       # CLI downloads it on demand from this release's assets (SqliteNativeResolver).
-      # Publish the PRISTINE per-RID native (from the restored NuGet package, not the
-      # re-signed publish output) as a release asset, named to match the resolver's pin
-      # table: <basename>-<rid>.<ext> (e.g. libe_sqlite3-osx-arm64.dylib, e_sqlite3-win-x64.dll).
+      # Source the per-RID native straight from the publish output: it's exactly the bytes
+      # the binary was built against (== the pinned SHA-256 in SqliteNativeResolver) and is
+      # always present, unlike a NuGet cache path that may be absent/polluted on a clean
+      # runner. Name it to match the resolver's pin table: <basename>-<rid>.<ext>
+      # (e.g. libe_sqlite3-osx-arm64.dylib, e_sqlite3-win-x64.dll).
       - name: Stage e_sqlite3 native release asset
         shell: bash
         run: |
           set -euo pipefail
-          src=$(find "$HOME/.nuget/packages/sqlitepclraw.lib.e_sqlite3" \
-                  -path "*/runtimes/${{ matrix.rid }}/native/*e_sqlite3*" -type f | head -1)
+          src=$(ls publish/cli/*e_sqlite3* 2>/dev/null | head -1)
           if [ -z "$src" ]; then
-            echo "::error::pristine e_sqlite3 native not found in NuGet cache for ${{ matrix.rid }}"
-            find "$HOME/.nuget/packages/sqlitepclraw.lib.e_sqlite3" -maxdepth 3 -type d || true
+            echo "::error::e_sqlite3 native not found in publish/cli for ${{ matrix.rid }}"
+            ls -la publish/cli/ || true
             exit 1
           fi
           mkdir -p sqlite-native
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -138,15 +138,43 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
-          src=$(ls publish/cli/*e_sqlite3* 2>/dev/null | head -1)
-          if [ -z "$src" ]; then
-            echo "::error::e_sqlite3 native not found in publish/cli for ${{ matrix.rid }}"
+          rid="${{ matrix.rid }}"
+
+          # Require exactly one match. nullglob so zero matches is an empty array (not the
+          # literal pattern), and avoids set -e aborting mid-assignment before we can report.
+          shopt -s nullglob
+          matches=(publish/cli/*e_sqlite3*)
+          shopt -u nullglob
+          if [ "${#matches[@]}" -ne 1 ]; then
+            echo "::error::expected exactly one e_sqlite3 native in publish/cli for $rid, found ${#matches[@]}"
             ls -la publish/cli/ || true
             exit 1
           fi
+          src="${matches[0]}"
+
+          # Verify the bytes match the SHA-256 the resolver pins for this RID (single source
+          # of truth: src/Capacitor.Cli/Commands/SqliteNativeResolver.cs). A mismatch (e.g.
+          # publish post-processed the native) would ship an asset that every cold-cache
+          # OpenCode import fails to verify — fail the release here instead.
+          pin=$(grep -E "\[\"$rid\"\]" src/Capacitor.Cli/Commands/SqliteNativeResolver.cs \
+                | grep -oE '[0-9a-f]{64}' | head -1)
+          if [ -z "$pin" ]; then
+            echo "::error::no SqliteNativeResolver pin found for $rid"
+            exit 1
+          fi
+          if command -v sha256sum >/dev/null 2>&1; then
+            actual=$(sha256sum "$src" | cut -d' ' -f1)
+          else
+            actual=$(shasum -a 256 "$src" | cut -d' ' -f1)
+          fi
+          if [ "$actual" != "$pin" ]; then
+            echo "::error::e_sqlite3 native for $rid ($actual) does not match resolver pin ($pin)"
+            exit 1
+          fi
+
           mkdir -p sqlite-native
           base=$(basename "$src"); name="${base%.*}"; ext="${base##*.}"
-          cp "$src" "sqlite-native/${name}-${{ matrix.rid }}.${ext}"
+          cp "$src" "sqlite-native/${name}-${rid}.${ext}"
           ls -la sqlite-native/
 
       - name: Copy binaries to npm package
--- a/src/Capacitor.Cli/Commands/OpenCodeImportSource.cs
+++ b/src/Capacitor.Cli/Commands/OpenCodeImportSource.cs
@@ -77,7 +77,14 @@ internal sealed class OpenCodeImportSource : IImportSource {
         } catch (OperationCanceledException) {
             throw;
         } catch (Exception ex) {
-            Console.Error.WriteLine($"[kcap] OpenCode discovery skipped (unreadable {_dbPath}): {ex.Message}");
+            // Surface the actionable cause, not "a type initializer threw an exception".
+            // Prefer the on-demand native-lib failure (SqliteNativeResolver throws a
+            // DllNotFoundException with download/mirror guidance) anywhere in the chain;
+            // otherwise the base exception (e.g. a genuinely corrupt db's SQLite error).
+            Exception cause = ex.GetBaseException();
+            for (var e = ex; e is not null; e = e.InnerException)
+                if (e is DllNotFoundException) { cause = e; break; }
+            Console.Error.WriteLine($"[kcap] OpenCode discovery skipped ({_dbPath}): {cause.Message}");
             return Task.FromResult<IReadOnlyList<DiscoveredSession>>([]);
         }
         return Task.FromResult<IReadOnlyList<DiscoveredSession>>(result);
@@ -85,6 +92,11 @@ internal sealed class OpenCodeImportSource : IImportSource {
 
     public async Task<IReadOnlyList<ImportCommand.SessionClassification>> ClassifyAsync(
         IReadOnlyList<DiscoveredSession> sessions, ClassifyContext ctx, CancellationToken ct) {
+        // DiscoverAsync defensively swallows db-open failures (returns []), but Classify
+        // is still invoked for every source. Don't open the db for an empty slice — a
+        // single vendor's unreadable db must skip OpenCode, not crash the whole import run.
+        if (sessions.Count == 0) return [];
+
         using var db = new OpenCodeDb(_dbPath);
         var results = new List<ImportCommand.SessionClassification>(sessions.Count);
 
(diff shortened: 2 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -125,6 +125,29 @@ jobs:
...
+      # deliberately do NOT bundle it (it's only needed for OpenCode import) — instead the
+      # CLI downloads it on demand from this release's assets (SqliteNativeResolver).
+      # Publish the PRISTINE per-RID native (from the restored NuGet package, not the
+      # re-signed publish output) as a release asset, named to match the resolver's pin
+      # table: <basename>-<rid>.<ext> (e.g. libe_sqlite3-osx-arm64.dylib, e_sqlite3-win-x64.dll).
+      - name: Stage e_sqlite3 native release asset
+        shell: bash
+        run: |
+          set -euo pipefail
+          src=$(find "$HOME/.nuget/packages/sqlitepclraw.lib.e_sqlite3" \
```
realtonyyoung: **blocker**: This release step is looking in `~/.nuget/packages/sqlitepclraw.lib.e_sqlite3`, but the PR pins `SQLitePCLRaw.bundle_e_sqlite3` 3.0.3, whose nuspec restores the native payload through `SourceGear.sqlite3` 3.50.4.5 instead. On a clean release runner this directory will be absent, so the build exits here before uploading any SQLite native assets; on a runner with an old/polluted NuGet cache it can pick unrelated `SQLitePCLRaw.lib.e_sqlite3` bytes that do not match the actual restored bundle. Fix by making the package source a single source of truth: either add/pin an explicit `SQLitePCLRaw.lib.e_sqlite3` dependency that matches `EngineVersion` and use that exact versioned path, or switch the resolver pins, tests, and workflow to the currently restored `SourceGear.sqlite3/<version>` native assets.

test/Capacitor.Cli.Tests.Unit/SqliteNativeResolverTests.cs
```diff
@@ -0,0 +1,121 @@
...
+        await Assert.That(() =>
+                SqliteNativeResolver.EnsureNativeLibrary("plan9-pdp11", null, cache.Path, "0.0.0"))
+            .Throws<DllNotFoundException>();
+    }
+
+    // --- helpers -----------------------------------------------------------
+
+    static string PristineNativePath(string rid) {
+        var home = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+        return Path.Combine(home, ".nuget", "packages", "sqlitepclraw.lib.e_sqlite3",
```
realtonyyoung: **major**: The new guard tests also hard-code `sqlitepclraw.lib.e_sqlite3`, but that package is not restored by the current `SQLitePCLRaw.bundle_e_sqlite3` 3.0.3 dependency chain; the restored native package is `SourceGear.sqlite3` 3.50.4.5. On a clean CI agent this path is missing, while it passes on a developer machine only if an old/manual NuGet cache entry happens to exist. Fix this helper to locate the same versioned package that the project actually restores, or add an explicit pinned package reference so the test, resolver, and release workflow all validate the same bytes.

src/Capacitor.Cli/Commands/SqliteNativeResolver.cs
```diff
@@ -0,0 +1,203 @@
...
+        var actual = Sha256(bytes);
+        if (!string.Equals(actual, asset.Sha256, StringComparison.OrdinalIgnoreCase)) {
+            throw new DllNotFoundException(
+                $"The SQLite native library fetched from {source} failed its integrity check " +
+                $"(expected sha256 {asset.Sha256}, got {actual}). Refusing to load it.");
+        }
+
+        // Atomic publish into the cache: write a unique temp file then move into place,
+        // tolerating a concurrent process that already won the race.
+        var tmp = $"{target}.{Environment.ProcessId}.tmp";
```
realtonyyoung: **major**: The temp name is only unique per process, so two concurrent cold-cache first uses in the same `kcap` process can both write `target.<pid>.tmp`. One thread can then get an `IOException` from `File.WriteAllBytes` or lose after the other moves/deletes the temp, which bubbles out of the DllImport resolver and makes OpenCode discovery/import fail even though the downloaded bytes were valid. Fix by making the temp path unique per attempt, for example with `Guid.NewGuid()` or `Path.GetRandomFileName()` in the same cache directory, and when swallowing a move race only return the target after re-checking that its SHA-256 matches the expected asset hash.

src/Capacitor.Cli/Commands/SqliteNativeResolver.cs
```diff
@@ -0,0 +1,203 @@
...
+            return (http.GetByteArrayAsync(url).GetAwaiter().GetResult(), url);
+        } catch (Exception ex) {
+            throw NotAvailable(asset, url, ex);
+        }
+    }
+
+    static DllNotFoundException NotAvailable(NativeAsset asset, string source, Exception inner) =>
+        new($"kcap import --opencode needs the SQLite native library and tried to fetch it from " +
+            $"{source}, but that failed: {inner.Message}. If you are offline or behind a proxy, " +
+            $"download {asset.AssetName} into {Path.Combine(DefaultCacheRoot(), EngineVersion, "<rid>")} " +
```
realtonyyoung: **minor**: The offline guidance says to pre-seed `.../<rid>` literally because `NotAvailable` no longer has the actual RID, even though `EnsureNativeLibrary` knows it. In the failure scenario this is the main actionable path an airgapped user sees, and copying `libe_sqlite3-osx-arm64.dylib` into a directory named `<rid>` will not satisfy the cache lookup at `.../osx-arm64/libe_sqlite3.dylib`. Pass `rid` into `Fetch`/`NotAvailable` or include the computed `cacheDir` in the error so the message prints the exact directory to seed.

.github/workflows/release.yml
realtonyyoung: Confirmed and fixed in 4cb2c0a. Verified via project.assets.json: bundle_e_sqlite3 3.0.3 restores the native through **SourceGear.sqlite3 3.50.4.5**, not SQLitePCLRaw.lib.e_sqlite3 (the lib.e_sqlite3/3.50.3 in my cache was unrelated leftover). The release step now sources the per-RID native from the publish output (`publish/cli/*e_sqlite3*`) — always present and byte-identical to the bytes the binary links against — rather than a NuGet cache path that's absent/polluted on a clean runner. EngineVersion + all 6 resolver pins are now SourceGear's; verified the publish-output hash equals the pin (7b319cd3…).

test/Capacitor.Cli.Tests.Unit/SqliteNativeResolverTests.cs
realtonyyoung: Fixed in 4cb2c0a. The drift-guard helper now resolves `sourcegear.sqlite3/<EngineVersion>` (the package actually restored), so it validates the same bytes the resolver pins and the release publishes. Confirmed it exists on a clean restore.

src/Capacitor.Cli/Commands/SqliteNativeResolver.cs
realtonyyoung: Fixed in 4cb2c0a. Temp file is now per-attempt unique (`{FileName}.{Guid:N}.tmp`), and on a lost move-race we only return the existing target after re-verifying its SHA-256 — otherwise rethrow. No more collision on `target.<pid>.tmp`.

src/Capacitor.Cli/Commands/SqliteNativeResolver.cs
realtonyyoung: Fixed in 4cb2c0a. `rid` is now threaded into Fetch/NotAvailable; the message prints the exact seed path `~/.cache/kcap/native/e_sqlite3/<EngineVersion>/<rid>/<filename>` instead of a literal `<rid>`.

.github/workflows/release.yml
```diff
@@ -125,6 +125,30 @@ jobs:
...
+      # Source the per-RID native straight from the publish output: it's exactly the bytes
+      # the binary was built against (== the pinned SHA-256 in SqliteNativeResolver) and is
+      # always present, unlike a NuGet cache path that may be absent/polluted on a clean
+      # runner. Name it to match the resolver's pin table: <basename>-<rid>.<ext>
+      # (e.g. libe_sqlite3-osx-arm64.dylib, e_sqlite3-win-x64.dll).
+      - name: Stage e_sqlite3 native release asset
+        shell: bash
+        run: |
+          set -euo pipefail
+          src=$(ls publish/cli/*e_sqlite3* 2>/dev/null | head -1)
```
realtonyyoung: **major**: This still trusts whichever file the glob happens to return and uploads it without checking the bytes against the resolver pin for the matrix RID. If `dotnet publish` ever post-processes `e_sqlite3` on one runner, or the glob matches more than one file and `head -1` chooses the wrong one, the release will contain an asset with the exact name the resolver downloads but every cold-cache OpenCode import on that RID will fail the SHA-256 check. Also, with `set -euo pipefail`, zero matches exits during this assignment before the custom `::error::` guard below can run. Fix by collecting matches with `nullglob`, requiring exactly one match, and comparing `sha256sum`/`shasum` against the expected pin for `${{ matrix.rid }}` before uploading.

test/Capacitor.Cli.Tests.Unit/SqliteNativeResolverTests.cs
```diff
@@ -0,0 +1,124 @@
...
+/// </summary>
+public class SqliteNativeResolverTests {
+    /// <summary>
+    /// Drift guard: the SHA-256 pinned for THIS platform must match the native that
+    /// SQLitePCLRaw actually restored. If SQLitePCLRaw.bundle_e_sqlite3 is bumped, this
+    /// fails until EngineVersion + the Assets hashes are regenerated.
+    /// </summary>
+    [Test]
+    public async Task pinned_hash_matches_restored_native_for_current_rid() {
+        var rid = SqliteNativeResolver.CurrentRid();
```
realtonyyoung: **major**: This drift guard only checks the current test runner RID, but the resolver ships hard-coded pins for six RIDs and CI only exercises a subset of them. A typo in, say, the macOS or musl-arm64 hash would pass on linux-x64/windows CI, the release job would still upload that asset, and users on that RID would fail integrity verification on first OpenCode import. Since the SourceGear package contains all six runtime assets after restore, make this test iterate every `SqliteNativeResolver.Assets` entry and compare each package file hash to its pin (or add an equivalent release-time hash check for every matrix RID).

src/Capacitor.Cli/Commands/OpenCodeImportSource.cs
```diff
@@ -77,14 +77,26 @@ public Task<IReadOnlyList<DiscoveredSession>> DiscoverAsync(DiscoveryFilters fil
...
         }
         return Task.FromResult<IReadOnlyList<DiscoveredSession>>(result);
     }
 
     public async Task<IReadOnlyList<ImportCommand.SessionClassification>> ClassifyAsync(
         IReadOnlyList<DiscoveredSession> sessions, ClassifyContext ctx, CancellationToken ct) {
+        // DiscoverAsync defensively swallows db-open failures (returns []), but Classify
+        // is still invoked for every source. Don't open the db for an empty slice — a
+        // single vendor's unreadable db must skip OpenCode, not crash the whole import run.
+        if (sessions.Count == 0) return [];
```
realtonyyoung, line 91: **minor**: The empty-slice guard prevents the common “discovery failed, then classify opens the DB again” crash, but this constructor is still outside any catch. If discovery succeeds and then `opencode.db` is deleted, permissions change, the WAL sidecars become unreadable, or the native resolver fails on this second open, `ClassifyAsync` throws out of the source task and aborts the whole multi-vendor probe instead of marking only OpenCode as failed. Wrap the `OpenCodeDb` open in the same kind of catch as discovery and return `ProbeError` classifications for this slice (or `[]` if there is no safe per-session result).

src/Capacitor.Cli/Commands/SqliteNativeResolver.cs
```diff
@@ -0,0 +1,208 @@
...
+
+    static (byte[] bytes, string source) Fetch(NativeAsset asset, string rid, string? baseUrlOverride, string selfVersion) {
+        var baseUrl = string.IsNullOrWhiteSpace(baseUrlOverride)
+            ? $"https://github.com/kurrent-io/kcap-cli/releases/download/v{selfVersion}"
+            : baseUrlOverride.TrimEnd('/', '\\');
+
+        // Local mirror (airgap): a plain directory path or file:// URL.
+        if (!baseUrl.StartsWith("http://", StringComparison.OrdinalIgnoreCase) &&
+            !baseUrl.StartsWith("https://", StringComparison.OrdinalIgnoreCase)) {
+            var dir = baseUrl.StartsWith("file://", StringComparison.OrdinalIgnoreCase)
```
realtonyyoung: **minor**: The `file://` URI parse happens before the `try`, so a malformed `KCAP_SQLITE_NATIVE_BASE_URL` such as an invalid file URI throws `UriFormatException` directly from the resolver. Discovery then surfaces a raw URI error instead of the actionable `NotAvailable` message with the expected asset name/cache seed path. Move the URI parsing inside the `try` or use `Uri.TryCreate` and wrap failures through `NotAvailable` so bad airgap configuration gets the same guidance as a missing local mirror file.

.github/workflows/release.yml
realtonyyoung: Fixed in 050dac8. The step now uses `nullglob` + requires exactly one match (so zero matches reports our `::error::` instead of `set -e` aborting mid-assignment), and verifies the staged asset's SHA-256 against the resolver pin for `${{ matrix.rid }}` — grepped from SqliteNativeResolver.cs as the single source of truth — before upload. A post-processed or wrong-file native now fails the release rather than shipping an asset cold-cache imports reject. Simulated locally: osx-arm64 pin extracted from source == native hash.

test/Capacitor.Cli.Tests.Unit/SqliteNativeResolverTests.cs
realtonyyoung: Fixed in 050dac8. The guard now iterates every `SqliteNativeResolver.Assets` entry and compares each to the restored SourceGear native (the package ships all six runtimes after restore), so a typo in any RID's pin fails on a single agent. Belt-and-suspenders with the new release-time per-RID pin check.

src/Capacitor.Cli/Commands/OpenCodeImportSource.cs
realtonyyoung: Fixed in 050dac8. ClassifyAsync now wraps the OpenCodeDb open in try/catch; on failure it returns ProbeError classifications for the slice (reusing SurfaceCause for the message) instead of throwing out of the multi-vendor probe. Extracted MetaFor so the probe-error path builds the same metadata as the per-session loop.

src/Capacitor.Cli/Commands/SqliteNativeResolver.cs
realtonyyoung: Fixed in 050dac8. The file:// parse is now inside the try, so a malformed KCAP_SQLITE_NATIVE_BASE_URL surfaces the actionable NotAvailable guidance (asset name + seed path) instead of a raw UriFormatException out of the resolver.

.github/workflows/release.yml
```diff
@@ -125,6 +125,58 @@ jobs:
...
+            ls -la publish/cli/ || true
+            exit 1
+          fi
+          src="${matches[0]}"
+
+          # Verify the bytes match the SHA-256 the resolver pins for this RID (single source
+          # of truth: src/Capacitor.Cli/Commands/SqliteNativeResolver.cs). A mismatch (e.g.
+          # publish post-processed the native) would ship an asset that every cold-cache
+          # OpenCode import fails to verify — fail the release here instead.
+          pin=$(grep -E "\[\"$rid\"\]" src/Capacitor.Cli/Commands/SqliteNativeResolver.cs \
```
realtonyyoung, line 161: **minor**: The “no pin found” guard below is still mostly unreachable under `set -euo pipefail`: if this grep chain finds zero lines (for example a future matrix RID is added without an `Assets` entry, or the resolver table is reformatted so the hash is not on the same line), the command substitution returns non-zero and bash exits before `[ -z "$pin" ]` can print the intended `::error::`. The release still fails, but without the actionable message this block is trying to provide. Wrap the pipeline in a non-errexit form, e.g. `pin=$(... || true)` followed by the existing empty check, or use `awk`/`sed` in a way that returns 0 for no match and then validate exactly one pin.

src/Capacitor.Cli/Commands/SqliteNativeResolver.cs
```diff
@@ -0,0 +1,211 @@
...
+            if (!(File.Exists(target) && FileSha256(target) == asset.Sha256))
+                throw;
+        }
+        return target;
+    }
+
+    static (byte[] bytes, string source) Fetch(NativeAsset asset, string rid, string? baseUrlOverride, string selfVersion) {
+        var baseUrl = string.IsNullOrWhiteSpace(baseUrlOverride)
+            ? $"https://github.com/kurrent-io/kcap-cli/releases/download/v{selfVersion}"
+            : baseUrlOverride.TrimEnd('/', '\\');
```
realtonyyoung: **minor**: Trimming both `/` and `\` before deciding whether this is a local path breaks root local mirrors. For example `KCAP_SQLITE_NATIVE_BASE_URL=/` becomes an empty string and the resolver looks for `libe_sqlite3-<rid>.*` relative to the current working directory; on Windows, a root like `C:\` becomes `C:` (drive-relative) rather than `C:\`. That makes an otherwise valid airgap mirror at the filesystem/drive root fail with a misleading missing-file message. Avoid trimming local paths, or use path-aware trimming that preserves roots (and only strip separators when constructing HTTP URLs).

.github/workflows/release.yml
realtonyyoung: Fixed in dddcc32. The pin-extraction substitution now ends in `|| true`, so a zero-match falls through to the `[ -z "$pin" ]` guard and prints the actionable `::error::no SqliteNativeResolver pin found` instead of set -e aborting the substitution.

src/Capacitor.Cli/Commands/SqliteNativeResolver.cs
realtonyyoung: Fixed in dddcc32. No longer trimming separators off the base value — local mirror paths are used verbatim (Path.Combine tolerates a trailing one), preserving `/` and `C:\`. Only the HTTP branch trims a trailing slash when appending the asset name.

PR #200, opened 2026-06-29
[AI-9] Phase 3: daemon consumes eval catalog + posts V3 results

realtonyyoung:
## AI-9 Phase 3 (kcap-cli) — daemon consumes the eval catalog + posts V3 results

The eval daemon no longer ships hardcoded judge prompts. It fetches the server's runtime catalog, reconciles the run against it, and posts versioned (V3) results — so prompt edits land via the admin catalog with no CLI release after this one-time bump.

### What's here (7 commits)
- **DTOs + V3 payload** (`Models.cs`): `EvalCatalogDto`/`EvalCatalogQuestionDto`, additive nullable `prompt_version` on `EvalQuestionDto`/`EvalQuestionVerdict`, `SessionEvalCompletedPayloadV3` — all source-gen registered (AOT-clean).
- **`EvalCatalogClient`**: fetches `GET /api/eval/catalog` with fail-fast validation (null → run aborts).
- **`EvalService` reconciliation + V3 posting**: `PrepareAsync` reconciles the run question list from the catalog (rendered prompt + raw text + version + `needs_tools`, in selected order); `Aggregate` stamps each verdict's `prompt_version`; `FinalizeAsync` posts V3 to `/evals/v3`, filling the retrospective `{TRACE_JSON}` from the daemon's trace; `IEvalObserver`/`EvalCommand`/daemon `EvalRunner` flipped to V3.
- **WireMock integration tests**: V3 post wire-shape, catalog fetch/reconcile e2e, tools routing (the four `needs_tools` ids), and an alias raw-text double-wrap guard.
- **Cleanup**: dropped the now-dead embedded text-question + retrospective prompts (tools wrapper kept).

### Notes
- AOT: source-gen JSON only; `dotnet publish` clean (0 IL2026/IL3050).
- Tools path keeps the embedded `prompt-eval-question-tools.txt` (catalog has no tools template — follow-up tracked).
- Built/tested against WireMock (Phase-2-independent); the live end-to-end is covered in the paired kcap-server PR.
- Reviewed via subagent-driven development (per-task spec+quality reviews + a final whole-branch review: ready to merge).

**Merge order:** this kcap-cli PR merges FIRST; the paired kcap-server PR then re-points its `src/cli` submodule at the merged commit before it merges.


src/Capacitor.Cli.Core/Eval/EvalCatalogClient.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            var json   = await resp.Content.ReadAsStringAsync(ct);
+            var parsed = JsonSerializer.Deserialize(json, CapacitorJsonContext.Default.EvalCatalogDto);
+            if (parsed is null) {
+                observer.OnFailed("eval catalog response was not valid JSON");
+                return null;
+            }
+
+            // SF#4 -- fail-fast integrity checks before any judge invocation.
+            if (parsed.Questions.Count == 0) {
```
realtonyyoung: FetchAsync is required to return null for malformed catalog responses, but this validation assumes `parsed.Questions` is non-null. A catalog like `{"retrospective_prompt":"r","retrospective_prompt_version":"1","questions":null}` will deserialize successfully and then throw `NullReferenceException` here instead of returning null; `questions:[null]` similarly throws in the foreach below. That bubbles out of the catalog client on the daemon prepare path, so it no longer fails closed with a catalog-specific `OnFailed`. Please treat a null `Questions` list or null question element as malformed and return null.

realtonyyoung: Fixed in 89f4c36. `FetchAsync` now fails closed (returns null + OnFailed) on both a JSON `"questions":null` (which overrides the [] initializer — guarded via `is not { Count: > 0 }`) and a `[null]` element (a per-item `q is null` guard before field access), instead of NRE-ing past the HttpRequestException/JsonException handlers. Added `Returns_null_on_null_questions_array` + `Returns_null_on_null_question_entry`.

src/Capacitor.Cli.Core/Eval/EvalService.cs
```diff
@@ -277,12 +278,20 @@ bool                                         ForceTools
                 return null;
             }
 
-            var ctx = await PrepareAsync(baseUrl, httpClient, sessionId, questions, chain, thresholdBytes, observer, ct, model, evalRunId);
+            // AI-9 Phase 3 — fetch the full catalog (rendered prompts + raw text +
+            // versions) so PrepareAsync can reconcile the run question list from it.
+            var catalog = await EvalCatalogClient.FetchAsync(baseUrl, httpClient, observer, ct);
+            if (catalog is null) return null;   // FetchAsync already emitted OnFailed
+
+            var ctx = await PrepareAsync(baseUrl, httpClient, sessionId, questions, catalog, chain, thresholdBytes, observer, ct, model, evalRunId);
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in 59c42da. Added a **Server requirement** note to the `### Session evaluation (LLM-as-judge)` section of README.md: `kcap eval` now fetches its catalog from `GET /api/eval/catalog` and posts results to `POST /api/sessions/{id}/evals/v3`, fails fast against a server that doesn't expose the catalog endpoint, and the docs tell the user to upgrade the server in that case.

PR #203, opened 2026-06-30
[AI-1061] kcap mcp flows: infinite POST timeout + opt-in async poll loop

realtonyyoung:
## Problem
The `kcap mcp flows` MCP server called the long-blocking `/api/flows/review/start` with the default `HttpClient.Timeout` (**100s**), so any real review aborted at 100s and the server tore the reviewer down (~2 min). Even with that fixed, the server's synchronous long-poll hits **Cloudflare's ~100s 524** behind `kurrent.kcap.ai`. (See [AI-1061](https://linear.app/kurrent/issue/AI-1061).)

## Fix (CLI side)
1. **Infinite client timeout** on the flows client (so a blocking / self-hosted / non-Cloudflare server can complete) — mirrors `PermissionRequestCommand` / `CodexHookCommand`.
2. **Opt-in async + poll** (pairs with the server async rework in kcap-server #881): `start`/`submit` send `async:true`; if the POST returns a terminal result (old/blocking server) it's used directly (no poll); otherwise the CLI **polls `GET /api/flows/{id}`** keyed on `round_number` until the round is terminal, then returns the same findings text. Each poll GET uses its own **short timeout** (while the POST stays infinite), the **401-refresh** wrapper, bounded transient retry, a ~10s **404 grace**, an ~8-min **graceful cap** ("still running…"), and **stops on a terminal run status** (can't hang on a failed run).

MCP tool names/arguments and the `FINDINGS:`/`NO FINDINGS` UX are unchanged. The `async` flag makes this backward-compatible: old CLI ↔ new server keeps blocking behavior, so **deploy order is unconstrained**.

## Verification
`McpFlowsServerTests` 12/0 (incl. async-polls-until-findings, terminal-result-from-POST-no-poll, transient-500-retry, run-terminal-stop). AOT publish: **no IL2xxx/IL3xxx warnings**.

### DIFF
--- a/src/Capacitor.Cli/Commands/McpFlowsServer.cs
+++ b/src/Capacitor.Cli/Commands/McpFlowsServer.cs
@@ -15,6 +15,15 @@ static class McpFlowsServer {
     public static async Task<int> RunAsync(string baseUrl) {
         using var client = await HttpClientExtensions.CreateAuthenticatedClientAsync(baseUrl);
 
+        // The review-flow endpoints are long-polls: start_review_flow / submit_review_round
+        // block server-side until the AI reviewer returns findings (FlowResultWaiter allows
+        // up to 10 minutes). The default HttpClient.Timeout (100s) would abort the POST first,
+        // which the server sees as a cancelled request and tears the reviewer down (~2 min) —
+        // so no real review could ever complete (AI-1061). Disable the client-side deadline and
+        // let the server's FlowResultWaiter and the harness MCP tool timeout bound the wait,
+        // mirroring the long-poll clients in PermissionRequestCommand / CodexHookCommand.
+        client.Timeout = System.Threading.Timeout.InfiniteTimeSpan;
+
         var cwd      = Directory.GetCurrentDirectory();
         var repoRoot = GitRepository.FindRoot(cwd);
         var tools    = BuildToolsList();
--- a/src/Capacitor.Cli/Commands/McpFlowsServer.cs
+++ b/src/Capacitor.Cli/Commands/McpFlowsServer.cs
@@ -93,9 +93,25 @@ static class McpFlowsServer {
         try {
             var apiRoot = baseUrl.TrimEnd('/');
 
+            // start_review_flow and submit_review_round may need async poll — handle separately.
+            if (toolName is "start_review_flow" or "submit_review_round") {
+                using var postResponse = toolName == "start_review_flow"
+                    ? await SendWithRefreshRetryAsync(client, c => StartReviewFlowAsync(c, apiRoot, arguments, cwd, repoRoot, repoInfo))
+                    : await SendWithRefreshRetryAsync(client, c => SubmitReviewRoundAsync(c, apiRoot, arguments));
+
+                var postBody = await postResponse.Content.ReadAsStringAsync();
+
+                if (postResponse.StatusCode == HttpStatusCode.Unauthorized)
+                    return BuildToolResult(id, "Not logged in. Run 'kcap login' on the host shell.", isError: true);
+
+                if (!postResponse.IsSuccessStatusCode)
+                    return BuildToolResult(id, $"Error: HTTP {(int)postResponse.StatusCode} — {postBody}", isError: true);
+
+                var payload = await ResolveRoundResultAsync(client, apiRoot, postBody);
+                return BuildToolResult(id, payload);
+            }
+
             using var httpResponse = toolName switch {
-                "start_review_flow"      => await SendWithRefreshRetryAsync(client, c => StartReviewFlowAsync(c, apiRoot, arguments, cwd, repoRoot, repoInfo)),
-                "submit_review_round"    => await SendWithRefreshRetryAsync(client, c => SubmitReviewRoundAsync(c, apiRoot, arguments)),
                 "get_review_flow_status" => await SendWithRefreshRetryAsync(client, c => c.GetAsync(BuildFlowUrl(apiRoot, arguments))),
                 "close_review_flow"      => await SendWithRefreshRetryAsync(client, c => c.PostAsync(BuildFlowUrl(apiRoot, arguments) + "/close", null)),
                 _                        => throw new ArgumentException($"Unknown tool: {toolName}")
@@ -111,13 +127,13 @@ static class McpFlowsServer {
                 return BuildToolResult(id, $"Error: HTTP {(int)httpResponse.StatusCode} — {body}", isError: true);
             }
 
-            var payload = toolName switch {
+            var statusPayload = toolName switch {
                 "get_review_flow_status" => FormatStatusResponse(body),
                 "close_review_flow"      => FormatCloseResponse(body),
                 _                        => FormatRoundResponse(body)
             };
 
-            return BuildToolResult(id, payload);
+            return BuildToolResult(id, statusPayload);
         } catch (ArgumentException ex) {
             return BuildToolResult(id, $"Error: {ex.Message}", isError: true);
         } catch (HttpRequestException ex) {
@@ -182,7 +198,8 @@ static class McpFlowsServer {
             RepoName:             repoInfo?.RepoName,
             DaemonName:           null,
             RepoPath:             repoRoot,
-            Mode:                 mode
+            Mode:                 mode,
+            Async:                true
         );
 
         return await client.PostAsync(
@@ -207,7 +224,7 @@ static class McpFlowsServer {
             );
         }
 
-        var body = new SubmitReviewRoundDto(Context: context, Instructions: instructions);
+        var body = new SubmitReviewRoundDto(Context: context, Instructions: instructions, Async: true);
 
         return client.PostAsync(
             $"{apiRoot}/api/flows/{Uri.EscapeDataString(flowRunId)}/rounds",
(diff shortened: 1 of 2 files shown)


src/Capacitor.Cli/Commands/McpFlowsServer.cs
```diff
@@ -213,6 +239,95 @@ static string BuildFlowUrl(string apiRoot, JsonObject? arguments) {
...
+                var rs       = node?["round_status"]?.GetValue<string>();
+                var runStatus = node?["status"]?.GetValue<string>();
+
+                // If the run-level status is terminal, stop regardless of round status
+                // (guards against a run that failed with an in-flight round left at "running").
+                if (runStatus is "closed" or "failed") {
+                    if (rn == roundNumber && rs is not null)
+                        return FormatPolledRoundResult(node!, flowRunId);
+                    // Run terminal but round not matching yet — return what we have
+                    return FormatPolledRoundResult(node!, flowRunId);
```
realtonyyoung: [Major] This run-terminal branch can return a stale prior round. The normal terminal-round path correctly waits for `rn == roundNumber`, but when the run itself becomes `failed` or `closed` this code formats whatever the status projection currently exposes. For `submit_review_round` round 2, a failed run can still have the GET projection showing round 1 findings/clean text, so the MCP response would hand the caller old `FINDINGS:` or `NO FINDINGS` for the wrong round. The run-level terminal condition should stop the loop, but if `rn != roundNumber` it needs to return an explicit run-failed/closed result for the requested round, not `round_result_text` from the stale projection.

src/Capacitor.Cli/Commands/McpFlowsServer.cs
```diff
@@ -84,9 +93,25 @@ static async Task<string> HandleToolCallAsync(
...
+
+                var postBody = await postResponse.Content.ReadAsStringAsync();
+
+                if (postResponse.StatusCode == HttpStatusCode.Unauthorized)
+                    return BuildToolResult(id, "Not logged in. Run 'kcap login' on the host shell.", isError: true);
+
+                if (!postResponse.IsSuccessStatusCode)
+                    return BuildToolResult(id, $"Error: HTTP {(int)postResponse.StatusCode} — {postBody}", isError: true);
+
+                var payload = await ResolveRoundResultAsync(client, apiRoot, postBody);
```
realtonyyoung: [Major] Poll-path errors lose the MCP error bit. `PollUntilTerminalAsync` can return strings like `Error: flow_run_id ... not found.` or `Not logged in...`, but this caller always wraps the payload with `BuildToolResult(id, payload)` and leaves `isError` false. That is inconsistent with the existing POST/GET error path above and weakens the 404 requirement from “surfaced as an error” to a successful tool response containing error-looking text. Please return a structured poll result, or throw/catch a typed exception, so 404-after-grace and 401-after-refresh become `BuildToolResult(..., isError: true)`.

src/Capacitor.Cli/Commands/McpFlowsServer.cs
```diff
@@ -213,6 +239,95 @@ static string BuildFlowUrl(string apiRoot, JsonObject? arguments) {
...
+            HttpResponseMessage resp;
+            try {
+                resp = await SendWithRefreshRetryAsync(client, c => c.GetAsync(url, getCts.Token));
+            } catch (Exception ex) when (ex is HttpRequestException or OperationCanceledException) {
+                await Task.Delay(PollInterval); continue; // transient — retry within cap
+            }
+
+            using (resp) {
+                if (resp.StatusCode == HttpStatusCode.NotFound) {
+                    firstSeenNotFound ??= DateTimeOffset.UtcNow;
```
realtonyyoung: [Major] The 404 grace is measured from the first 404, not from the POST/poll start. The requirement is to tolerate not-yet-projected 404s only for the short window immediately after the async POST. With this implementation, if every earlier poll is 200/500/network and the first 404 arrives minutes later, `firstSeenNotFound` starts then and the code grants a fresh grace period. Near the 8-minute cap that can even turn a persistent not-found into the misleading “Review still running” cap message. Capture a `notFoundGraceDeadline = pollStartedAt + NotFoundGrace` and treat any 404 after that deadline as an error.

src/Capacitor.Cli/Commands/McpFlowsServer.cs
```diff
@@ -213,6 +239,95 @@ static string BuildFlowUrl(string apiRoot, JsonObject? arguments) {
...
+                    if (DateTimeOffset.UtcNow - firstSeenNotFound > NotFoundGrace)
+                        return $"Error: flow_run_id {flowRunId} not found.";
+                    await Task.Delay(PollInterval); continue;
+                }
+                if (resp.StatusCode == HttpStatusCode.Unauthorized)
+                    return "Not logged in. Run 'kcap login' on the host shell.";
+
+                var body = await resp.Content.ReadAsStringAsync();
+                if (!resp.IsSuccessStatusCode) { await Task.Delay(PollInterval); continue; }
```
realtonyyoung: [Major] This is not a bounded transient retry; it swallows every non-success poll response until the full 8-minute cap and then reports “Review still running”. A persistent 500, 502, TLS/proxy failure, or even a non-transient 403/400 should not consume the whole cap or be reclassified as an in-progress review. The poll loop needs a small retry budget for transient 5xx/network failures and should surface the last error once that budget is exhausted; non-transient 4xx should fail immediately, except for the explicit short 404 projection grace.

realtonyyoung (commented):
Major findings posted inline:

- Run-level terminal branch can return stale prior-round text when the requested round is not the projected round.
- Poll-path errors are returned with `isError` unset, so 404/401 failures become successful MCP tool results.
- 404 grace starts at the first 404 rather than immediately after the POST.
- Non-success/transient poll GETs are retried until the 8-minute cap and then misreported as still running.

I verified the POST keeps the infinite client timeout while poll GETs use per-call cancellation tokens, each poll GET goes through the refresh wrapper, `async:true` is sent via source-generated DTOs, and old-server terminal POSTs skip polling.

Overall verdict: changes required. The async flow is close, but the poll engine still violates the round ownership and error-surfacing requirements in ways that can produce misleading review results.

src/Capacitor.Cli/Commands/McpFlowsServer.cs
realtonyyoung: Fixed in 77005c7. The run-terminal branch now only returns the polled round when its `round_number == roundNumber` AND its `round_status` is terminal; otherwise it returns an explicit `review run {status} before round {N} produced a result` with `isError: true` — never another round's text.

src/Capacitor.Cli/Commands/McpFlowsServer.cs
realtonyyoung: Fixed in 77005c7. `PollUntilTerminalAsync`/`ResolveRoundResultAsync` now return a `PollResult(string Payload, bool IsError)`; the caller destructures it and passes `isError` into `BuildToolResult`, so 404-after-grace, 401-after-refresh-failure, exhausted-retry, run-failed/closed-without-result, and non-transient 4xx all surface as `isError: true`.

src/Capacitor.Cli/Commands/McpFlowsServer.cs
realtonyyoung: Fixed in 77005c7. Replaced `firstSeenNotFound` with `notFoundGraceDeadline = pollStartedAt + NotFoundGrace`, computed once at poll start. Any 404 after that deadline returns an error immediately; a late first-404 no longer grants a fresh grace window or falls through to the cap message.

src/Capacitor.Cli/Commands/McpFlowsServer.cs
realtonyyoung: Fixed in 77005c7. Added a bounded transient-retry budget (`consecutiveTransient`, max 5, reset on a 200) for 5xx/network; exceeding it surfaces the last error (`isError: true`). Non-transient 4xx fail immediately (except 404 within the projection grace and 401-with-refresh). The cap "still running" message is now reached only when the round is genuinely still `running`.

realtonyyoung (commented):
NO FINDINGS on re-review after 77005c7.

Verified the prior four Major findings are addressed:

- Stale prior-round: run-terminal (`failed`/`closed`) GETs only format the polled round when `round_number` matches and `round_status` is terminal; otherwise they return an explicit `Error: review run {status} before round {N} produced a result.` with `isError:true`, so stale prior-round `FINDINGS:` text is not returned.
- Error bit: `ResolveRoundResultAsync` / `PollUntilTerminalAsync` now return a structured `PollResult`, and the caller passes `isError` into `BuildToolResult`. 404-after-grace, 401-after-refresh-failure, exhausted retry, run-terminal-without-result, and non-transient 4xx all surface as MCP errors; terminal poll success stays non-error.
- 404 grace: the deadline is anchored once at `pollStartedAt + NotFoundGrace`; late/persistent 404s return `isError:true` and do not fall through to the 8-minute “still running” cap message.
- Bounded retry: 5xx/network/timeout failures are counted with a consecutive retry budget that resets on 200; exhaustion returns the last error with `isError:true`; non-transient 4xx fail immediately except the 404 grace and the 401 refresh path.

Regression checks also look good: the POST client keeps `Timeout.InfiniteTimeSpan`, each poll GET uses its own short CTS through `GetAsync(url, getCts.Token)`, every poll GET goes through `SendWithRefreshRetryAsync`, terminal-in-POST old-server behavior still skips polling, `async:true` is still source-generated DTO output, and I did not see new reflection/dynamic/JsonArray AOT risk.

Did not run tests or builds, per review instructions. Verdict: ready from this review pass.

PR #208, opened 2026-06-30
fix(auth): harden TokenStore against corrupt token files (AI-1082)

realtonyyoung:
## What & why

Fixes [AI-1082](https://linear.app/kurrent/issue/AI-1082/kcap-crashes-on-corrupt-token-file-harden-tokenstore-load-save). A customer hit an **unhandled `System.Text.Json.JsonException`** running `kcap status` — their local token file was corrupt (a complete JSON object followed by a stray comma + trailing bytes at byte ~492). `TokenStore.LoadAsync` deserialized it with no error handling, so the exception propagated out and crashed the command. The same load path backs **hooks, the watcher, the daemon, the MCP servers, and all authenticated HTTP**, so one corrupt token byte wedged the entire CLI.

## Changes

**1. Tolerate a corrupt token file on read (the customer-facing fix).**
`LoadAsync` now routes both overloads through a private `ReadTokensAsync` that catches `JsonException` and returns `null` — a corrupt / empty / partially-written / hand-edited token file degrades to "not authenticated → run `kcap login`" instead of crashing every command, hook, the daemon, and MCP. The catch is scoped to `JsonException` only, so real `IOException`/permission faults still propagate rather than being masked as auth-absence.

**2. Unique temp filename per write (prevents recurrence).**
`SaveAsync` previously wrote to a fixed, shared `{path}.tmp`. The token store is shared across hooks/watcher/daemon/MCP/login, so two concurrent saves wrote the same temp file and could splice a shorter token over a longer one (the byte-492 corruption signature). Each write now uses a `{pid}.{guid}` temp, cleaned up best-effort on failure; the existing atomic `File.Move` then publishes one complete document, last-writer-wins. `SetUnixFileMode` and the legacy-file migration are unchanged.

## Testing

- 6 new tests in `TokenStoreProfileTests` (4 read-resilience incl. the `GetValidTokensAsync` regression through the public entry point, + concurrency-never-corrupts and no-temp-residue). The concurrency test fails deterministically (RED) against the old shared temp and passes after the fix.
- Full unit suite: **1860/1860 passing** on .NET 10.
- AOT publish: **clean** — no IL2026/IL3050 warnings (`Environment.ProcessId`/`Guid.NewGuid()` are AOT-safe).
- Reviewed by an independent whole-branch pass: ready to merge, no Critical/Important findings.

## Customer workaround (already available)

`kcap login` recovers an affected user now — the login/discovery flow doesn't read the existing token before overwriting it.

## Docs

No README change — internal robustness only, no user-facing CLI surface change (per the CLAUDE.md README-sync rule). Design + plan added under `docs/superpowers/`.

### DIFF
--- a/docs/superpowers/plans/2026-06-30-token-store-corruption-resilience.md
+++ b/docs/superpowers/plans/2026-06-30-token-store-corruption-resilience.md
@@ -1,0 +1,295 @@
+# TokenStore Corruption Resilience Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Stop `kcap` from crashing on a corrupt local token file, and stop concurrent writers from producing one.
+
+**Architecture:** Two centralized changes in `TokenStore` (`src/Capacitor.Cli.Core/Auth/TokenStore.cs`). (1) A private `ReadTokensAsync(path)` reader wraps the deserialize in `try { … } catch (JsonException) { return null; }`; both `LoadAsync` overloads route through it, so a corrupt/empty/hand-edited file degrades to "not authenticated" across every caller. (2) `SaveAsync` writes to a per-write **unique** temp filename (`{path}.{pid}.{guid}.tmp`) instead of a shared `{path}.tmp`, with best-effort temp cleanup in `finally`; the existing atomic `File.Move` then guarantees the destination is always one complete document. No call sites change.
+
+**Tech Stack:** .NET 10, NativeAOT, `System.Text.Json` source-gen (`CapacitorJsonContext`), TUnit on Microsoft Testing Platform.
+
+**Issue:** [AI-1082](https://linear.app/kurrent/issue/AI-1082/kcap-crashes-on-corrupt-token-file-harden-tokenstore-load-save)
+**Spec:** `docs/superpowers/specs/2026-06-30-token-store-corruption-resilience-design.md`
+
+## Global Constraints
+
+- Build/test/publish with `~/.dotnet/dotnet` — the PATH `dotnet` is 8.0 and cannot target .NET 10.
+- Catch **`JsonException` only** on the read path — not `IOException`/`UnauthorizedAccessException` (those are real faults that must not be masked as "not authenticated"). Empty files surface as `JsonException`, so they are covered.
+- AOT: no new reflection/dynamic codegen. After changes, `dotnet publish -c Release` must show **no** IL2026/IL3050 warnings (`CLAUDE.md`).
+- No public API or call-site changes; no `README.md` change required (internal robustness only — no user-facing CLI surface change).
+- TUnit filtering uses `--treenode-filter` glob syntax, NOT `--filter`.
+
+---
+
+## File Structure
+
+| File | Responsibility | Action |
+|---|---|---|
+| `src/Capacitor.Cli.Core/Auth/TokenStore.cs` | Add `ReadTokensAsync` reader; route both `LoadAsync` overloads through it; unique temp name + `finally` cleanup in `SaveAsync` | Modify |
+| `test/Capacitor.Cli.Tests.Unit/TokenStoreProfileTests.cs` | Add resilience + concurrency regression tests (reuses the class's `KCAP_CONFIG_DIR` temp-dir harness, `[Before(Test)]` cleanup, and `MakeTokens`) | Modify |
+
+**Reference files to read before starting:**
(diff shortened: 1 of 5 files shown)


realtonyyoung:
## Re-review of 4c78fe0

**Critical**
None.

**Important**
- `src/Capacitor.Cli.Core/Auth/TokenStore.cs:96`: the temp file is chmodded only after the token secret has already been written. `File.WriteAllTextAsync(tempPath, ...)` creates and fills the file using the process umask, then lines 100-101 change it to owner-only. On a typical/default umask of `022` (also what this checkout reports), the temp can briefly exist as `0644` with live token material before the chmod runs. That leaves exactly the local-read window the fix was intended to close for crash-leaked temps. Concrete fix: create the temp empty with `FileMode.CreateNew`/`FileShare.None`, set `0600` before writing any bytes (or use `FileStreamOptions.UnixCreateMode = UserRead | UserWrite` where available), then serialize/write through that stream. Keep the final-file chmod at lines 112-113 as defense in depth.

**Minor**
- `test/Capacitor.Cli.Tests.Unit/TokenStoreProfileTests.cs:23` / `test/Capacitor.Cli.Tests.Unit/SetupCommandTests.cs:135`: the new `Before(Test)` cleanup deletes the shared `AppConfig.GetConfigPath()` before every TokenStore profile test, but `SetupCommandTests.Setup_save_profile_config_round_trips_active_profile` writes and rereads that same assembly-wide `config.json` without sharing the `NotInParallel(nameof(TokenStoreProfileTests))` key. Because `PathHelpers.ConfigDir` is static for the whole test process, these can run concurrently: TokenStore cleanup can delete `config.json` between Setup's save and load, or Setup can leave a non-default active profile while a TokenStore test is resolving it. Concrete fix: put all tests that touch the shared `AppConfig.GetConfigPath()` under one common not-in-parallel key, or stop deleting the global config in TokenStore cleanup and instead make the active-profile tests set up their required config in an isolated/serialized fixture.

**Prior findings status**
1. Corrupt active profile must not fall back to legacy: **Resolved**. `ReadTokenFileAsync` returns `Missing` only for `FileNotFoundException`/`DirectoryNotFoundException`; corrupt/empty/null JSON is `Unusable`, and parameterless `LoadAsync()` falls back to legacy only when state is `Missing`.
2. Leaked temp secrets and logout cleanup: **Partially resolved**. Unique temp sweeps are present (`Delete(profile)` scoped by filename prefix, `DeleteAsync()` unscoped over `*.tmp` and not `*.lock`), and the final file mode is owner-only. The remaining issue is the pre-chmod exposure window above.
3. Read TOCTOU: **Resolved**. The `File.Exists` precheck is gone; absent file/dir are caught as `Missing`, while permission/other IO errors still propagate.
4. Test gaps: **Partially resolved**. The new tests cover corrupt-active/no-legacy, missing-active/legacy, temp sweeps, failed publish cleanup, and final Unix mode, but they miss the temp creation permission window and introduce the shared-config race above.

**Verification notes**
- Profile-name validation still runs through `ProfileTokenPath(profile)` in parameterless `LoadAsync()` before reading the active profile.
- Scoped temp sweep uses a literal filename prefix (`{profile}.json.`), not a glob or regex, so metacharacters cannot widen the match after profile validation; unscoped logout sweep targets only `*.tmp`, not live `.json` or `.lock` files.
- Happy-path save/load, legacy migration-on-save, and the refresh lock shape are unchanged except for the read helper semantics.
- Focused TUnit command with `~/.dotnet/dotnet run --project test/Capacitor.Cli.Tests.Unit/Capacitor.Cli.Tests.Unit.csproj --treenode-filter "/*/*/TokenStoreProfileTests/*"` produced no test output after ~90s in this environment; I interrupted it and the SDK reported build cancellation/failure. AOT grep was run under a 300s cap and timed out with no IL warning lines emitted, so I am not treating AOT as failed, just not completed locally.

Ready to merge: **With fixes** — the tri-state and cleanup direction is right, but the temp-file permission window is still a real token exposure and the new config cleanup can make the unit suite flaky.

realtonyyoung:
## Fixed in `6dac3fd`

**Important — temp permission window:** `SaveAsync` now creates the temp owner-only from the first byte via `FileStreamOptions.UnixCreateMode = UserRead | UserWrite` (Unix; ignored on Windows) instead of `WriteAllText`-then-chmod. No window where the token secret exists at `0644`, and a crash-leaked temp is already `0600`. The rename carries the mode to the final file; the post-move chmod stays as defense-in-depth.

**Minor — shared-config test race:** `SetupCommandTests.Setup_save_profile_config_round_trips_active_profile` now carries `[NotInParallel(nameof(TokenStoreProfileTests))]`, so it can't run concurrently with the `TokenStoreProfileTests` cleanup that resets the process-wide `config.json`. It was the only shared-config writer/reader missing the key.

Full suite 1866/1866; AOT publish clean (no IL2026/IL3050).

PR #218, opened 2026-07-01
docs(review-flows skill): daemon prerequisite, roles, and failure handling

realtonyyoung:
## What

Enhances the `review-flows` skill (`kcap/skills/review-flows/SKILL.md`) so the driving agent knows what has to be true for a flow to work, and what to do when it isn't.

- **Roles** — the driver calls these tools; the **reviewer is a separate hosted Codex agent** the server launches on a `kcap` daemon (never configured from the driver).
- **Prerequisites** — `start_review_flow` needs `kcap login` **and** a running daemon with *this* repo checked out, or it errors instead of returning a `flow_run_id`.
- **Failure handling** — on a no-daemon / ambiguous-daemon error, surface the fix (`kcap daemon start -d` with the target repo) rather than retrying blindly; don't start a flow (which spends real reviewer compute) unprompted.

## Why

Setting the feature up end-to-end surfaced that these prerequisites were undocumented on the agent-facing surface — the skill explained the loop mechanics but not the daemon dependency or the driver/reviewer split, so an agent hitting a no-daemon error had no guidance.

## Scope / relationship to other work

Doc-only; **no CLI surface change** (no new command/flag), so no README sync is required by the repo convention.

Complements **#217 (AI-1056)**, which auto-registers the flows MCP server for Claude Code and rewrites the README/registration guidance. This PR touches only `SKILL.md` (which #217 does not), so the two don't conflict.

Linear: **AI-774** (feature), **AI-1056** (auto-register follow-up).

### DIFF
--- a/kcap/skills/review-flows/SKILL.md
+++ b/kcap/skills/review-flows/SKILL.md
@@ -28,6 +28,18 @@ Once the user has explicitly opted into a flow (see above), pick the `kind`:
 - Spec or design document → `kind: "spec-review"`
 - Code changes or a pull request → `kind: "code-review"`
 
+## Prerequisites
+
+A flow hands your work to a **hosted reviewer** the daemon launches for you (its vendor and model come from the flow definition — you don't pick or configure the reviewer). For `start_review_flow` to succeed, two things must be true:
+
+1. You are logged in (`kcap login`).
+2. A `kcap` daemon is running with **this** repo checked out — that daemon hosts the reviewer.
+
+If `start_review_flow` errors, do **not** retry blindly — the two cases differ:
+
+- **No daemon** — tell the user to start one with the target repo checked out (`kcap daemon start -d`), then start the flow again.
+- **Ambiguous match** ("multiple daemons/checkouts match this repo") — retrying won't help: the tool exposes no `daemon_name`/`repo_path`, so it can't disambiguate when several checkouts of the repo are registered (e.g. multiple git worktrees). Surface this to the user rather than looping (tracked in AI-1112).
+
 ## If the flows MCP tools are not loaded
 
 If `start_review_flow` / `submit_review_round` are not among the tools available in this session, do NOT try to obtain them:


kcap/skills/review-flows/SKILL.md
```diff
@@ -11,6 +11,19 @@ description: >-
...
+You are the **driver** — you call these tools and act on the findings. The **reviewer is a separate hosted Codex agent** the server launches for you on a connected `kcap` daemon. You never configure or talk to the reviewer directly; you only drive the loop.
+
+## Prerequisites
+
+`start_review_flow` needs both of these, or it will return an error instead of a `flow_run_id`:
+
+1. **You are logged in** (`kcap login`).
+2. **A `kcap` daemon is running with *this* repo checked out** — that daemon launches the reviewer.
+
+If `start_review_flow` reports **no daemon** (or an **ambiguous** match), do **not** retry blindly. Tell the user to start a daemon with the target repo checked out (`kcap daemon start -d`), then start the flow again. The reviewer consumes real compute, so only start a flow when the user has asked for a review.
```
realtonyyoung: This remediation is right for the no-daemon case, but it is misleading for an ambiguous daemon match. Since `start_review_flow` exposes no `daemon_name` / `repo_path` argument and the MCP wrapper sends `daemon_name: null` with only the detected repo root, the agent cannot disambiguate by retrying; asking the user to start another daemon with the same repo also will not resolve ambiguity. Could this split the cases, e.g. no daemon => start `kcap daemon start -d`; ambiguous => ask the user to leave exactly one matching daemon available (stop/rename the extra daemon or otherwise resolve the duplicate) and then retry?

realtonyyoung (commented):
Found one documentation correctness issue in the new failure-handling guidance. The rest of the added sections look consistent with the flow tool surface, the daemon launch path, and PR #217 auto-registration wording.

realtonyyoung:
Fixed in `1039c4d`. The failure-handling guidance now splits the two cases instead of prescribing "start a daemon + retry" for both:

- **No daemon** → start one (`kcap daemon start -d`) and retry.
- **Ambiguous match** ("multiple daemons/checkouts match this repo") → retrying won't help; the tool exposes no `daemon_name`/`repo_path`, so the agent can't disambiguate. The skill now says to surface it to the user rather than loop, and references **AI-1112** (the tracked fix: server-side auto-select by the requester's repo root).

Docs-only; no CLI surface change.

kcap/skills/review-flows/SKILL.md
realtonyyoung: Split the two cases in `1039c4d`: no-daemon → start one and retry; ambiguous multi-checkout match → retrying cannot disambiguate (no `daemon_name`/`repo_path`), so the skill now says to surface it (AI-1112) instead of looping.

realtonyyoung:
Rebased onto current `main` to clear the conflict (`main` moved a lot: #238 opt-in gating, #235 hosted-reviewer misfire guard, and #233/#234 vendor-uniform hosting all rewrote `SKILL.md`).

Given those, I **trimmed this PR** rather than force my original text back in:
- Dropped the **"Roles"** section — now redundant with `main`'s intro + "When NOT to use" gating.
- Dropped the **Codex-specific** wording — post-#234 the reviewer's vendor/model come from the flow definition, not hardcoded Codex.

What remains is the still-unique, vendor-neutral bit `main` doesn't cover: a **Prerequisites** section (`start_review_flow` needs `kcap login` + a daemon with *this* repo checked out) and split failure handling — **no-daemon** (start one, retry) vs **ambiguous multi-checkout match** (can't disambiguate; surface it, **AI-1112**).

Now `MERGEABLE` (+12/−0, one file). Force-pushed; earlier resolved threads were anchored to the pre-rebase text.

realtonyyoung:
Closing as stale — superseded by the broader review-flows docs work tracked separately. Can reopen if this specific skill-prereqs doc is still wanted.

PR #224, opened 2026-07-01
Defer kcap mcp review auth client until first tool call

### LINKED ISSUE #223: kcap mcp review does eager auth at startup — avoidable per-session network + stderr
## Problem

`kcap-review` is auto-registered via the plugin manifest (`.mcp.json` for Claude Code, `.codex-mcp.json` for Codex), so the host spawns `kcap mcp review` for **every** session. But `McpReviewServer.RunCoreAsync` eagerly calls `HttpClientExtensions.CreateAuthenticatedClientAsync` at startup, which:

- does a `GET /auth/config` round-trip + token load, and
- prints a re-auth hint to **stderr** ("Not authenticated…" / "token has expired…") when not logged in.

So every session pays that network/token work — and unauthenticated users get stderr noise — even when no review tool is ever invoked. `initialize` / `tools/list` need no auth.

This is the same issue already fixed for `kcap mcp sessions` (lazy client, with a comment citing the auto-register manifest) and for `kcap mcp flows` (kcap-cli #217). `review` was missed; it predates both.

## Fix

Defer client creation to the first `tools/call` via `Lazy<Task<HttpClient>>`, mirroring `McpSessionsServer` / `McpFlowsServer`. Repo/PR auto-detection stays at startup (local git only).

Add integration coverage for the review MCP server (it had none): initialize, tools/list, and a regression asserting `initialize` + `tools/list` trigger zero `GET /auth/config` calls.

alexeyzimarev:
## Problem

`kcap-review` auto-registers via the plugin manifest (`.mcp.json` + `.codex-mcp.json`), so the host spawns `kcap mcp review` for **every** session. `McpReviewServer.RunCoreAsync` eagerly called `CreateAuthenticatedClientAsync` at startup — which does a `GET /auth/config` round-trip + token load, and prints a re-auth hint to **stderr** when not logged in. Every session paid that cost (and unauthenticated users saw stderr noise) even when no review tool was invoked. `initialize` / `tools/list` need no auth.

Same issue already solved for `kcap mcp sessions` (lazy client) and `kcap mcp flows` (#217). `review` predates both and was missed — this is a **pre-existing** issue, independent of #217/#220.

## Change

- Defer client creation to the first `tools/call` via `Lazy<Task<HttpClient>>`, mirroring `McpSessionsServer` / `McpFlowsServer`. `initialize` / `tools/list` stay local-only; PR auto-detection stays at startup (local git only); dispose in a `finally`.
- Add the review MCP server's **first integration tests** (it had none): `initialize`, `tools/list`, and a regression asserting `initialize` + `tools/list` make **zero** `GET /auth/config` calls.

## Verification

- Review integration suite: **3 passed** (incl. the regression, which would fail on the old eager code).
- `dotnet build` clean; `dotnet publish -c Release` — no IL3050/IL2026 AOT warnings.

## Issue references

Closes #223 (a Linear issue will auto-import from it). Same class of fix as #217 (flows) / the sessions lazy client.

### DIFF
--- a/src/Capacitor.Cli/Commands/McpReviewServer.cs
+++ b/src/Capacitor.Cli/Commands/McpReviewServer.cs
@@ -41,6 +41,26 @@ static class McpReviewServer {
         // McpSessionsServer / McpFlowsServer.
         var clientLazy = new Lazy<Task<HttpClient>>(() => HttpClientExtensions.CreateAuthenticatedClientAsync(baseUrl));
 
+        // Validate the server_url shape once, locally (pure string check — no network, token,
+        // or stderr). Used to fail gracefully instead of hard-exiting mid-request (below).
+        var urlOk = HttpClientExtensions.IsAcceptableUrl(baseUrl);
+
+        // Guarded tool dispatch: never let the stdio JSON-RPC loop die on one bad request.
+        // An unusable server_url would otherwise reach EnsureAbsolute inside the lazy
+        // auth-client factory, which hard-exits the process (Environment.Exit(2)) mid-request;
+        // and an unexpected client-creation/token failure would bubble out of the loop. Return
+        // a JSON-RPC tool error in both cases so the server keeps serving.
+        async Task<string> DispatchToolCallAsync(JsonNode callId, JsonObject callRequest) {
+            if (!urlOk)
+                return BuildToolResult(callId, HttpClientExtensions.SchemeMissingHint, isError: true);
+
+            try {
+                return await HandleToolCallAsync(callId, callRequest, await clientLazy.Value, baseUrl, sessionDefault);
+            } catch (Exception ex) {
+                return BuildToolResult(callId, $"Error: {ex.Message}", isError: true);
+            }
+        }
+
         await using var stdin  = Console.OpenStandardInput();
         await using var stdout = Console.OpenStandardOutput();
         using var       reader = new StreamReader(stdin, Encoding.UTF8);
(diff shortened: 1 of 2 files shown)


alexeyzimarev:
Thanks @qodo-code-review — assessed both findings:

**1. Auth init crashes tools/call (Reliability) — addressed in this PR (commit 04b9f448b).** Verified the failure surface first: `DiscoverProviderAsync` already *swallows* network errors (`HttpClientExtensions.cs` — `catch { }` → token fallback), and "not logged in / expired" returns a client (no throw) so the tool call just gets a 401 that `DispatchAsync` catches. The one genuinely-uncatchable path is a scheme-less `server_url` → `EnsureAbsolute` → `Environment.Exit(2)` (a deliberate CLI-wide fail-fast, and `try/catch` can't intercept `Environment.Exit`).

Fix: validate `server_url` shape once at startup with the pure local `IsAcceptableUrl` (no network/token/stderr) and route `tools/call` through a guarded dispatcher that returns a JSON-RPC tool error for an unusable URL **or** any unexpected exception — so the server never hard-exits or drops the loop, and keeps serving. Regression test added (`Tool_call_with_invalid_server_url_returns_error_and_server_survives`). Applied uniformly to the sibling servers too: flows in #217, sessions in #226.

**2. Git command failures ignored (Reliability) — declined (test-only, intentional).** `RunGit` is in the test fixture and is copied verbatim from `McpFlowsServerTests`/`McpSessionsServerTests`. These tests don't depend on git succeeding — `DetectPrFromGitAsync` failure is caught and none of them need a detected PR. Changing just this copy would diverge from the established helpers; not worth it.

Review integration suite (4) green; no IL3050/IL2026 AOT warnings.

PR #226, opened 2026-07-01
Harden kcap mcp sessions dispatch: guard invalid URL + catch per-request

### LINKED ISSUE #225: Harden MCP server dispatch: invalid server_url hard-exits mid tools/call; unexpected errors kill the loop
## Problem

All three auto-registered MCP servers (`sessions`, `flows`, `review`) now create the authenticated `HttpClient` lazily on first `tools/call`. Two gaps in the tool-dispatch path:

1. **Invalid `server_url` hard-exits mid-request.** A scheme-less/invalid `server_url` reaches `EnsureAbsolute` inside the auth-client factory (`HttpClientExtensions.DiscoverProviderAsync` → `EnsureAbsolute` → `Environment.Exit(2)`), so the first `tools/call` terminates the process instead of returning an error. (A `try/catch` can't catch `Environment.Exit`.)
2. **Unexpected exceptions kill the stdio loop.** The `tools/call` arm awaits `clientLazy.Value` and dispatches with no surrounding `catch`, so any unexpected failure (e.g. a token-refresh throw) bubbles out of the JSON-RPC loop and the server closes stdout without responding.

Both are pre-existing / latent; raised by Qodo on #224.

## Fix

Uniformly across `sessions` / `flows` / `review`:
- Validate the `server_url` shape once at startup with the pure, local `IsAcceptableUrl` (no network/token/stderr).
- Route `tools/call` through a guarded dispatcher that returns a JSON-RPC tool error for an unusable URL or any unexpected exception — keeping the server alive to serve subsequent requests. `initialize` / `tools/list` stay local-only.

Add an integration test per server: invalid `server_url` → graceful tool error + server survives.

## Rollout

Because each server's lazy-client change lives in a different place, the hardening lands alongside it:
- **flows** → #217
- **review** → #224
- **sessions** → this PR (its lazy client is already on `main`)

alexeyzimarev:
## Problem

The three auto-registered MCP servers now create the authenticated `HttpClient` lazily on first `tools/call`. Two dispatch-path gaps (raised by Qodo on #224):

1. **Invalid `server_url` hard-exits mid-request** — reaches `EnsureAbsolute` → `Environment.Exit(2)` inside the lazy auth-client factory (uncatchable), terminating the process on first `tools/call` instead of returning an error.
2. **Unexpected exceptions kill the stdio loop** — the `tools/call` arm has no surrounding `catch`, so an unexpected failure bubbles out and the server closes stdout without a JSON-RPC response.

## Change (sessions — the third server)

- Validate `server_url` once at startup via the pure, local `IsAcceptableUrl` (no network/token/stderr).
- Route `tools/call` through a guarded dispatcher returning a JSON-RPC tool error for an unusable URL or any unexpected exception; the server keeps serving. `initialize` / `tools/list` stay local-only.
- Test: `Tool_call_with_invalid_server_url_returns_error_and_server_survives`.

This is the same hardening applied to **flows (#217)** and **review (#224)** — landing per-server since each one's lazy-client change lives in a different place; sessions' lazy client is already on `main`, so this is standalone.

## Verification

- Sessions integration suite: **8 passed** (incl. the new graceful-failure test).
- `dotnet build` clean; `dotnet publish -c Release` — no IL3050/IL2026 AOT warnings.

Closes #225.

### DIFF
--- a/src/Capacitor.Cli/Commands/McpSessionsServer.cs
+++ b/src/Capacitor.Cli/Commands/McpSessionsServer.cs
@@ -25,6 +25,26 @@ static class McpSessionsServer {
         // don't need auth, so we keep startup local-only.
         var clientLazy = new Lazy<Task<HttpClient>>(() => HttpClientExtensions.CreateAuthenticatedClientAsync(baseUrl));
 
+        // Validate the server_url shape once, locally (pure string check — no network, token,
+        // or stderr). Used to fail gracefully instead of hard-exiting mid-request (below).
+        var urlOk = HttpClientExtensions.IsAcceptableUrl(baseUrl);
+
+        // Guarded tool dispatch: never let the stdio JSON-RPC loop die on one bad request. An
+        // unusable server_url would otherwise reach EnsureAbsolute inside the lazy auth-client
+        // factory, which hard-exits the process (Environment.Exit(2)) mid-request; and an
+        // unexpected client-creation/token failure would bubble out of the loop. Return a
+        // JSON-RPC tool error in both cases so the server keeps serving.
+        async Task<string> DispatchToolCallAsync(JsonNode callId, JsonObject callRequest) {
+            if (!urlOk)
+                return BuildToolResult(callId, HttpClientExtensions.SchemeMissingHint, isError: true);
+
+            try {
+                return await HandleToolCallAsync(callId, callRequest, await clientLazy.Value, baseUrl, cwdRepoHash);
+            } catch (Exception ex) {
+                return BuildToolResult(callId, $"Error: {ex.Message}", isError: true);
+            }
+        }
+
         await using var stdin  = Console.OpenStandardInput();
         await using var stdout = Console.OpenStandardOutput();
         using var       reader = new StreamReader(stdin, Encoding.UTF8);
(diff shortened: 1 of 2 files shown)


alexeyzimarev:
Thanks @qodo-code-review — both fixed in c80660474.

**1. Faulted lazy client sticks (Reliability) — fixed.** Good catch: a `Lazy<Task<HttpClient>>` caches a *faulted* task, so a single transient client-creation failure would make every later `tools/call` re-throw for the rest of the session — my guard had turned a would-be crash (which the host restarts → retries) into a permanently-erroring server. Replaced the `Lazy<Task>` with an on-demand nullable `HttpClient?` created inside the dispatcher (`client ??= await …`); a failed attempt leaves it null so the next call retries. No lock needed — the stdio loop handles one request at a time. Startup stays local-only (initialize/tools-list never create it), so the deferral property is preserved.

**2. Exception text disclosure (Security) — fixed.** The catch-all now logs the full exception to `stderr` and returns a generic `"internal error handling the request."` to the client instead of `ex.Message`. It's a safety net for *unexpected* exceptions only — the expected ones (`ArgumentException`, `HttpRequestException`, 401/HTTP errors) are still handled inside `HandleToolCallAsync` with their specific, safe messages.

Applied the same two fixes to the sibling servers for consistency: flows in #217, review in #224.

Sessions integration suite (8) green; no IL3050/IL2026 AOT warnings.

PR #242, opened 2026-07-02
Cursor: attribute sessions to their repo + nest subagents under the parent

realtonyyoung:
Fixes two Cursor ingest gaps. Pairs with the **kcap-server** PR (shared repo recorder + Cursor normalizer id-scoping for subsessions) — both halves ship together. Tracks Linear **AI-1152** (repo attribution), **AI-1153** (subagent nesting), and the subagent half of **AI-1154** (backfill).

## AI-1152 — repo attribution
Cursor sessions never grouped under a repository (only "All repos") because the sidebar keys on `repo_hash`, set only by a `RepositoryDetected` event, which needs a git-detected `owner`/`repo_name`. Every other vendor's hook path attaches a `repository` node via `RepositoryDetection.EnrichWithRepositoryInfo` — but that reads `cwd`, and Cursor payloads carry `workspace_roots` instead, and `CursorHookCommand` never called it.

- Add `RepositoryDetection.EnrichWithRepositoryInfoFromCwd` + extract `BuildRepositoryNode` (shared).
- `CursorHookCommand` enriches `sessionStart` from `workspace_roots[0]` (budget-bounded, fail-open).
- `CursorImportSource` detects the repo from the workspace folder and attaches it to the synthetic `sessionStart` (also fixes historical import/backfill).

## AI-1153 — nest subagents under the parent
A Cursor subagent (`Task`/`Agent`) runs as its **own** session/transcript with **no explicit parent link** — the only in-data signal is that the child's first `user_query` (minus the `<user_query>` wrapper) is byte-identical to the parent's `Task` `input.prompt`.

- New pure `CursorSubagentCorrelator.Correlate` recovers the link by SHA-256 prompt hash (self-link guarded).
- `CursorImportSource.ClassifyAsync` builds the map and stamps `ParentSessionId`/`SubagentType`; `ImportSessionAsync` routes a detected child through `ImportSubagentAsync` — `/hooks/subagent-start` (session_id=parent, agent_id=child) + transcript batches with `agent_id`=child (→ `AgentSubsession-{parent}-{child}`) + `/hooks/subagent-stop` — and emits **no** standalone session lifecycle. This mirrors how Gemini/OpenCode subagents nest, so no server read-model change is needed; the child never creates a top-level card.

## Tests
`CursorSubagentCorrelatorTests` (3), `CursorImportSourceTests` incl. repo-attach + subagent-routing (34 total). AOT publish clean (no IL2026/IL3050).

### DIFF
--- a/src/Capacitor.Cli/Commands/CursorHookCommand.cs
+++ b/src/Capacitor.Cli/Commands/CursorHookCommand.cs
@@ -111,6 +111,25 @@ public static class CursorHookCommand {
 
             if (sessionId is not null && DisabledSessions.IsDisabled(sessionId)) return 0;
 
+            // AI-1152: attach a `repository` node on sessionStart so the session groups
+            // under its repo in the sidebar. Cursor payloads carry `workspace_roots`
+            // rather than `cwd`, so the generic EnrichWithRepositoryInfo (which reads
+            // `cwd`) can't be used — detect from workspace_roots[0] instead. Bounded by
+            // the remaining dispatcher budget; fail-open (git detection is cached).
+            if (eventName == "sessionStart") {
+                var workspaceRoot = node["workspace_roots"] is JsonArray roots && roots.Count > 0
+                    ? roots[0]?.GetValue<string>()
+                    : null;
+                if (!string.IsNullOrEmpty(workspaceRoot)) {
+                    var remaining = budgetTotal - sw.Elapsed;
+                    if (remaining > TimeSpan.Zero) {
+                        node = JsonNode.Parse(
+                            await RepositoryDetection.EnrichWithRepositoryInfoFromCwd(node.ToJsonString(), workspaceRoot, remaining)
+                        ) ?? node;
+                    }
+                }
+            }
+
             var normalized = node.ToJsonString();
 
             if (sessionId is not null) {
--- a/src/Capacitor.Cli/Commands/CursorHookCommand.cs
+++ b/src/Capacitor.Cli/Commands/CursorHookCommand.cs
@@ -117,9 +117,12 @@ public static class CursorHookCommand {
             // `cwd`) can't be used — detect from workspace_roots[0] instead. Bounded by
             // the remaining dispatcher budget; fail-open (git detection is cached).
             if (eventName == "sessionStart") {
-                var workspaceRoot = node["workspace_roots"] is JsonArray roots && roots.Count > 0
-                    ? roots[0]?.GetValue<string>()
-                    : null;
+                // Safe extract: workspace_roots[0] may be absent or a non-string; GetValue<string>
+                // would throw and (via the outer catch) drop the whole sessionStart hook.
+                string? workspaceRoot = null;
+                if (node["workspace_roots"] is JsonArray roots && roots.Count > 0
+                 && roots[0] is JsonValue wv && wv.TryGetValue<string>(out var wr))
+                    workspaceRoot = wr;
                 if (!string.IsNullOrEmpty(workspaceRoot)) {
                     var remaining = budgetTotal - sw.Elapsed;
                     if (remaining > TimeSpan.Zero) {
--- a/src/Capacitor.Cli/Commands/CursorImportSource.cs
+++ b/src/Capacitor.Cli/Commands/CursorImportSource.cs
@@ -196,6 +196,16 @@ internal sealed class CursorImportSource : IImportSource {
         var repoCache    = new Dictionary<string, string?>(StringComparer.Ordinal); // cwd → "owner/repo" or null
         var hasExcludes  = ctx.ExcludedRepos is { Count: > 0 };
 
+        // AI-1153: correlate subagent (child) sessions to their parent by prompt-hash across
+        // all discovered transcripts. A child is then ingested under the parent's
+        // AgentSubsession stream (subagent-start → transcript-with-agent_id → subagent-stop)
+        // rather than as a separate top-level session.
+        var subagentLinks = CursorSubagentCorrelator.Correlate(
+            sessions.Select(s => (
+                s.SessionId,
+                s.SourceMeta!.TryGetValue("TranscriptPath", out var tp) && tp is string p ? p : ""
+            )));
+
         foreach (var s in sessions) {
             var transcriptPath = (string)s.SourceMeta!["TranscriptPath"]!;
 
@@ -287,7 +297,12 @@ internal sealed class CursorImportSource : IImportSource {
                 ExcludedRepoKey = excludedRepoKey,
                 ExcludedPathKey = excludedPathKey,
                 TotalLines      = nonBlankCount,
-                SourceMeta      = s.SourceMeta,
+                SourceMeta      = subagentLinks.TryGetValue(s.SessionId, out var link)
+                    ? new Dictionary<string, object?>(s.SourceMeta!) {
+                        ["ParentSessionId"] = link.ParentSessionId,
+                        ["SubagentType"]    = link.SubagentType,
+                    }
+                    : s.SourceMeta,
             });
         }
 
@@ -299,6 +314,13 @@ internal sealed class CursorImportSource : IImportSource {
             ImportContext                       ctx,
             CancellationToken                   ct
         ) {
+        // AI-1153: a correlated subagent child is ingested under its parent's
+        // AgentSubsession stream instead of as a standalone top-level session.
+        if (classification.SourceMeta!.TryGetValue("ParentSessionId", out var parentObj)
+         && parentObj is string parentSessionId && !string.IsNullOrEmpty(parentSessionId)) {
+            return await ImportSubagentAsync(classification, parentSessionId, ctx, ct);
+        }
+
         var transcriptPath = (string)classification.SourceMeta!["TranscriptPath"]!;
 
         if (!File.Exists(transcriptPath)) return ImportOutcome.Failed;
@@ -309,6 +331,20 @@ internal sealed class CursorImportSource : IImportSource {
 
         var (createdUtc, modifiedUtc) = TryGetTranscriptTimes(transcriptPath);
 
+        // AI-1152: detect the repo from the workspace folder so the synthetic
+        // sessionStart carries a `repository` node → server emits RepositoryDetected
+        // and the (historical/backfilled) session groups under its repo. Fail-open:
+        // a non-git workspace or detection error leaves it null and unattributed.
+        JsonObject? repositoryNode = null;
+        if (workspaceFolder is not null) {
+            try {
+                var repo = await _repoDetector(workspaceFolder);
+                if (repo is not null) repositoryNode = RepositoryDetection.BuildRepositoryNode(repo);
+            } catch {
+                repositoryNode = null;
+            }
+        }
+
         // sessionStart MUST succeed before transcript advances the server
         // watermark — otherwise a transient lifecycle failure plus a successful
         // transcript would leave the session permanently lifecycle-less
@@ -318,7 +354,7 @@ internal sealed class CursorImportSource : IImportSource {
         // (canonical event ids are deterministic — AI-731).
         var startOk = await PostSyntheticHookAsync(
             ctx.HttpClient, ctx.BaseUrl, "session-start/cursor",
-            BuildSessionStartPayload(classification.SessionId, workspaceFolder, transcriptPath, createdUtc),
+            BuildSessionStartPayload(classification.SessionId, workspaceFolder, transcriptPath, createdUtc, repositoryNode),
             ct);
         if (!startOk) return ImportOutcome.Failed;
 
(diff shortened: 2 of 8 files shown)


src/Capacitor.Cli/Commands/CursorImportSource.cs
```diff
@@ -363,8 +417,97 @@ CancellationToken                   ct
...
+                sessionId:  parentSessionId,
+                filePath:   child.TranscriptPath,
+                agentId:    agentId,
+                startLine:  startLine,
+                vendor:     Vendor);
+        } catch {
+            return false;
+        }
+
+        return await PostSyntheticHookAsync(ctx.HttpClient, ctx.BaseUrl, "subagent-stop", new JsonObject {
```
realtonyyoung, line 508: [P1] This synthetic stop body is missing the required fields on `SubagentStopHook` (`stop_hook_active`, `agent_transcript_path`, and `last_assistant_message`; the existing Gemini/OpenCode builders send them). A real `/hooks/subagent-stop` bind can reject the request before `HandleSubagentStop`, so every correlated child can make the parent import fail after sending the child transcript and before the parent session-end. Please include the full stop payload shape here, and add `strict: true` if this import is meant to fail-closed on lifecycle write failures.

src/Capacitor.Cli/Commands/CursorImportSource.cs
```diff
@@ -363,8 +417,97 @@ CancellationToken                   ct
...
+
+        try {
+            await SessionImporter.SendTranscriptBatches(
+                httpClient: ctx.HttpClient,
+                baseUrl:    ctx.BaseUrl,
+                sessionId:  parentSessionId,
+                filePath:   child.TranscriptPath,
+                agentId:    agentId,
+                startLine:  startLine,
+                vendor:     Vendor);
```
realtonyyoung: [P1] `SendTranscriptBatches` is being called in its default non-strict mode here. That helper ignores non-success transcript POST responses and swallows transport `HttpRequestException`s unless `failOnError: true`, so this path can still post `subagent-stop` and then the parent `session-end` after the child transcript was not accepted. That breaks the new rerun-repairs-it contract and can leave an empty completed subagent while the import reports success. Please pass `failOnError: true` for the child transcript send, and serialize server-side `strict` if normalization failures should also be fatal.

src/Capacitor.Cli/Commands/CursorImportSource.cs
```diff
@@ -196,6 +196,27 @@ CancellationToken                 ct
...
+        // stream *by the parent's own import* (subagent-start → transcript-with-agent_id →
+        // subagent-stop, before the parent's session-end) rather than as a separate top-level
+        // session. Doing it inside the parent import guarantees the parent's SessionEnded stays
+        // the last event on its stream — a subagent-start posted after a parallel parent's
+        // session-end would reactivate the ended parent (SubagentStarted is a reactivation event).
+        var pathBySession = new Dictionary<string, string>(StringComparer.Ordinal);
+        foreach (var s in sessions)
+            pathBySession[s.SessionId] = s.SourceMeta!.TryGetValue("TranscriptPath", out var tp) && tp is string p ? p : "";
+
+        var subagentLinks    = CursorSubagentCorrelator.Correlate(pathBySession.Select(kv => (kv.Key, kv.Value)));
```
realtonyyoung, line 210: [P2] Correlation only sees this already-discovered `sessions` slice. Cursor discovery applies `--session`, `--cwd`, and `--since` before this point, so an incremental import whose cutoff lands between the parent start and child start, or a targeted import of the child id, will not see the parent Task prompt. In that case the child is not stamped `IsSubagentChild` and can be imported as a top-level Cursor session, which reintroduces AI-1153. Consider scanning the unfiltered Cursor transcript set for correlation metadata, or explicitly pulling correlated parents/children into the routed import plan before filtering prevents the parent-owned import.

src/Capacitor.Cli/Commands/CursorSubagentCorrelator.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        foreach (var (sid, path) in sessions) {
+            if (string.IsNullOrEmpty(path) || !File.Exists(path)) continue;
+            scanned.Add((sid, ScanFirstUserHash(path, out var tasks), tasks));
+        }
+
+        // promptHash → (parentSessionId, subagentType). First writer wins on the (rare)
+        // case of two parents Tasking an identical prompt.
+        var parentByPrompt = new Dictionary<string, (string Parent, string? SubType)>(StringComparer.Ordinal);
+        foreach (var s in scanned)
```
realtonyyoung, line 56: [P2] The collision policy here is “first writer wins”, but the caller feeds this from filesystem discovery order (`Directory.EnumerateDirectories`), which is not guaranteed to be stable. If two parents issue the same Task prompt, a re-run can pick a different parent and route the same child under a different `AgentSubsession-{parent}-{child}` stream, defeating the idempotency goal. Please sort by a stable tie-breaker before building `parentByPrompt` (for example transcript creation time plus session id, or session id alone) so “first” is deterministic.

src/Capacitor.Cli/Commands/CursorImportSource.cs
realtonyyoung: Fixed in 7391313 — the subagent-stop payload now sends the full `SubagentStopHook` shape (`stop_hook_active`, `agent_transcript_path`, `last_assistant_message`).

src/Capacitor.Cli/Commands/CursorImportSource.cs
realtonyyoung: Fixed in 7391313 — the child transcript send now passes `failOnError: true`, so a rejected/failed POST throws → `SendSubagentLifecycleAsync` returns false → the parent import returns `Failed` **before** session-end (fail-closed; re-run repairs). New test `parent_import_fails_when_a_subagent_transcript_post_is_rejected`.

src/Capacitor.Cli/Commands/CursorSubagentCorrelator.cs
realtonyyoung: Fixed in 7391313 — `Correlate` now orders sessions by session id before building the prompt→parent map, so 'first writer wins' is deterministic across runs. New test `duplicate_task_prompt_resolves_to_a_stable_parent_regardless_of_input_order`.

src/Capacitor.Cli/Commands/CursorImportSource.cs
realtonyyoung: Acknowledged as a real limitation for **filtered/incremental** imports (`--session`/`--cwd`/`--since`). The default full import and the backfill path discover both parent and child, so they correlate correctly and this isn't a regression. A proper fix (unfiltered correlation + pulling the parent into the plan, without dropping the child or reactivating an ended parent) is tracked in Linear AI-1156.

src/Capacitor.Cli/Commands/CursorImportSource.cs
```diff
@@ -363,8 +399,67 @@ CancellationToken                   ct
...
+        } catch {
+            return ImportOutcome.Failed;
+        }
+
+        var stopOk = await PostSyntheticHookAsync(ctx.HttpClient, ctx.BaseUrl, "subagent-stop", new JsonObject {
+            ["hook_event_name"] = "subagent_stop",
+            ["session_id"]      = parentSessionId,
+            ["agent_id"]        = agentId,
+            ["agent_type"]      = subagentType,
+        }, ct);
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in 7391313 — subagent-stop now sends the full `SubagentStopHook` shape (`stop_hook_active`, `agent_transcript_path`, `last_assistant_message`).

src/Capacitor.Cli/Commands/CursorSubagentCorrelator.cs
```diff
@@ -0,0 +1,122 @@
...
+    static string? ScanFirstUserHash(string path, out List<(string Hash, string? SubType)> tasks) {
+        tasks = [];
+        string? firstUserHash = null;
+
+        foreach (var line in File.ReadLines(path)) {
+            if (string.IsNullOrWhiteSpace(line)) continue;
+
+            JsonDocument doc;
+            try { doc = JsonDocument.Parse(line); } catch { continue; }
+
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in 7391313 — `Correlate` wraps each per-session scan in try/catch, so a single unreadable/malformed transcript is skipped instead of throwing out of `ClassifyAsync`. New test `a_bad_transcript_entry_is_skipped_without_aborting_correlation`.

src/Capacitor.Cli/Commands/CursorSubagentCorrelator.cs
```diff
@@ -0,0 +1,122 @@
...
+            if (string.IsNullOrEmpty(path) || !File.Exists(path)) continue;
+            scanned.Add((sid, ScanFirstUserHash(path, out var tasks), tasks));
+        }
+
+        // promptHash → (parentSessionId, subagentType). First writer wins on the (rare)
+        // case of two parents Tasking an identical prompt.
+        var parentByPrompt = new Dictionary<string, (string Parent, string? SubType)>(StringComparer.Ordinal);
+        foreach (var s in scanned)
+            foreach (var (hash, subType) in s.Tasks)
+                parentByPrompt.TryAdd(hash, (s.Sid, subType));
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in 7391313 — sessions are ordered by id before building `parentByPrompt`, so duplicate-prompt attribution is deterministic across runs (see the determinism test).

src/Capacitor.Cli/Commands/CursorHookCommand.cs
```diff
@@ -111,6 +111,25 @@ TimeSpan   budgetTotal
...
 
+            // AI-1152: attach a `repository` node on sessionStart so the session groups
+            // under its repo in the sidebar. Cursor payloads carry `workspace_roots`
+            // rather than `cwd`, so the generic EnrichWithRepositoryInfo (which reads
+            // `cwd`) can't be used — detect from workspace_roots[0] instead. Bounded by
+            // the remaining dispatcher budget; fail-open (git detection is cached).
+            if (eventName == "sessionStart") {
+                var workspaceRoot = node["workspace_roots"] is JsonArray roots && roots.Count > 0
+                    ? roots[0]?.GetValue<string>()
+                    : null;
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in 7391313 — `workspace_roots[0]` is now extracted with `JsonValue.TryGetValue<string>`, so a missing/non-string element can't throw and drop the whole sessionStart hook.

src/Capacitor.Cli/Commands/CursorImportSource.cs
```diff
@@ -363,8 +417,106 @@ CancellationToken                   ct
...
+            CancellationToken ct
+        ) {
+        if (string.IsNullOrEmpty(child.TranscriptPath) || !File.Exists(child.TranscriptPath))
+            return true; // missing child transcript — skip, non-fatal
+
+        var agentId      = child.SessionId; // the child session id doubles as the subagent id
+        var subagentType = string.IsNullOrEmpty(child.SubagentType) ? "task" : child.SubagentType!;
+
+        // subagent-start on the parent stream (creates AgentSubsession-{parent}-{child}).
+        var startOk = await PostSyntheticHookAsync(ctx.HttpClient, ctx.BaseUrl, "subagent-start", new JsonObject {
```
realtonyyoung, line 468: [P1] This `/hooks/subagent-start` body still does not satisfy the server `HookBase`: `SubagentStartHook` inherits required `cwd` in addition to `session_id`, `transcript_path`, and `hook_event_name`. The existing Gemini/OpenCode builders send `cwd` (often as an empty string). Because System.Text.Json treats C# `required` members as required during binding, this can be rejected before `HandleSubagentStart`, making every correlated child fail before transcript import. Please include `cwd` here (empty string is fine if Cursor has no better value).

src/Capacitor.Cli/Commands/CursorImportSource.cs
```diff
@@ -363,8 +417,106 @@ CancellationToken                   ct
...
+                startLine:   startLine,
+                vendor:      Vendor,
+                failOnError: true);
+        } catch {
+            return false;
+        }
+
+        // Full SubagentStopHook shape (mirrors the Gemini/OpenCode builders) — the server
+        // binds all fields, so an incomplete body can be rejected before HandleSubagentStop.
+        return await PostSyntheticHookAsync(ctx.HttpClient, ctx.BaseUrl, "subagent-stop", new JsonObject {
```
realtonyyoung, line 508: [P1] The stop body now has the stop-specific fields, but it still misses the required `HookBase` fields `transcript_path` and `cwd`. `agent_transcript_path` does not populate the base `transcript_path`, and existing Gemini/OpenCode builders send both fields. With the current shape `/hooks/subagent-stop` can be rejected before `HandleSubagentStop`, so the parent import fails after sending the child transcript and before its own `session-end`. Please include `transcript_path` (the child transcript path is fine) and `cwd` here too.

src/Capacitor.Cli/Commands/CursorImportSource.cs
```diff
@@ -363,8 +417,106 @@ CancellationToken                   ct
...
+            // rather than leaving an empty completed subagent while reporting success.
+            await SessionImporter.SendTranscriptBatches(
+                httpClient:  ctx.HttpClient,
+                baseUrl:     ctx.BaseUrl,
+                sessionId:   parentSessionId,
+                filePath:    child.TranscriptPath,
+                agentId:     agentId,
+                startLine:   startLine,
+                vendor:      Vendor,
+                failOnError: true);
```
realtonyyoung, line 501: [P2] `failOnError: true` only makes this helper throw when the HTTP response is non-success or the transport throws. The server only turns per-line write failures into a non-success response when `TranscriptBatch.Strict` is true, but the CLI `TranscriptBatch` model still does not serialize `strict`, so failed entries can return 200 and this parent import will continue to `subagent-stop` and `session-end`. If the child transcript is meant to fail closed, add a `Strict` field to the CLI batch and set it from `failOnError` (or an explicit parameter) for this path.

src/Capacitor.Cli/Commands/CursorImportSource.cs
realtonyyoung, line 468: [P2] Even after the required base fields are added, these imported subagent lifecycle hooks are still non-strict. The server `HandleSubagentStart/Stop` paths return 200 after `RecordAgentStartAsync`/`RecordAgentStopAsync` failures unless `strict` is true, so `PostSyntheticHookAsync` can return success and the parent import can proceed to `session-end` even though `SubagentStarted` or `SubagentCompleted` was not persisted. Since this inline import path is intended to fail closed and be repairable on re-run, please send `strict: true` on both the start and stop payloads.

src/Capacitor.Cli/Commands/CursorImportSource.cs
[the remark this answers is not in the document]
realtonyyoung: Addressed in b48b719 — subagent-start/-stop now include `cwd` ("") and the stop adds `transcript_path`, matching `SendAgentLifecycle` / the other vendors' wire shape. (Verified the server's `SubagentStartHook`/`SubagentStopHook` bind only agent_id/agent_type + the stop fields — `cwd` isn't required — but sending it keeps parity and is defensive.)

src/Capacitor.Cli/Commands/CursorSubagentCorrelator.cs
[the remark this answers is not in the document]
realtonyyoung: Addressed in b48b719 — when the same Task prompt is issued by two *distinct* parents the correlator now marks it ambiguous and drops the link (no guessed attribution); a single parent Tasking the same prompt more than once still links. Replaces the earlier order-by-id "first writer wins". New tests: `ambiguous_prompt_across_two_parents_yields_no_link…` and `same_parent_tasking_a_prompt_twice_is_not_ambiguous`.

src/Capacitor.Cli/Commands/CursorImportSource.cs
realtonyyoung: You're right — `HookBase` declares `cwd` and `transcript_path` as **required**, so the original subagent-start would 400. Fixed in b48b719: subagent-start now sends `cwd` + `transcript_path` (plus the required `session_id`/`hook_event_name`). (My earlier reply claiming cwd wasn't required was wrong — I'd only checked `SubagentStartHook`'s own fields, not the `HookBase` it inherits. Good catch.)

src/Capacitor.Cli/Commands/CursorImportSource.cs
realtonyyoung: Fixed in b48b719 — subagent-stop now sends the required `HookBase` `transcript_path` and `cwd` (distinct from `agent_transcript_path`), alongside the stop-specific fields.

src/Capacitor.Cli/Commands/CursorImportSource.cs
realtonyyoung: Fixed in 4ff9967 — the CLI `TranscriptBatch` now has a `strict` field and `PostTranscriptBatch` sets `Strict = failOnError`. So the child transcript send (`failOnError: true`) is strict → the server returns non-2xx on per-line normalization failures (not just HTTP/transport errors) → the parent import fails closed.

src/Capacitor.Cli/Commands/CursorImportSource.cs
realtonyyoung: Fixed in 4ff9967 — subagent-start and subagent-stop now send `strict: true`. `HandleSubagentStart`/`HandleSubagentStop` return 500 (not 200) when the `SubagentStarted`/`SubagentCompleted` write fails under strict, so `PostSyntheticHookAsync` sees the failure and the parent import aborts before session-end.

src/Capacitor.Cli/Commands/CursorImportSource.cs
realtonyyoung: Follow-up to the earlier `failOnError` fix: 4ff9967 also propagates server `strict` (`TranscriptBatch.Strict = failOnError`), so **server-side normalization** failures — not just HTTP/transport errors — now fail the import closed.

realtonyyoung:
[P1] This appears to revert the AI-30 bracketed-paste path back to raw text plus Enter, and the dedicated regression test is deleted in the same branch comparison. The previous code intentionally wrapped server-origin input in ESC[200~/ESC[201~ so large or multiline messages are consumed as one paste before the separate CR. With raw `WriteAsync(message)`, Codex/Claude can fold the CR into the still-ingesting paste/newline sequence and leave the prompt sitting in the composer instead of submitting. Please keep the bracketed-paste write (and its test) unless this PR adds an equivalent submit guarantee.

Note: I am posting this as a top-level comment because GitHub would not accept it inline; the file was not present in the authoritative `gh pr diff` patch even though it appeared in the local branch comparison.

realtonyyoung:
This is a stale-base diff artifact, not a real change — the PR does not touch the bracketed-paste path. Root cause: this branch was cut from the pinned submodule commit (`f72828c`), which was **3 commits behind** `kcap-cli` main — including AI-30 (#237). So a two-dot `git diff main..branch` showed AI-30's `AgentOrchestrator` bracketed-paste code (and its test) as "removed on the branch", even though the branch never modified those files (which is why it wasn't in the authoritative `gh pr diff`). On squash-merge the 3-way merge keeps main's version — no revert.

Fixed regardless: I rebased the branch onto current `kcap-cli` origin/main (now 0 behind). The bracketed-paste path in `AgentOrchestrator.cs` and `AgentOrchestratorBracketedPasteTests.cs` are present and untouched; the PR's only changes remain the 8 Cursor files. CLI suite green, AOT clean.

PR #243, opened 2026-07-03
Record silent daemon deaths: unclean-exit breadcrumb + stderr capture

realtonyyoung:
## Problem

The `kcap` agent daemon can vanish with **no trace in `daemon.log`** — no graceful-shutdown lines, no signal log, no exception (AI-1155). This recurred many times/day on the reporter's machine and broke in-progress `kcap mcp flows` reviews (the hosted reviewer died mid-round).

## Root cause

An **uncatchable `SIGKILL`** — almost always the macOS OOM killer (**jetsam**) under memory pressure; also `kill -9`, power loss, etc. Ruling out the alternatives:

- The daemon already logs every **catchable** signal (SIGINT/TERM/HUP/QUIT) + managed crashes (shipped since v0.6.10). None appeared ⟹ not a catchable signal / managed crash.
- macOS writes `.ips` crash reports for native aborts — **zero** for `kcap-daemon` ⟹ not a native crash.
- Silent + not-catchable + not-a-crash ⟹ `SIGKILL`.

Two gaps make it un-diagnosable: (1) `SIGKILL` can't be caught in-process; (2) the detached path (`kcap daemon start -d`) redirects the daemon's std streams to a pipe then immediately closes it (anti-hang, AI-839), so a runtime `"Fatal error."`/native abort written to fd 2 hits a broken pipe and is lost — `daemon.log` only carries ILogger output.

## Changes

- **Unclean-exit breadcrumb** — `DaemonLock` reports `PriorExitWasUnclean`/`PriorHolderPid`: a PID file still present once we hold the exclusive flock means the prior holder skipped `Dispose` (= uncatchable kill). `DaemonRunner` logs a startup warning naming the dead PID, suppressed for clean `--await-lock` update handoffs.
- **stderr capture** — new `StdErrCapture`: `--stderr-file <path>` `dup2`s fds 1/2 onto a file so native/runtime fatal output survives. `StartDetached` passes `daemon.out.log`; no-op under launchd (already captures via `StandardErrorPath`) and foreground.
- **Docs** — README "Diagnosing a hard death".

The real recovery mechanism already exists (`kcap daemon service install`, launchd `KeepAlive`); this makes a recurrence **diagnosable** and leaves a breadcrumb.

## Testing

- 4 new `DaemonLockTests` (breadcrumb) + 2 new `StdErrCaptureTests` (redirect gate). Full unit suite: **2075 passed**.
- E2E with the AOT binary: breadcrumb fires only after `kill -9` (correct dead PID); `lsof` confirms fds 1&2 → `daemon.out.log` with `--stderr-file` and no self-redirect without it.
- CLI + daemon Release builds clean; no IL/AOT warnings.

Linear: AI-1155

_Note: no companion kurrent-io/kcap-cli GitHub issue was filed for this; link one if the repo convention requires it._

### DIFF
--- a/src/Capacitor.Cli.Daemon/DaemonLock.cs
+++ b/src/Capacitor.Cli.Daemon/DaemonLock.cs
@@ -147,10 +147,12 @@ internal sealed class DaemonLock : IDisposable {
         // present-but-unreadable file (permissions, transient IO, corruption)
         // still means the prior holder skipped cleanup — report unclean with an
         // unknown PID rather than silently masking a real hard-death as clean.
+        // Read only the first line (the PID; the daemon writes "{pid}\n{token}")
+        // instead of slurping the whole file, so a corrupt/oversized file can't
+        // force a large allocation on every startup.
         try {
-            var first = File.ReadAllText(pidPath)
-                .Split('\n', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
-                .FirstOrDefault();
+            using var reader = File.OpenText(pidPath);
+            var first = reader.ReadLine();
 
             return (true, int.TryParse(first, out var pid) ? pid : null);
         } catch {
--- a/src/Capacitor.Cli.Daemon/DaemonLock.cs
+++ b/src/Capacitor.Cli.Daemon/DaemonLock.cs
@@ -133,17 +133,28 @@ internal sealed class DaemonLock : IDisposable {
     /// case we still report unclean but with a null PID.
     /// </summary>
     static (bool unclean, int? pid) InspectPriorHolder(string pidPath) {
+        bool present;
         try {
-            if (!File.Exists(pidPath)) return (false, null);
+            present = File.Exists(pidPath);
+        } catch {
+            // Can't even stat it — don't fabricate a breadcrumb.
+            return (false, null);
+        }
 
+        if (!present) return (false, null);
+
+        // Presence is the unclean-exit signal; parsing the PID is secondary. A
+        // present-but-unreadable file (permissions, transient IO, corruption)
+        // still means the prior holder skipped cleanup — report unclean with an
+        // unknown PID rather than silently masking a real hard-death as clean.
+        try {
             var first = File.ReadAllText(pidPath)
                 .Split('\n', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
                 .FirstOrDefault();
 
             return (true, int.TryParse(first, out var pid) ? pid : null);
         } catch {
-            // Can't read it — don't fabricate a breadcrumb.
-            return (false, null);
+            return (true, null);
         }
     }
 
@@ -177,27 +188,17 @@ internal sealed class DaemonLock : IDisposable {
         if (_disposed) return;
         _disposed = true;
 
-        // Releasing the FileStream releases the kernel-level flock. After
-        // that, a follow-up start with the same name can acquire cleanly.
-        try { _stream.Dispose(); } catch { /* best-effort */ }
-
-        // Do NOT delete the lock file. The kernel flock is what enforces
-        // exclusion; file presence on disk is irrelevant. Deleting it here
-        // races against another daemon that may already have acquired the
-        // path between our Dispose() and the unlink: our `File.Delete`
-        // would unlink the inode they're holding open, and a third daemon
-        // could then create a brand-new `<name>.lock` at the same path
-        // and acquire a SECOND independent flock — defeating the whole
-        // AI-630 guard. `kcap daemon doctor --clean` removes truly
-        // stale files.
-        //
-        // Delete the PID file only if it still points to our own PID. A
-        // legitimate successor daemon that ran while we were disposing
-        // would have already rewritten the file to its own PID, and an
-        // unconditional Delete here would orphan their entry. The version
-        // marker follows the SAME ownership guard: if a successor has taken
-        // over the name it has already written its own version marker, so
-        // deleting here would clobber the successor's fresh value.
+        // AI-1155: delete the PID file (and version marker) BEFORE releasing the
+        // flock, not after. A successor waiting on the lock (`--await-lock`
+        // restart-after-update) can't acquire it until we release below, so by
+        // then our PID file is already gone — it can never observe a leftover PID
+        // file from our *clean* shutdown and misread it as an unclean death.
+        // That makes "PID file present once you hold the flock" an unambiguous
+        // uncatchable-kill signal (see TryAcquire), so the successor's startup
+        // breadcrumb no longer needs to special-case the handoff. Because we
+        // still hold the flock here, no other daemon can have rewritten these
+        // files, so the ownership check below always passes for us — it stays as
+        // defence-in-depth (e.g. a stale PID file we never owned).
         bool ours;
         try { ours = PidFileMatchesCurrentProcess(_pidPath); } catch { ours = false; }
 
@@ -205,6 +206,13 @@ internal sealed class DaemonLock : IDisposable {
             try { File.Delete(_pidPath); } catch { /* best-effort */ }
             try { File.Delete(_versionPath); } catch { /* best-effort */ }
         }
+
+        // Release the kernel-level flock last. Do NOT delete the lock file: the
+        // kernel flock is what enforces exclusion, file presence on disk is
+        // irrelevant, and unlinking it here would race a daemon that acquired the
+        // path between our unlink and a re-create — reopening the AI-630 hole.
+        // `kcap daemon doctor --clean` removes truly stale files.
+        try { _stream.Dispose(); } catch { /* best-effort */ }
     }
 
     static bool PidFileMatchesCurrentProcess(string pidPath) {
--- a/src/Capacitor.Cli.Daemon/DaemonLock.cs
+++ b/src/Capacitor.Cli.Daemon/DaemonLock.cs
@@ -31,12 +31,33 @@ internal sealed class DaemonLock : IDisposable {
 
     public string InstanceId { get; }
 
-    DaemonLock(FileStream stream, string lockPath, string pidPath, string versionPath, string instanceId) {
-        _stream      = stream;
-        _lockPath    = lockPath;
-        _pidPath     = pidPath;
-        _versionPath = versionPath;
-        InstanceId   = instanceId;
+    /// <summary>
+    /// True when a PID file for this name was still on disk at the moment we
+    /// acquired the exclusive flock (AI-1155). A graceful shutdown deletes the
+    /// PID file in <see cref="Dispose"/>, so a leftover one means the previous
+    /// holder exited WITHOUT running its cleanup — the signature of an
+    /// uncatchable termination (external <c>SIGKILL</c> from macOS jetsam/OOM
+    /// or <c>kill -9</c>, a power loss, or a hard native crash), none of which
+    /// a signal handler inside the dying process can log. The successor logs a
+    /// startup breadcrumb so the otherwise-silent death leaves a trace.
+    /// </summary>
+    public bool PriorExitWasUnclean { get; }
+
+    /// <summary>
+    /// The PID recorded in the leftover PID file, when <see cref="PriorExitWasUnclean"/>
+    /// is true and the file's first line parsed as an integer; otherwise null.
+    /// </summary>
+    public int? PriorHolderPid { get; }
+
+    DaemonLock(FileStream stream, string lockPath, string pidPath, string versionPath, string instanceId,
+               bool priorExitWasUnclean, int? priorHolderPid) {
+        _stream             = stream;
+        _lockPath           = lockPath;
+        _pidPath            = pidPath;
+        _versionPath        = versionPath;
+        InstanceId          = instanceId;
+        PriorExitWasUnclean = priorExitWasUnclean;
+        PriorHolderPid      = priorHolderPid;
     }
 
     /// <summary>
@@ -70,6 +91,13 @@ internal sealed class DaemonLock : IDisposable {
 
         var instanceId = Guid.NewGuid().ToString("N");
 
+        // AI-1155: inspect any leftover PID file BEFORE WritePidFile overwrites
+        // it. Now that we hold the exclusive flock the prior holder is provably
+        // gone, so a still-present PID file means it never ran Dispose (which
+        // deletes it) — i.e. it died via an uncatchable SIGKILL / crash. Capture
+        // that for the successor's startup breadcrumb.
+        var (priorUnclean, priorPid) = InspectPriorHolder(pidPath);
+
         try {
             // Rewrite the file content with the fresh instance id. Truncate
             // first so a smaller new id doesn't leave trailing bytes from
@@ -94,7 +122,29 @@ internal sealed class DaemonLock : IDisposable {
             try { DaemonVersionMarker.Write(daemonName, version); } catch { /* best-effort */ }
         }
 
-        return new DaemonLock(stream, lockPath, pidPath, versionPath, instanceId);
+        return new DaemonLock(stream, lockPath, pidPath, versionPath, instanceId, priorUnclean, priorPid);
+    }
+
+    /// <summary>
+    /// Reads a leftover PID file (call only while holding the flock, before
+    /// overwriting it). Returns (unclean: whether the file was present at all,
+    /// pid: its parsed first line if any). A present file ⇒ the prior holder
+    /// skipped its cleanup ⇒ unclean exit; the PID may be unparseable, in which
+    /// case we still report unclean but with a null PID.
+    /// </summary>
+    static (bool unclean, int? pid) InspectPriorHolder(string pidPath) {
+        try {
+            if (!File.Exists(pidPath)) return (false, null);
+
+            var first = File.ReadAllText(pidPath)
+                .Split('\n', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
+                .FirstOrDefault();
+
+            return (true, int.TryParse(first, out var pid) ? pid : null);
+        } catch {
+            // Can't read it — don't fabricate a breadcrumb.
+            return (false, null);
+        }
     }
 
     /// <summary>
--- a/src/Capacitor.Cli.Daemon/DaemonRunner.cs
+++ b/src/Capacitor.Cli.Daemon/DaemonRunner.cs
@@ -26,6 +26,7 @@ public static partial class DaemonRunner {
 
     public static async Task<int> RunAsync(string[] args) {
         string?    logFile     = null;
+        string?    stderrFile  = null;
         LogLevel?  logLevelArg = null;
         var        config      = new DaemonConfig();
 
@@ -47,6 +48,7 @@ public static partial class DaemonRunner {
         for (var i = 0; i < args.Length - 1; i++) {
             switch (args[i]) {
                 case "--log-file": logFile = args[++i]; break;
+                case "--stderr-file": stderrFile = args[++i]; break;
                 case "--log-level": logLevelArg = ParseLogLevel(args[++i]); break;
                 case "--max-agents" when int.TryParse(args[i + 1], out var n) && n >= 1:
                     config.MaxConcurrentAgents = n;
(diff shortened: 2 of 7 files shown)


src/Capacitor.Cli.Daemon/DaemonLock.cs
```diff
@@ -70,6 +91,13 @@ internal sealed class DaemonLock : IDisposable {
 
         var instanceId = Guid.NewGuid().ToString("N");
 
+        // AI-1155: inspect any leftover PID file BEFORE WritePidFile overwrites
+        // it. Now that we hold the exclusive flock the prior holder is provably
+        // gone, so a still-present PID file means it never ran Dispose (which
+        // deletes it) — i.e. it died via an uncatchable SIGKILL / crash. Capture
+        // that for the successor's startup breadcrumb.
+        var (priorUnclean, priorPid) = InspectPriorHolder(pidPath);
```
realtonyyoung, line 99: The breadcrumb invariant is still not quite sound while the CLI supervisor can also write this PID file outside the daemon flock. In the detached start path, `StartDetached` waits 1.5s and then writes `<name>.pid`; if the child exits cleanly after `WaitForExit(1500)` returns false but before that supervisor write, `DaemonLock.Dispose` has already removed its PID while holding the flock, and the supervisor can recreate a stale PID after cleanup. The next daemon then acquires the lock and reports `PriorExitWasUnclean = true` even though the prior holder did run `Dispose`. To make PID-file presence under the flock an unambiguous skipped-cleanup signal, the detached supervisor write needs to stop recreating the daemon-owned PID file after startup, or otherwise be made unable to leave a PID after a clean child exit.

src/Capacitor.Cli.Daemon/StdErrCapture.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    /// <summary>
+    /// Point fds 1 and 2 at <paramref name="path"/> (append mode). Best-effort;
+    /// returns true only if both redirect syscalls succeeded. Call once, as
+    /// early as possible in startup, so even a crash during host construction
+    /// is captured.
+    /// </summary>
+    public static bool Apply(string path) {
+        try {
+            var fs = new FileStream(path, FileMode.Append, FileAccess.Write, FileShare.ReadWrite);
```
realtonyyoung, line 60: `Apply` runs before `RollingFileLoggerProvider` creates the config directory, so on a fresh install/profile where `~/.config/kcap` does not exist yet this open fails, the catch returns false, and stderr capture stays disabled for the daemon lifetime. The daemon still starts, but a later native/runtime abort on that first detached run is lost. Please create the parent directory inside this best-effort try before opening the file, so the early redirect does not depend on the later logging provider setup.

src/Capacitor.Cli.Daemon/DaemonLock.cs
realtonyyoung: Fixed in eaa6f02. You're right — the redundant CLI supervisor `WritePidFile(name, process)` was a second, out-of-flock writer that could recreate a stale PID after the daemon's clean `Dispose`. Removed both supervisor writes (foreground + detached) and the now-unused method, so the daemon is the sole PID-file owner (write under the flock in TryAcquire, delete under the flock in Dispose). "PID file present once the successor holds the flock ⇒ skipped cleanup" is now sound. The daemon writes its PID early in startup (before the 1.5s readiness wait completes), so status/stop still find it; the foreground `finally` cleanup stays (guarded by pid match).

src/Capacitor.Cli.Daemon/StdErrCapture.cs
realtonyyoung: Fixed in eaa6f02. `StdErrCapture.Apply` now does `Directory.CreateDirectory(Path.GetDirectoryName(path))` inside the best-effort try before opening, so a fresh install / custom KCAP_CONFIG_DIR (dir not yet created by the logging provider) no longer disables capture. Verified E2E: pointing --stderr-file at a nonexistent nested dir now creates it and `lsof` shows fd 2 redirected onto the file.

src/Capacitor.Cli.Daemon/StdErrCapture.cs
```diff
@@ -0,0 +1,65 @@
...
+            _capture = fs; // keep alive for the process lifetime
+            var fd = (int)fs.SafeFileHandle.DangerousGetHandle();
+
+            // stdout then stderr onto the same open file description.
+            return dup2(fd, 1) != -1 && dup2(fd, 2) != -1;
+        } catch {
+            // Directory missing, permission denied, redirected to a closed pipe,
+            // etc. Already best-effort — never let this abort daemon startup.
+            return false;
+        }
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in eaa6f02 — `StdErrCapture.Apply` now creates the parent directory before opening the capture file (still best-effort; never aborts startup). Verified E2E with a nonexistent parent dir.

src/Capacitor.Cli.Daemon/DaemonLock.cs
```diff
@@ -94,7 +122,29 @@ internal sealed class DaemonLock : IDisposable {
...
+
+            var first = File.ReadAllText(pidPath)
+                .Split('\n', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
+                .FirstOrDefault();
+
+            return (true, int.TryParse(first, out var pid) ? pid : null);
+        } catch {
+            // Can't read it — don't fabricate a breadcrumb.
+            return (false, null);
+        }
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in eaa6f02 — `InspectPriorHolder` now reads only the first line via `File.OpenText`/`ReadLine` instead of `ReadAllText().Split()`, avoiding a large allocation on a corrupt/oversized file. Behavior unchanged (present⇒unclean; PID parsed if possible).

src/Capacitor.Cli.Daemon/DaemonRunner.cs
```diff
@@ -47,6 +48,7 @@ public static async Task<int> RunAsync(string[] args) {
         for (var i = 0; i < args.Length - 1; i++) {
             switch (args[i]) {
                 case "--log-file": logFile = args[++i]; break;
+                case "--stderr-file": stderrFile = args[++i]; break;
                 case "--log-level": logLevelArg = ParseLogLevel(args[++i]); break;
                 case "--max-agents" when int.TryParse(args[i + 1], out var n) && n >= 1:
                     config.MaxConcurrentAgents = n;
```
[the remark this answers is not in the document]
realtonyyoung: Addressed in eaa6f02 — documented in README under "Diagnosing a hard death", noting `--stderr-file` is set automatically by `kcap daemon start -d` (internal wiring, not a user-facing knob).

src/Capacitor.Cli/Commands/DaemonCommands.cs
```diff
@@ -121,7 +121,11 @@ static async Task<int> SpawnForegroundAsync(string name, string[] args) {
             return 1;
         }
 
-        WritePidFile(name, process);
+        // The daemon writes its own PID file inside the flock during startup
```
realtonyyoung: This foreground path still is not sole-owner: the `finally` just below deletes `DaemonLockPaths.PidPath(name)` if it still matches the child PID. After removing the supervisor writes, the daemon-owned PID file is now the unclean-exit breadcrumb; but if a foreground daemon is killed with SIGKILL or aborts natively while `kcap daemon start` is waiting, `WaitForExitAsync` returns and that `finally` deletes the matching PID before any successor can inspect it. That erases exactly the breadcrumb this PR is trying to preserve for hard deaths. Since clean exits already run `DaemonLock.Dispose` and delete the PID under the flock, the foreground supervisor should not delete this file on child exit except for an intentionally explicit stop/cleanup path.

realtonyyoung: Fixed in be779ed. Removed the foreground `finally` PID-file deletion entirely. You're right that it erased the breadcrumb: on a clean exit the daemon already deletes the PID under the flock (Dispose), so the finally was a no-op; on a hard death it's the only trace, and deleting it defeated the purpose. A leftover stale PID is harmless — the kernel releases the flock, the next `TryAcquire` logs the breadcrumb and overwrites it, and status/doctor already treat dead-PID/foreign-token files as stale. Daemon is now sole writer AND deleter in both modes. Verified E2E: foreground daemon SIGKILLed → CLI exits without deleting the PID → restart logs the breadcrumb.

src/Capacitor.Cli/Commands/DaemonCommands.cs
```diff
@@ -121,21 +121,22 @@ static async Task<int> SpawnForegroundAsync(string name, string[] args) {
...
+        // the flock in Dispose. The supervisor neither writes nor deletes it.
+        //
+        // We deliberately do NOT delete the PID file when the child exits here. On
+        // a clean exit the daemon already removed it via Dispose, so there's
+        // nothing to clean up; on a hard death (SIGKILL / native abort — Dispose
+        // never runs) the leftover PID file IS the unclean-exit breadcrumb the
+        // next start reads, so deleting it here would erase exactly what this
+        // change exists to preserve. A stale PID from a hard death is harmless:
+        // the kernel released the flock, the next TryAcquire overwrites it (after
+        // logging the breadcrumb), and status/doctor already treat a
```
realtonyyoung, line 135: `status` currently does more than treat the stale PID as stale: in the stale branch below it deletes `DaemonLockPaths.PidPath(name)` and the version marker. That gives the new breadcrumb a false negative: after a SIGKILL/native abort, running `kcap daemon status --name X` before the next start will classify the dead PID as stale and remove the file, so the successor later acquires the flock with no PID file present and cannot log `PriorExitWasUnclean`. With PID presence now being the durable hard-death signal, passive status/diagnostic reads should not remove it; keep cleanup to explicit stop/doctor --clean-style paths.

realtonyyoung: Fixed in 41ada21. `status` no longer deletes the stale PID file / version marker — it just reports "not running (stale PID file)". You're right that a passive read erasing the durable hard-death signal is a false negative for the breadcrumb. Stale-entry cleanup stays with the explicit paths (`daemon stop` and `daemon doctor --clean`), both unchanged. Verified E2E: `status` on a stale PID leaves the file in place, while `daemon stop` on a stale PID still removes it.

src/Capacitor.Cli/Commands/DaemonCommands.cs
```diff
@@ -452,13 +465,14 @@ static async Task<int> Status(string[] args) {
                 if (DaemonRestartMarker.TryRead(name) is { } marker)
                     await Console.Out.WriteLineAsync($"  {marker.Describe()}");
             } else {
+                // AI-1155: report the stale PID file but do NOT delete it. Its
```
realtonyyoung, line 524: This fixes the stale/foreign-token branch, but `ReadPidFile` is still a passive deleter for unparseable PID files. The daemon writes this file with `File.WriteAllText` while holding the flock; a SIGKILL/native abort during that truncate/write window can leave a present but empty/partial/unparseable file. `InspectPriorHolder` would correctly report that as `(true, null)`, but `status` and both pre-spawn guards call `ReadPidFile` first, so its unparseable-file delete can remove the breadcrumb before the successor daemon sees it, producing a false-negative clean prior exit. Since PID-file presence is now the durable hard-death signal, unparseable should be reported as no usable PID without unlinking, with cleanup left to explicit `stop` / `doctor --clean` paths.

realtonyyoung: Fixed in 0750c8f. `ReadPidFile` no longer unlinks an unparseable PID file — it returns null (no usable PID) and leaves the file in place. You're right about the truncate/write window: since the daemon writes with `File.WriteAllText` under the flock, a mid-write SIGKILL leaves a present-but-partial file that `InspectPriorHolder` correctly reads as `(unclean, null)`, and `status`/the pre-spawn guards must not delete it out from under the successor. Cleanup of corrupt/stale files stays with the explicit `daemon stop` / `doctor --clean` paths (the next `TryAcquire` also overwrites it after logging the breadcrumb). Verified E2E: empty, partial, and garbage PID files all survive a `status` read.

src/Capacitor.Cli/Commands/DaemonCommands.cs
```diff
@@ -637,9 +635,15 @@ static void WritePidFile(string daemonName, Process process) {
...
+            // AI-1155: report "no usable PID" but do NOT delete the file. The
+            // daemon writes it with File.WriteAllText (truncate+write) under the
+            // flock, so a SIGKILL/native abort mid-write can leave a present but
+            // empty/partial/unparseable file — which is itself a hard-death
+            // breadcrumb that DaemonLock.InspectPriorHolder reports as
+            // (unclean, null). Since callers here (status, the pre-spawn guards)
+            // read this before the successor daemon runs, unlinking a corrupt
+            // file would erase that breadcrumb. Cleanup of corrupt/stale files is
+            // left to the explicit paths (`daemon stop`, `daemon doctor --clean`).
             return null;
```
realtonyyoung, line 705: `ReadPidFile` is now side-effect-free, which fixes the passive-read false negative, but this leaves the advertised explicit cleanup path incomplete for the corrupt/empty PID case. `StopByName` still starts with `if (ReadPidFile(name) is not { } entry) return 1;`, so an empty/partial/unparseable `<name>.pid` from a mid-write SIGKILL is treated the same as an absent file and `daemon stop --name <name>` exits without deleting it. That contradicts this comment/commit message that corrupt cleanup is left to `daemon stop` or `doctor --clean`; only doctor can remove it now. Please make the stop path distinguish “file absent” from “file present but unusable” and remove the latter as an explicit cleanup, while keeping status/pre-spawn reads non-mutating.

realtonyyoung: Fixed in e13165b. `StopByName` now distinguishes an absent PID file from a present-but-unusable one: it still reports "no daemon running (no PID file found)" (rc 1) when the file is genuinely absent, but removes a present-but-unparseable/empty file as explicit cleanup ("removed unusable PID file"). So corrupt cleanup is once again covered by both `daemon stop` and `doctor --clean`, while `status` and the pre-spawn guards stay non-mutating. Verified E2E on the published binary: stop on a corrupt PID removes it; stop on an absent PID reports not-found; status on a corrupt PID preserves it.

src/Capacitor.Cli/Commands/DaemonCommands.cs
```diff
@@ -305,6 +318,20 @@ static int StopByName(string name) {
         }
 
         if (ReadPidFile(name) is not { } entry) {
+            // ReadPidFile returns null for BOTH an absent file and a present-but-
+            // unparseable one (empty/partial — e.g. a mid-write SIGKILL; it no
+            // longer unlinks the latter, AI-1155). `stop` is an explicit cleanup
+            // path, so distinguish them: remove a corrupt/empty file here (the
+            // user asked to stop this daemon, and it's the leftover hard-death
+            // breadcrumb), but report cleanly when the file is genuinely absent.
+            if (File.Exists(DaemonLockPaths.PidPath(name))) {
```
realtonyyoung: This fallback has a TOCTOU race because the `File.Exists`/`File.Delete` decision is not tied to the `ReadPidFile` result. If `stop --name X` reads `null` because the file was absent or corrupt, a concurrent `daemon start --name X` can acquire the flock and write a fresh, valid daemon-owned PID file before this check runs; this branch will then delete that live daemon’s PID file/version marker and report it as “removed unusable PID file.” That leaves a running daemon with no PID file, so status/stop misreport it until the flock-only start guard catches later attempts. Please make the explicit corrupt cleanup prove the file is still unusable under the relevant coordination (for example by serializing with starts / the daemon lock, or by re-reading and not deleting if a valid entry appeared) before unlinking it.

realtonyyoung: Fixed in e44a925. You're right — the File.Exists/File.Delete decision wasn't tied to any coordination. All three of StopByName's marker deletions (corrupt-cleanup, stale-branch, and post-kill) now go through a new `TryCleanupMarkersUnderLock(name)` that deletes the PID file + version marker **only while holding the exclusive daemon flock**. Holding it proves no daemon is live under the name and blocks a concurrent `daemon start` (its `DaemonLock.TryAcquire` waits on the same flock) until we release — after which it writes a fresh PID, so we can never unlink a live daemon's file. It's the same authoritative liveness signal `doctor` uses, and works for supervisor-started, foreground, AND self-respawned daemons (the start-lock wouldn't cover self-respawn). Post-kill also `WaitForExit`s so the flock is released before we reclaim it. A corrupt PID under a held flock is now reported ("appears to be running… not removing") instead of deleted. Verified E2E: (A) live daemon + corrupted PID → stop does NOT delete it, daemon stays up; (B) corrupt+no-daemon → removed; (C) absent → not-found; (D) live daemon → killed + cleaned.

src/Capacitor.Cli/Commands/DaemonCommands.cs
```diff
@@ -305,38 +318,93 @@ static int StopByName(string name) {
...
+    /// Remove the daemon-owned on-disk markers (PID file + version marker) for
+    /// <paramref name="name"/>, but ONLY while holding the exclusive daemon flock.
+    /// Holding it proves no daemon is live under this name and blocks any
+    /// concurrent <c>daemon start</c> until we release — its
+    /// <c>DaemonLock.TryAcquire</c> waits on this same flock, then writes a fresh
+    /// PID afterward. Every daemon (supervisor-started, foreground, or
+    /// self-respawned) takes this flock before writing its PID, so this is the
+    /// race-free way for the CLI to clean up markers without unlinking a live
+    /// daemon's PID file (AI-1155). The lock file itself is never deleted
+    /// (AI-630); <c>doctor --clean</c> owns that. Returns false if a live daemon
```
realtonyyoung, line 389: This leaves the remaining explicit cleanup path with the same class of race the new helper avoids. `doctor --clean`’s no-holder branch successfully opens the lock, disposes that probe, and only then deletes `<name>.lock`, `<name>.pid`, and the version marker. A concurrent `daemon start` can acquire the flock and write a fresh PID in that gap; doctor can then unlink the live daemon’s PID and, worse, unlink the lock path while the daemon still holds the old inode. A later start can create a new lock file and acquire a second independent flock, reopening the AI-630 duplicate-name hole. I don’t think this should be left as a manual-maintenance exception in this PR: either keep doctor cleanup under the acquired flock and avoid deleting the lock file, or at least hold the probe until all deletes that remain are complete so it cannot race a start.

realtonyyoung: Fixed in 747ec69. `doctor --clean` is now flock-safe and no longer deletes the lock file. The "no holder" branch deletes pid + version + restart-pending **while still holding the probe's flock** (dispose moved to after the deletes) so a concurrent `daemon start` can't slip a live daemon in and have us unlink its fresh PID; and it stops unlinking `<name>.lock` entirely — as you noted, unlinking it (even under the flock) lets a later start create a new inode and take a second independent flock, reopening AI-630. The lock file is inert and reused, matching `DaemonLock.Dispose`'s policy. The "orphan, no lock" branch routes through the same `TryCleanupMarkersUnderLock` (acquires the flock first, skips if a start won the race). Verified E2E (isolated $HOME): a stale entry keeps its lock file but loses pid+version; a live daemon is classified HELD and left fully intact.

On the other item I flagged (atomic `WritePidFile`): I'm going to **not** do it — it would hurt the breadcrumb. A mid-write SIGKILL currently leaves a *partial* PID file that `InspectPriorHolder` reads as `(unclean, null)`, so the breadcrumb fires; an atomic temp+rename would leave no PID file (only a `.tmp`), so that hard death would go **unrecorded**. The partial file is effectively a feature, and it's now handled gracefully by every reader (status/guards/ReadPidFile non-mutating; stop/doctor flock-guarded). Happy to revisit if you disagree.

realtonyyoung:
Re-reviewed latest commit 747ec69 and did the whole marker lifecycle pass. No substantive findings remain: PID writes are daemon-only under the flock, pid/version/restart cleanup is now flock-held in daemon Dispose, stop, and doctor --clean, passive reads stay non-mutating, and no path deletes the .lock file. The non-atomic PID write is the right tradeoff for this breadcrumb: a mid-write hard death can leave a partial present file that still records the unclean exit, and the PR now handles that case consistently.

PR #246, opened 2026-07-03
Mirror the requester's working tree into review-flow reviewer worktrees

### LINKED ISSUE #245: Review-flow reviewer doesn't read the requester's in-progress/uncommitted source
## Summary

Hosted review-flow reviewers don't read the repository source — they review the submitted context (e.g. a spec) in isolation, even though the daemon runs on the same machine as the requester. The reviewer's daemon-created worktree is only refreshed from the requester's working tree on **follow-up** rounds, and only when the requester's repo root **exactly** equals the daemon's checkout path. So the initial round always runs against committed HEAD, and a review launched from a git worktree never mirrors the requester's changes at all.

## Daemon-side change (this repo)

- Add optional `LaunchAgentCommand.SyncFromRepoRoot` (appended field, wire-compatible over SignalR).
- On launch, after creating the reviewer worktree and **before spawning**, mirror the requester's working tree (`SyncFromSourceAsync` — tracked + untracked, gitignore-respected) so round 1 sees uncommitted/in-progress code.
- **Daemon-validated source**: only sync when the source resolves on this host and is a checkout of the same repo (`origin` match) — a different machine or repo is skipped, leaving the worktree at HEAD. Apply the same validation to the existing follow-up `RefreshAgentWorktree` handler so relaxed (non-exact-path) syncs are safe.
- Update the `kcap mcp flows` `mode` tool description, the review-flows skill, and README to reflect that the reviewer now sees the working tree by default (with `mode=context-only` as the opt-out).

Server-side counterpart (flow definitions, orchestrator, wire field): kurrent-io/kcap-server#927.

Tracked in Linear AI-1163.

alexeyzimarev:
## Problem

Hosted review-flow reviewers don't read the repository source — they review the submitted context (e.g. a spec) in isolation, even on the same machine as the requester. The reviewer's daemon-created worktree was only refreshed from the requester's working tree on **follow-up** rounds, and only when the requester's repo root **exactly** equalled the daemon's checkout path. So round 1 always ran against committed HEAD, and a review launched from a git worktree never mirrored the requester's changes.

## Change (daemon side)

- Add optional `LaunchAgentCommand.SyncFromRepoRoot` (appended field; SignalR positional binding stays wire-compatible — old daemon ignores it, old server → new daemon defaults it to null; covered by `LaunchAgentCommandWireFormatTests`).
- On launch, after creating the reviewer worktree and **before spawning**, mirror the requester's working tree (`SyncFromSourceAsync` — tracked + untracked, gitignore-respected) so round 1 grounds the review in in-progress/uncommitted code.
- **Daemon-validated source**: only sync when the source resolves on this host and is a checkout of the same repo (`origin` match). The same validation is added to the existing follow-up `RefreshAgentWorktree` handler, so the server can now relax its exact-path guard and let a worktree launch sync too; a different machine/repo is safely skipped (reviewer keeps HEAD).
- Docs: the `kcap mcp flows` `mode` tool description, the review-flows skill, and README now say the reviewer sees the working tree by default, with `mode=context-only` as the opt-out.

## Cross-repo

Server counterpart (flow definitions → `mirror-requester`, orchestrator, matching wire field, guard rework): kurrent-io/kcap-server#927. Deploy the server and CLI together; the optional wire field keeps either-order rollout safe.

## Tests

- `LaunchAgentCommandWireFormatTests`: new field round-trips over the server wire options and defaults to null on legacy (absent) JSON.
- `WorktreeManagerTests` already covers `SyncFromSourceAsync`.
- Full `Capacitor.Cli.Tests.Unit` suite green (2092); no IL2026/IL3050 AOT warnings on `dotnet publish -c Release` for both `kcap` and `kcap-daemon`.

Closes #245
Linear: AI-1163

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ args    = ["mcp", "flows"] # desktop app, which launches MCP servers without you
 
 It provides four tools:
 
-- **`start_review_flow`** — start a new review flow (`spec-review` or `code-review`). Provide `kind`, `target_kind`, `target_ref`, `target_title`, and `context`. Requester context (session ID, cwd, repo root, owner, name) is resolved automatically from the environment. Returns a `flow_run_id`.
+- **`start_review_flow`** — start a new review flow (`spec-review` or `code-review`). Provide `kind`, `target_kind`, `target_ref`, `target_title`, and `context`. Requester context (session ID, cwd, repo root, owner, name) is resolved automatically from the environment. Returns a `flow_run_id`. Optional `mode`: by default, when the daemon runs on the same machine, the reviewer works in a worktree mirrored from your working tree — **uncommitted and untracked changes included** — so it grounds the review in your actual in-progress source; pass `mode="context-only"` to opt out and have the reviewer treat only the submitted context as authoritative.
 - **`submit_review_round`** — submit a follow-up round to an existing flow with updated context or a response to the reviewer's findings. Returns the new round's findings.
 - **`get_review_flow_status`** — get the current status (running, waiting, completed, failed) and last result of a flow.
 - **`close_review_flow`** — mark a completed review flow as closed.
--- a/kcap/skills/review-flows/SKILL.md
+++ b/kcap/skills/review-flows/SKILL.md
@@ -68,7 +68,7 @@ if FINDINGS:
 
 | Tool | Required args | Optional args | When to call |
 |---|---|---|---|
-| `start_review_flow` | `kind` (`spec-review`\|`code-review`), `target_kind` (what is being reviewed: `spec`, `code`, `pr`, `branch`, `file`, etc.), `target_ref` (a path, branch name, or PR URL/number that identifies the target), `target_title` (short human-readable title, e.g. spec name or PR title), `context` (background context: what to focus on, constraints, definition of done) | `instructions`, `mode` (`context-only` — required for code-review unless the reviewer runs in your exact repo checkout) | Once, at the start of a review task. |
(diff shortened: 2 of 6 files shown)


alexeyzimarev:
### Response to Qodo review

**1. Sync bypasses repo allowlist — fixed** (df6058b).
Good catch. The launch repo was allowlist-checked but the sync *source* wasn't. Folded `_config.IsRepoAllowed(sourceRepoRoot)` into the shared validator (now `IsAllowedSyncSourceAsync`), so both launch-time mirroring and the follow-up `RefreshAgentWorktree` enforce the same repo-path policy as the launch repo. No-op when no allowlist is configured; only tightens behavior for operators who set one.

**2. Sync blocked without origin — declining (with rationale).**
Technically true that `SourceMatchesRepoOriginAsync`/`GetOriginRemoteAsync` fail closed when `origin` is unset, but that state can't occur for a review flow: flows are only ever discovered and launched for repos with a matching GitHub origin — `DiscoverDaemonsAsync` resolves the daemon by owner/repo *from* the origin remote, and the server requires `repo_owner`/`repo_name`. So both the source and the daemon checkout always have an `origin` here; a no-origin (local-only) repo can't be a flow target. Git worktrees inherit the clone's `origin`, so the worktree-requester case (the whole point of this PR) matches too. Adding a `git rev-parse --git-common-dir` fallback would be extra code + a git call for an unreachable case, so I've instead documented why origin-match is the intended identity check on `IsAllowedSyncSourceAsync`.

(The earlier ubuntu CI failure was an unrelated flaky real-socket test, `Local_socket_list_round_trips_...` — green on re-run.)

PR #256, opened 2026-07-03
[AI-1158] Antigravity ingest (Phase 2, CLI): plugin, hook dispatch, watcher, cost backfill

realtonyyoung:
Phase 2 (CLI) of Google Antigravity ingest — the 9th coding-agent vendor. Server side is Phase 1 (kcap-server#933, AI-1157). This makes `kcap` install/capture/report Antigravity like the other eight.

Linear: AI-1158 (parent AI-1150). Depends on the server Phase 1 (AI-1157) for the ingest endpoints + normalizer.

## What's here (8 TDD commits)
- **Filesystem layout** (`AntigravityPaths`) — Antigravity shares `~/.gemini` with the Gemini CLI, in an `antigravity` subdir (honours `GEMINI_CLI_HOME`).
- **Gemini-collision fix** — `GeminiPaths.IsInstalled` now requires a Gemini-specific marker so an Antigravity-only `~/.gemini` doesn't read as Gemini.
- **Plugin installer** (`AntigravityHooks` + `AntigravityHooksInstaller`) — a `kcap` block in `hooks.json` (two entry shapes: tool events `[{matcher,hooks:[…]}]`, lifecycle events a direct handler list). Merge preserves user blocks; malformed JSON is backed up; version marker.
- **Hook dispatcher** (`kcap hook --antigravity <Event>`) — acts on the first `PreInvocation` (session-start + watcher tailing `transcript_full.jsonl`); session-end is watcher-owned via idle (GUI process outlives a conversation, like the Codex desktop; `KCAP_ANTIGRAVITY_IDLE_MINUTES`).
- **Watcher extractors** — USER_INPUT (strip `<USER_REQUEST>`) / PLANNER_RESPONSE title text; `IsEvent` gate; `ShouldEndOnIdle` generalized to the antigravity vendor.
- **Cost backfill** — decode each conversation's SQLite `gen_metadata` protobuf (field numbers pinned empirically + **verified against real on-disk blobs**) → synthetic `USAGE` lines → server `AntigravityUsageBackfilledEvent` (priced on read; cost never stored).
- **Vendor registration** — `plugin`, `setup` (+ `--skip-antigravity-hooks`), `status`, `uninstall`, `watch --vendor antigravity`.
- **Docs** — README section + help text.

## Scope notes (deliberate)
- **Subagents are captured as standalone sessions** (they fire their own lifecycle hooks). *Nesting* them under the parent is deferred: the parent's `INVOKE_SUBAGENT` transcript step carries child conversation ids at spawn, but children also fire their own hooks, so de-duping requires the annotations-absence signal — and mis-timing it would drop top-level sessions. Needs live multi-agent E2E; tracked as a follow-up.
- **Historical import** (`kcap import --antigravity`) is Phase 4 (AI-1160) — intentionally not wired into `VendorSelection`.
- **cache-write / reasoning tokens** are omitted from cost (no write count in implicit caching; reasoning field not yet confidently identified). Server folds reasoning into output, so omission under-counts, never mis-bills.

## Testing
Full CLI unit suite green (one unrelated import-timing flake passes in isolation). AOT publish clean (no IL2026/IL3050). gen_metadata decoder verified against real Antigravity 2.2.1 blobs. Live end-to-end against a running Antigravity IDE is the remaining manual verification (Phase 3, AI-1159).


src/Capacitor.Cli/Commands/WatchCommand.cs
```diff
@@ -1180,6 +1211,68 @@ internal static bool IsEvent(string line, string vendor = "claude") {
...
+
+    static void AppendAntigravityUsageLines(WatchState state, List<string> newLines, List<int> newLineNumbers, string conversationId) {
+        try {
+            var dbPath = AntigravityPaths.ConversationDb(conversationId);
+            var rows   = AntigravityGenMetadataDb.ReadUsageLines(dbPath, state.LastAntigravityGenIdx);
+
+            foreach (var (idx, line) in rows) {
+                newLines.Add(line);
+                newLineNumbers.Add((int)(AntigravityUsageLineBase + idx));
+                if (idx > state.LastAntigravityGenIdx) state.LastAntigravityGenIdx = idx;
```
realtonyyoung: This advances `LastAntigravityGenIdx` before the synthetic `USAGE` lines are successfully sent. If `SendTranscriptBatch2` fails, `LinesProcessed` stays behind so real transcript lines retry, but these usage rows are skipped forever on the next poll because the high-water mark already moved. Keep the decoded rows pending and only advance the watermark after the batch send succeeds, or update it from the successfully-sent usage rows alongside `LinesProcessed`.

src/Capacitor.Cli/Program.cs
```diff
@@ -563,7 +563,7 @@
             storedOrg:               storedOrg);
     }
     case "watch" when args.Length < 3:
-        Console.Error.WriteLine("Usage: kcap watch <sessionId> <transcriptPath> [--agent-id <agentId>] [--cwd <cwd>] [--skip-title] [--parent-pid <pid>] [--vendor claude|codex|copilot|gemini|kiro|pi|opencode]");
+        Console.Error.WriteLine("Usage: kcap watch <sessionId> <transcriptPath> [--agent-id <agentId>] [--cwd <cwd>] [--skip-title] [--parent-pid <pid>] [--vendor claude|codex|copilot|gemini|kiro|pi|opencode|antigravity]");
```
realtonyyoung, line 566: This wires Antigravity into `watch`, but the shared parser immediately below still normalizes `args[1]` with `Replace("-", "")`. Antigravity's hook forwards the dashed `conversationId` as `session_id`, so the watcher connects, sends transcript/USAGE batches, and posts idle session-end under a different dashless id. The same mismatch makes `kcap disable` ineffective because disable writes a dashless marker while `AntigravityHookCommand` checks the dashed id. Antigravity needs one canonical id end-to-end: either preserve the dashed id through `watch`/disable for this vendor, or normalize the hook payload before both session-start and watcher spawn.

src/Capacitor.Cli/Commands/WatchCommand.cs
```diff
@@ -580,7 +588,7 @@ internal static bool ShouldEndOnIdle(
             TimeSpan       idleTimeout,
             bool           toolInFlight = false
         ) =>
-        vendor == "codex"
+        (vendor == "codex" || vendor == "antigravity")
```
realtonyyoung, line 602: `ShouldEndOnIdle` now applies to Antigravity, but the `toolInFlight` value passed to it is still derived only from Codex `response_item` shapes, so it is always false for Antigravity. A long-running Antigravity command/tool can produce no transcript line for longer than the idle window, causing the watcher to POST session-end while the turn is still active. Please add Antigravity-specific in-flight tracking (for example from RUN_COMMAND/tool status or control-hook events) or avoid applying the tool-suppressed idle policy to Antigravity until it has an equivalent signal.

src/Capacitor.Cli.Core/Antigravity/AntigravityGenMetadata.cs
```diff
@@ -0,0 +1,158 @@
...
+        return false;
+    }
+
+    static long? FindVarint(ReadOnlySpan<byte> buf, int field) {
+        var i = 0;
+        while (i < buf.Length) {
+            if (!TryReadTag(buf, ref i, out var fn, out var wt)) break;
+            if (wt == 0) {
+                if (!TryReadVarint(buf, ref i, out var v)) break;
+                if (fn == field) return (long)v;
```
realtonyyoung: Malformed/schema-drifted blobs can still escape the fail-open path here: an over-large varint casts from `ulong` to `long` unchecked, so values above `long.MaxValue` become negative token counts and are emitted as usage. The length-delimited walker has the same pattern with `(int)len` before checking bounds. Please reject counts larger than `long.MaxValue` and lengths larger than the remaining buffer before casting, so malformed blobs return `null` instead of producing bogus usage.

realtonyyoung:
Reviewed current head 912a9e8 without running tests/builds. Added four inline findings: Antigravity ID normalization/disable mismatch, USAGE watermark loss on send failure, idle ending while an Antigravity tool may still be in flight, and protobuf overflow fail-open gaps. Existing Qodo comments already cover the hook payload GetValue/missing-event fail-open issues.

src/Capacitor.Cli/Commands/WatchCommand.cs
```diff
@@ -580,7 +588,7 @@ internal static bool ShouldEndOnIdle(
...
         && isSessionWatcher
         && thresholdReached
```
realtonyyoung, line 604: `ShouldEndOnIdle` still requires `thresholdReached` for Antigravity, but Antigravity has already posted `/hooks/session-start/antigravity` before the watcher starts. A conversation that produces fewer than 10 buffered lines will never satisfy this predicate, so the idle path never runs; if it later exits through another path, the session-end POST is also skipped because `ThresholdReached` is false. That leaves short Antigravity conversations active on the server and/or leaves their watcher running until the IDE exits. Either defer session-start until the threshold is reached, or let Antigravity idle-end/post session-end even below threshold while deciding separately whether to flush the buffered lines.

realtonyyoung:
Re-reviewed current head 912a9e8, including the new WireMock Antigravity hook-dispatcher test; no tests/builds run. Added one new inline finding for below-threshold Antigravity sessions never idle-ending after session-start. Existing inline findings still cover the ID normalization mismatch, USAGE watermark loss on send failure, Antigravity tool-in-flight idle risk, and protobuf overflow fail-open gap; Qodo already covers the hook payload/missing-event fail-open issues. The new integration test meaningfully covers the session-start body/default visibility and excluded-path skip, and the pre-seeded pid neutralization looks self-cleaning.

realtonyyoung:
Addressed the re-review findings in 16a8505:

- **[blocking] Canonical dashless id (C2):** `kcap watch`/`kcap disable` normalize ids to dashless, but the hook forwarded the dashed `conversationId`, so transcript/USAGE and disable ran under a different id than session-start. The hook now uses the dashless id for `session_id`, the watcher key, and disable; the watcher derives the sibling `gen_metadata` db from the transcript path (which keeps the real conversation id) via `AntigravityPaths.ConversationDbFromTranscript`.
- **[blocking] Fail-open payload reads (Q1):** `JsonNode.GetValue<string>()` threw on a non-string shape; a safe accessor now no-ops instead of failing the control hook.
- **[should-fix] Control hook exits 0 on missing event (Q2).**
- **[should-fix] USAGE gen-watermark (C1):** advances only after the batch send succeeds — a failed send used to skip those cost rows forever; now they re-read next drain. USAGE is only injected outside the below-threshold buffering phase.
- **[should-fix] Short-conversation idle-end (C5):** Antigravity posts session-start before the watcher, so it now streams past-threshold from the start and short (<10-line) conversations still idle-end + post session-end instead of lingering Active.
- **[should-fix] Varint/length guards (C4):** reject a token varint above `long.MaxValue` (no negative counts) and a length past the buffer.

**Deferred (tracked follow-up):** an Antigravity-specific in-flight signal for the idle policy (C3) — the transcript has no such signal, and the 60-min window makes premature idle-end during a long command unlikely.

34 unit + 2 integration antigravity tests green; AOT publish clean.

src/Capacitor.Cli.Core/Antigravity/AntigravityPaths.cs
```diff
@@ -0,0 +1,81 @@
...
+        var sysGen  = Path.GetDirectoryName(logsDir);                        // …/.system_generated
+        var convDir = Path.GetDirectoryName(sysGen);                         // …/<id>
+        var brain   = Path.GetDirectoryName(convDir);                        // …/brain
+        var root    = Path.GetDirectoryName(brain);                          // …/antigravity
+        if (convDir is null || root is null) return null;
+
+        var convId = Path.GetFileName(convDir);
+        if (string.IsNullOrEmpty(convId)) return null;
+
+        return Path.Combine(root, "conversations", $"{convId}.db");
```
realtonyyoung, line 73: `ConversationDbFromTranscript` still returns a db path for many unexpected transcript paths because it only walks up the directory tree and checks for null parents. A deep path with the wrong file/segment names can still be mapped to some `<root>/conversations/<derived>.db` instead of failing open, so the watcher may poll a guessed SQLite file when Antigravity sends an unexpected `transcriptPath`. Since this helper is the guard for “derive the db only from `<root>/brain/<id>/.system_generated/logs/transcript_full.jsonl`”, please validate the filename and path segments (`transcript_full.jsonl`, `logs`, `.system_generated`, `brain`, and ideally the Antigravity root) before returning a db path.

realtonyyoung:
Re-reviewed latest head 16a8505; no tests/builds run. I added one inline finding on ConversationDbFromTranscript: it resolves the correct sibling db for the real Antigravity transcript shape, but it does not actually return null for many unexpected deep paths because it does not validate the path segments.

Fix confirmations:

1. C2 canonical id: stream/marker portion is sound. The hook now uses the dashless id for session_id, watcher key, and disabled-session check; kcap watch strips dashes and kcap disable normalizes to the same dashless marker. DB-from-transcript is correct for the real dashed conversation dir, with the shape-validation gap noted inline.
2. Q1 fail-open payload reads: sound. Production payload fields now use Str/AsString; I found no remaining GetValue string reads on Antigravity payload nodes.
3. Q2 control-hook exit: sound. Missing event logs guidance and returns 0.
4. C1 USAGE watermark: sound. AppendAntigravityUsageLines returns a staged max idx, and LastAntigravityGenIdx advances only after SendTranscriptBatch2 succeeds. Failed sends leave the watermark unchanged so rows re-read next drain; Antigravity no longer enters below-threshold buffering, and subagents never buffer.
5. C5 short-conversation idle-end: sound. Antigravity sets ThresholdReached true at watcher init, so short conversations stream live, final-drain, and idle-end/session-end. Title behavior still looks sane: first-user capture works, initial title can send immediately because the session already exists, and LLM title remains event-count gated.
6. C4 varint/length guards: sound for the reported issue. Token varints above long.MaxValue are rejected instead of cast negative, and length-delimited fields check remaining buffer before casting; I found no remaining unchecked ulong-to-long or unguarded int length casts on that path.
7. C3 deferred in-flight idle signal: acceptable follow-up, not a must-fix for this PR in my view. The risk remains if an Antigravity command produces no transcript line for longer than the idle window, but with the 60-minute default and no reliable transcript in-flight signal, deferring it is reasonable. Aggressively lowering KCAP_ANTIGRAVITY_IDLE_MINUTES would make this risk more visible.

realtonyyoung:
Addressed the re-review finding in cd562f9 (plus the S4 CLI half):

- **Path validation:** `ConversationDbFromTranscript` now validates every path segment (`transcript_full.jsonl` / `logs` / `.system_generated` / `brain`) so an unexpected `transcriptPath` fails open (null) instead of resolving to a guessed `conversations/<x>.db`. Tests cover the rejected shapes.
- **S4 anchor:** the watcher stamps each synthetic USAGE line with the conversation's most-recent transcript `created_at`, so the server anchors the backfill's recency to the turn rather than the write time.
- **C3 in-flight:** the idle-timeout session-end is now suppressed while an Antigravity tool is in flight — tracked from PLANNER_RESPONSE tool_calls vs. result steps, so a long command with no transcript line between call and result doesn't end the session mid-turn.

42 unit + 2 integration antigravity tests green; AOT clean.

PR #261, opened 2026-07-03
Cover multi-provider PR-detection edge cases; harden GitLab iid parse

### LINKED ISSUE #230: Add negative-path/edge-case unit tests for multi-provider PR detection
Follow-up to #229 (multi-provider PR/MR detection). The provider-detection paths are correct by inspection, but several negative/edge branches lack explicit unit tests. Add coverage:

- **`GitProviderRouter`** — malformed `gh auth status --json hosts` JSON, and a null runner result (probe failure/timeout), both fall through to the GitLab default. No test currently exercises that branch.
- **`GitHubPrDetector`** — JSON-parse-throws / non-numeric `number` → returns null (best-effort `catch`). No test.
- **`GitLabPrDetector`** — non-numeric `iid` (currently throws→caught→drops *all* matches, not just the bad record); a branch name containing `+` (encoding was only tested with `&`).
- **`RemoteMatcher.PathAfterHost`** — empty / lone-slash (`"host/"`) / leading-slash (`"/owner/repo"`) inputs (branches exist, traced-correct, untested).
- **Effective provider budget split** (added in #229) is timing-dependent and has no test — consider a small seam so a regression (reusing the full cap instead of the remainder) is caught by CI.
- **Test helper hygiene:** `RunGit` in `RepositoryDetectionCacheTests` doesn't drain `StandardOutput` — harmless for `init`/`remote add`, but a deadlock footgun if reused for a chattier git subcommand.

Low priority — all the underlying production branches are guarded; this is defense-in-depth coverage.

realtonyyoung:
## What & why

Follow-up to #229 (multi-provider PR/MR detection). The provider-detection paths were correct by inspection but several negative/edge branches had no explicit tests. This adds that defense-in-depth coverage, and fixes two small issues the coverage surfaced.

## Production changes (small)

- **`GitLabPrDetector` — non-numeric `iid`.** `mr["iid"]?.GetValue<int>()` *throws* on a non-integer `iid`, and the method's outer best-effort `catch` then dropped **every** matching MR, not just the bad record. Now guarded with `JsonValue.TryGetValue<int>` so a malformed record is skipped (`continue`) and the other valid MRs still resolve.
- **`RepositoryDetection` — testable budget-split seam.** Extracted the probe + detector block into an internal `ResolveAndDetectPrAsync` (behaviour-preserving) with an injectable timestamp. The detector still receives only the budget the probe left behind — but now that split is covered by a deterministic test, so a regression (handing the detector the full cap) is caught in CI rather than being timing-only.

## Test-helper hygiene

- `RepositoryDetectionCacheTests.RunGit` now drains **both** stdout and stderr before `WaitForExit` — harmless for `init`/`remote add`, but a deadlock footgun the moment it's reused for a chattier git subcommand.

## New coverage

- **`GitProviderRouter`** — malformed `gh auth status --json hosts` JSON, and a null probe result → GitLab fallback.
- **`GitHubPrDetector`** — malformed JSON and a non-numeric `number` → null.
- **`GitLabPrDetector`** — bad-`iid` record skipped (not the whole result); `+` in a branch name → `%2B` (encoding was only tested with `&`).
- **`RemoteMatcher.PathAfterHost`** — empty / `"host/"` / `"/owner/repo"` → null.
- **Provider budget split** — detector gets the remainder; no detection when the probe exhausts the cap.

## Testing

- Full unit suite green (**2246** tests). AOT publish of the CLI clean — no IL3050/IL2026 warnings.

No README/CLI-surface change: tests + an internal robustness fix + an internal refactor.

Closes #230
Linear: AI-1120

### DIFF
--- a/src/Capacitor.Cli/PrDetection/GitLabPrDetector.cs
+++ b/src/Capacitor.Cli/PrDetection/GitLabPrDetector.cs
@@ -31,13 +31,15 @@ internal static class GitLabPrDetector {
                 if (node is not JsonObject mr) continue;
                 // Defense in depth: only accept an exact source_branch match.
                 if (mr["source_branch"]?.GetValue<string>() != branch) continue;
-                var iid = mr["iid"]?.GetValue<int>();
-                if (iid is null) continue;
+                // Skip a record whose iid is missing or non-numeric rather than letting
+                // GetValue<int> throw — the outer catch would otherwise drop every match, not
+                // just the bad record.
+                if (mr["iid"] is not JsonValue iidVal || !iidVal.TryGetValue<int>(out var iid)) continue;
 
                 var updated = ParseTimestamp(mr["updated_at"]?.GetValue<string>());
                 if (best is null || updated > bestUpdated) {
                     bestUpdated = updated;
-                    best = new PrInfo(iid.Value, mr["title"]?.GetValue<string>(),
+                    best = new PrInfo(iid, mr["title"]?.GetValue<string>(),
                                       mr["web_url"]?.GetValue<string>(), mr["source_branch"]?.GetValue<string>());
                 }
             }
--- a/src/Capacitor.Cli/RepositoryDetection.cs
+++ b/src/Capacitor.Cli/RepositoryDetection.cs
@@ -244,8 +244,14 @@ static class RepositoryDetection {
         var kind  = await GitProviderRouter.ResolveAsync(host, cwd, providerCap, run);
 
         // The remainder after the probe — NOT the full providerCap. Passing providerCap here would
-        // let a slow probe + a full-length detector overrun the caller's deadline.
-        var detectCap = providerCap - Stopwatch.GetElapsedTime(start, getTs());
+        // let a slow probe + a full-length detector overrun the caller's deadline. Clamp the elapsed
+        // time to >= 0 so a non-monotonic timestamp seam can never inflate detectCap past providerCap
+        // (the invariant is "detector budget <= providerCap, always"). Production uses the monotonic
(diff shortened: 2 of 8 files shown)


realtonyyoung:
NO FINDINGS

Reviewed the PR diff against origin/main and verified the requested areas:
- GitLab iid guard accepts normal JSON integer iids and skips missing/non-numeric records without dropping valid MRs.
- ResolveAndDetectPrAsync preserves the old inline provider gating, budget split, early-outs, and GitLab owner/repo guard.
- Stopwatch.GetElapsedTime usage and injected timestamp seam are sound.
- New tests cover the claimed edge cases, including provider budget remainder and memoization isolation.
- RunGit stdout/stderr drain change avoids the deadlock shape while still surfacing stderr.
- Unit tests passed: 2246/2246.
- AOT publish passed unsandboxed with -p:UseSharedCompilation=false; no IL3050/IL2026 issues observed.

Conclusion: no actionable findings.

realtonyyoung:
Addressed both review notes in `7c7432e`:

1. **`PathAfterHost` lone-`"/"` case** — added an explicit assertion for `"/"` (it hits the same `slash <= 0` branch as `"/owner/repo"`, now pinned directly so the checklist case is covered).
2. **`ResolveAndDetectPrAsync` budget footgun** — clamp the probe's elapsed time to `>= 0` before subtracting, so a non-monotonic injected timestamp can't inflate `detectCap` past `providerCap` (the "detector budget ≤ providerCap" invariant). Production uses the monotonic `Stopwatch.GetTimestamp`, so this only hardens the test seam; added a test that a backwards-going timestamp keeps the detector budget within the ceiling.

Full unit suite green (2247); AOT publish clean.

PR #262, opened 2026-07-03
[AI-1160] Antigravity ingest (Phase 4, CLI): historical import

realtonyyoung:
**Draft — stacked on #256 (Phase 2 CLI).** Base is `tonyyoung/ai-1158-antigravity-cli` for a clean diff; retarget to `main` once #256 merges. Depends on Phase 1 (kcap-server#933) + Phase 2 (#256).

Phase 4 (AI-1160) of Google Antigravity ingest — historical import `kcap import --antigravity`, the last phase of the epic (parent AI-1150).

## What's here (3 commits)
- **`AntigravitySubagents.BuildParentMap`** (Core) — child→parent conversation map from each brain dir's `.system_generated/messages/*.json` (`{sender:child, recipient:parent}`). This linkage is unreliable at live child-start (why live nesting is deferred) but **complete on disk at import time**.
- **`AntigravityImportSource`** — mirrors `GeminiImportSource` (routed phase, JSONL transcript, **watermark-idempotent — no ledger** since the transcript is append-only). Discovers `~/.gemini/antigravity/brain/*` dirs, treats roots (non-children) as top-level sessions, and imports their children as subagents (`subagent-start` → child transcript routed by `agent_id` → `subagent-stop`, fail-closed). Wired into `Program.cs` `allSources` + `VendorSelection` (turns on the `--antigravity` import flag deferred from Phase 2).
- **`AntigravityImportTests`** (WireMock) + README/`help-import.txt`.

## Notes for review
- **Session id:** the import uses the raw (dashed) conversation id, which the server normalizes to the canonical dashless form in session-start/end, the transcript pipeline, AND the `/last-line` watermark probe (`CanonicalSessionId.Normalize` in all four). So a re-import and a live-captured session converge on the same dashless stream; no client-side dedup issue applies to import (unlike the live `disable` marker, which #256 fixed).
- **cwd** is left null (Antigravity records no machine-readable workspace path in the transcript — same v1 limitation as Gemini; live capture gets it from the hook payload).
- **Cost on import is deferred** (a documented follow-up): Antigravity's cost is off-transcript in `gen_metadata`, so importing it needs completeness tracking so a failed USAGE send after a complete transcript still retries. Imported sessions currently carry content + subagent nesting but not cost; live capture has cost.

## Testing
41 antigravity CLI unit tests + 2 import WireMock integration tests green (full import with subagent nesting + AlreadyLoaded idempotency). AOT publish clean on the Phase 2 base.

Linear: AI-1160. Closes # (GitHub issue to be linked).


src/Capacitor.Cli/Commands/AntigravityImportSource.cs
```diff
@@ -0,0 +1,367 @@
...
+                SourceMeta     = s.SourceMeta,
+            });
+        }
+
+        return results;
+    }
+
+    public async Task<ImportOutcome> ImportSessionAsync(
+            ImportCommand.SessionClassification c, ImportContext ctx, CancellationToken ct) {
+        if (c.Status == ImportCommand.ClassificationStatus.AlreadyLoaded) return ImportOutcome.Skipped;
```
realtonyyoung: Returning here means an AlreadyLoaded root never reaches ImportChildrenAsync. If the first import posts the parent transcript and session-end but a child was skipped (watermark probe failure, subagent-start failure, missing transcript, or transcript POST failure), the next run classifies the root from the parent watermark as AlreadyLoaded and permanently skips the child. That breaks the intended child-retry behavior; child completeness needs to participate in classification, or AlreadyLoaded roots with children need a child repair pass before skipping.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
```diff
@@ -0,0 +1,367 @@
...
+
+            // Fail-closed: no content unless the subagent registered first.
+            if (!await PostHookAsync(client, baseUrl, "subagent-start",
+                    BuildSubagentPayload("subagent_start", rootId, childId, childTranscript), ct))
+                continue;
+
+            try {
+                await SessionImporter.SendTranscriptBatches(
+                    httpClient: client, baseUrl: baseUrl, sessionId: rootId,
+                    filePath: childTranscript, agentId: childId, startLine: 0, vendor: Vendor,
```
realtonyyoung: Here the child HWM is applied as a line-number offset while still reading from line 0. For a child whose server HWM is 1, the next repair sends local lines 0 and 1 again as line_numbers 2 and 3, duplicating already imported content instead of resuming at raw line 2. Antigravity is append-only JSONL like the parent/Gemini path, so this should use the HWM as startLine (hwm + 1) rather than re-numbering the whole child transcript.

src/Capacitor.Cli.Core/Antigravity/AntigravitySubagents.cs
```diff
@@ -0,0 +1,58 @@
...
+                    var root = doc.RootElement;
+                    if (root.ValueKind != JsonValueKind.Object) continue;
+
+                    var sender    = root.TryGetProperty("sender",    out var s) ? s.GetString() : null;
+                    var recipient = root.TryGetProperty("recipient", out var r) ? r.GetString() : null;
+
+                    if (!string.IsNullOrEmpty(sender)
+                     && !string.IsNullOrEmpty(recipient)
+                     && !string.Equals(sender, recipient, StringComparison.Ordinal)
+                     && !map.ContainsKey(sender!)) {
```
realtonyyoung: Keeping the first parent for any sender makes invalid/non-tree linkages look valid. Because discovery later drops every key from top-level roots and only imports direct Children for each root, a chain like P<-C<-G imports C under P but never imports G, and a cycle A<->B produces no roots at all; conflicting parents are also nondeterministic because enumeration order is not guaranteed. We should either reject/ignore non-tree edges (multiple parents, cycles, child that is also a parent) or traverse/import recursively so these conversations are not silently lost or mis-nested.

src/Capacitor.Cli.Core/Resources/help-import.txt
```diff
@@ -10,7 +10,8 @@ machine and imports them in a single run. Each agent ships a per-session .jsonl
   Gemini  — ~/.gemini/tmp/<project>/chats/session-<ISO>-<shortId>.jsonl
   Kiro    — ~/.kiro/sessions/cli/<sid>.jsonl (+ <sid>.json metadata; honours KIRO_HOME)
   Pi      — ~/.pi/agent/sessions/.../<timestamp>_<sid>.jsonl (badlogic/pi-mono)
-kcap walks all seven the same way.
+  Antigravity — ~/.gemini/antigravity/brain/<id>/.system_generated/logs/transcript_full.jsonl
+kcap walks all eight the same way.
```
realtonyyoung: This now says all eight, but Program registers nine import sources and the vendor list below includes both OpenCode and Antigravity. The source table above also still omits the OpenCode DB path, so `kcap import --help` understates what the command actually walks.

realtonyyoung (commented):
Reviewed only the three commits unique to PR #262 on top of #256. I found issues in Antigravity subagent retry/idempotency, child watermark resume, non-tree parent-map handling, and the import help text. Per request, I did not run tests or builds.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
realtonyyoung: Fixed in 3280754. `ImportSessionAsync` no longer returns early for `AlreadyLoaded`: it now runs `ImportChildrenAsync` as a repair pass (each child classifies independently by its own `(rootId, childId)` HWM) before reporting `Skipped`, so a subagent skipped on a prior run is retried instead of lost. New test: `ImportSession_AlreadyLoaded_root_repairs_a_missing_child`.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
realtonyyoung: Fixed in 3280754. Children now resume via `startLine: hwm + 1` with `lineNumberOffset: 0` (removed), so surviving lines keep their true file position — matching the parent/Gemini resume. Also skips a child whose HWM already covers its last import-relevant line before touching any hook. New test: `ImportChildren_resumes_by_line_position_without_shifting_numbers` asserts a resumed child posts `line_numbers:[1]`, not the shifted `[1,2]`.

src/Capacitor.Cli.Core/Antigravity/AntigravitySubagents.cs
realtonyyoung: Fixed in 3280754. `BuildParentMap` is now deterministic on conflicts (smallest-ordinal parent wins). Added `ResolveTopLevelAncestor` (cycle-safe walk-up — a cycle member resolves to itself) and `BuildRootDescendants` (root → all transitive descendants). Discovery enumerates every brain dir and nests grandchildren+; so P<-C<-G imports both C and G under P, and A<->B import standalone rather than vanishing. New tests cover deterministic conflict, deep-chain, cycle, and transitive discovery.

src/Capacitor.Cli.Core/Resources/help-import.txt
realtonyyoung: Fixed in 3280754. Help text now says "all nine", adds OpenCode/Antigravity to the intro sentence, and adds the OpenCode SQLite db source line (`~/.local/share/opencode/opencode.db`) to the path table.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
```diff
@@ -0,0 +1,391 @@
...
+            // endless re-send, and a fully-ingested child is a no-op we skip before touching hooks.
+            int? childLastImportable;
+            try {
+                var (lastNonBlank, lastRelevant, _) = await ReadTranscriptStatsAsync(childTranscript, ct);
+                childLastImportable = lastRelevant ?? lastNonBlank;
+            } catch {
+                continue; // unreadable child transcript — retry on re-import
+            }
+            if (childLastImportable is null) continue;                       // empty child
+            if (chwm is { } done && done >= childLastImportable) continue;   // already fully ingested
```
realtonyyoung: This HWM-only skip makes fully ingested child content imply a complete child lifecycle, but `subagent-stop` is still best-effort below: if the child transcript POST succeeds and then the stop hook returns non-2xx or throws, the next import sees `done >= childLastImportable` here and skips before retrying any hook. That leaves the subagent without its completion event permanently, even though the failure happened after content import and should be repairable on re-import. Consider either treating stop failure as an incomplete child or still posting an idempotent stop for fully-loaded children when lifecycle completeness is not otherwise known.

realtonyyoung (commented):
Re-reviewed commit 3280754 and its interaction with the Antigravity import path. The four prior fixes look correct for parent skip/child repair, child resume-by-position, transitive subagent grouping, and help text/vendor count, but I found one remaining lifecycle retry issue in the new fully-loaded-child skip path. Per request, I did not run tests or builds.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
realtonyyoung: Fixed in a125072. A fully-ingested child no longer returns before the stop hook: it now re-posts subagent-stop (idempotent server-side — deterministic `SubagentCompleted` id via `AgentLifecycleDeterministicId`) so a prior run's failed-after-content stop is repaired, while an already-recorded stop dedupes. subagent-start is not re-posted (content is fail-closed behind it, so its success is implied by the ingested content) and the child content is not re-sent. New test: `ImportChildren_fullyIngestedChild_reposts_idempotent_stop_without_resending_content` asserts subagent-stop is posted but neither transcript nor subagent-start.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
```diff
@@ -0,0 +1,404 @@
...
+
+            if (chwm is { } done && done >= childLastImportable) {
+                // Content is fully ingested, but subagent-stop is best-effort below and may have
+                // failed AFTER the content POST on a prior run — leaving the subagent with no
+                // completion event permanently (HWM only tracks content lines, not lifecycle).
+                // Re-post the stop: it's idempotent server-side (deterministic SubagentCompleted
+                // id), so a prior failure is repaired and an already-recorded stop dedupes.
+                // subagent-start is not re-posted — content is fail-closed behind it, so its
+                // presence is implied by the ingested content (AI-1160 review).
+                await PostHookAsync(client, baseUrl, "subagent-stop",
```
realtonyyoung: This stop-only repair still depends on the server remembering the child as active. Server-side `RecordAgentStopAsync` returns before appending `SubagentCompleted` when `SessionStateTracker.TryClearActiveAgent(sessionId, agentId)` is false, and that tracker is transient in-memory state. So if the first import successfully posted `subagent-start` and child content, then `subagent-stop` failed and the server restarts before the re-import, this branch posts only stop, the server treats it as inactive/duplicate, and no completion event is written; the child HWM remains content-complete, so future imports repeat the no-op. To make the repair durable across process boundaries, the fully-ingested branch needs to re-post the idempotent start before stop (or the server stop endpoint needs to allow deterministic stop repair without an active mark).

realtonyyoung (commented):
Re-reviewed commit a125072. The focused change prevents content and subagent-start from being resent and keeps AlreadyLoaded-root outcome as Skipped, but I found one remaining repair gap around the server's transient active-agent guard for stop-only retries. I did not run tests or builds.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
realtonyyoung: Fixed in 57065cd. The fully-ingested branch now re-posts subagent-start THEN subagent-stop. Start re-marks the agent active (idempotent — deterministic SubagentStarted id via AgentLifecycleDeterministicId; already-active is a no-op), so a stop that follows clears the mark and appends SubagentCompleted even after a server restart wiped the transient active mark. This makes the completion repair durable across process boundaries. Child content is not re-sent. Test updated: `ImportChildren_fullyIngestedChild_reposts_idempotent_start_and_stop_without_resending_content` (asserts start-before-stop, no transcript).

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
```diff
@@ -0,0 +1,409 @@
...
+                // failed AFTER the content POST on a prior run — leaving the subagent with no
+                // completion event (HWM only tracks content lines, not lifecycle). Repair it by
+                // re-posting subagent-start THEN subagent-stop. The server's stop endpoint no-ops
+                // unless the agent is marked active, and that mark is transient in-memory — lost on
+                // a server restart between the failed stop and this re-import. Re-posting start
+                // restores the active mark (idempotent: deterministic SubagentStarted id, and
+                // already-active is a no-op) so the following stop clears it and appends the
+                // (idempotent, deterministic) SubagentCompleted. Child content is not re-sent
+                // (AI-1160 review).
+                if (await PostHookAsync(client, baseUrl, "subagent-start",
```
realtonyyoung, line 272: [P2] Make the repair start hook strict. This branch treats a successful subagent-start POST as proof that the server re-marked the child active before posting stop, but BuildSubagentPayload does not send strict=true. On the server, SubagentStartHook.Strict defaults false, and HandleSubagentStart returns 200 even when RecordAgentStartAsync throws after rolling back the active mark. In that case this code still posts subagent-stop, which no-ops without an active agent, so the durable stop repair can silently fail. Please include strict=true on the synthetic subagent lifecycle payloads so lifecycle write failures surface as non-2xx and are retried.

realtonyyoung (commented):
Found one remaining issue: the new start-before-stop repair still needs strict lifecycle payloads so server-side start failures are visible to the importer. I did not run tests or builds.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
realtonyyoung: Fixed in 93be269. The repair now sends strict=true on both synthetic subagent lifecycle payloads (BuildSubagentPayload gained a strict param). HandleSubagentStart returns 500 under strict when RecordAgentStartAsync throws+rolls back the active mark, so a start that actually failed no longer reads as 200 — stop is posted only when start truly re-marked the agent active, and a re-import retries otherwise. Test asserts the repair subagent-start body carries "strict":true.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+    static JsonObject BuildSessionEndPayload(string sid, string? cwd, DateTimeOffset? endedAt) {
+        var p = new JsonObject { ["hook_event_name"] = "sessionEnd", ["session_id"] = sid, ["reason"] = "antigravity-import" };
+        if (cwd is not null) p["cwd"] = cwd;
+        if (endedAt is { } ts) p["ended_at"] = ts.ToString("O");
+        return p;
+    }
+
+    static JsonObject BuildSubagentPayload(string eventName, string parentSid, string agentId, string transcriptPath, bool strict = false) {
```
realtonyyoung: [P1] Build the full stop-hook payload before relying on strict. This builder is now used for the strict repair subagent-stop POST, but it still only emits the start-shaped fields. The server stop route binds SubagentStopHook, whose required fields include stop_hook_active, agent_transcript_path, and last_assistant_message; Cursor/Gemini/OpenCode send those because an incomplete body can be rejected before HandleSubagentStop runs. If this Antigravity stop payload is rejected at binding, strict is never honored by the stop handler, PostHookAsync returns false, and this repair repeats start->failed-stop forever while the child remains without SubagentCompleted. Please split start/stop payload builders or add the stop-only fields for subagent_stop, and cover the stop body in the repair test.

realtonyyoung (commented):
Found one remaining issue: the strict repair payloads are now gated correctly for start, but the stop payload still needs the full SubagentStopHook shape before the server can honor strict on the stop route. I did not run tests or builds.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
realtonyyoung: Fixed in a0ee6d2. Split the payload builder into BuildSubagentStartPayload / BuildSubagentStopPayload; the stop builder now emits the full SubagentStopHook shape (stop_hook_active, agent_transcript_path, last_assistant_message) mirroring GeminiSubagentDiscovery.BuildStopPayload, so the strict stop body binds and the strict handler actually runs (no more start→failed-stop loop). Test asserts the repair stop body carries those fields alongside strict:true.

realtonyyoung:
NO FINDINGS

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
```diff
@@ -0,0 +1,444 @@
...
+
+        // An AlreadyLoaded root's parent transcript is fully imported, but a subagent may have
+        // been skipped on a prior run (watermark probe / subagent-start / transcript POST
+        // failure). Children classify independently by their own HWM, so still run a child-repair
+        // pass — complete children are no-ops, incomplete ones resume — then report Skipped.
+        // Without this, a once-skipped child would be lost forever (AI-1160 review).
+        if (c.Status == ImportCommand.ClassificationStatus.AlreadyLoaded) {
+            await ImportChildrenAsync(ctx.HttpClient, ctx.BaseUrl, c.SessionId, c.SourceMeta!, ct);
+            return ImportOutcome.Skipped;
+        }
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in efb1512 — the `AlreadyLoaded` branch now re-posts `session-start` → repairs children → re-posts `session-end` (all idempotent via deterministic lifecycle ids), and returns `Failed` if either lifecycle POST fails so a run that advanced the watermark but failed `session-end` is repaired on re-run. Both AlreadyLoaded integration tests updated to expect the re-assertion.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    static DateTimeOffset? TryGetLastWriteUtc(string path) {
+        try { return File.GetLastWriteTimeUtc(path); } catch { return null; }
+    }
+
+    static async Task<int?> FetchServerLastLineAsync(
+            HttpClient http, string baseUrl, string sessionId, string? agentId, CancellationToken ct) {
+        var url = $"{baseUrl}/api/sessions/{sessionId}/last-line" + (agentId is not null ? $"?agentId={agentId}" : "");
+        using var resp = await http.GetWithRetryAsync(url, ct: ct);
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in efb1512 — `FetchServerLastLineAsync` now builds the query with `Uri.EscapeDataString(agentId)`, matching `CursorImportSource`.

src/Capacitor.Cli/Commands/AntigravityImportSource.cs
```diff
@@ -0,0 +1,444 @@
...
+
+        var sinceUtc = filters.Since is { } since
+            ? new DateTimeOffset(since.ToDateTime(TimeOnly.MinValue, DateTimeKind.Utc), TimeSpan.Zero)
+            : (DateTimeOffset?)null;
+
+        foreach (var (convId, descendants) in byRoot) {
+            ct.ThrowIfCancellationRequested();
+
+            if (filters.FilterSession is { } fs && !string.Equals(convId, fs, StringComparison.Ordinal)) continue;
+
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in efb1512 — `BuildParentMap` now takes a `CancellationToken` and checks it between brain dirs and message files; `DiscoverAsync` threads its token through (added a cancellation unit test). Kept the full scan itself: correct subagent nesting requires it — a child records its parent, so a root's descendants can only be found by reading every link — but it's now interruptible.

PR #265, opened 2026-07-03
Fix Windows hosted-agent hang at trust dialog (empty terminal / "Waiting for session to start")

stktung:
## Problem

On **Windows**, a launched hosted agent shows **Active** in the web UI, but the **Terminal tab stays empty** and Chat is stuck on **"Waiting for session to start…"** forever. macOS/Linux are unaffected.

Screenshot 2026-06-30 024239

Screenshot 2026-06-30 020941

## Root cause

The agent never starts a Claude session, so the SessionStart hook never fires and the agent's `SessionId` never links — and both the Terminal tab and the "Waiting for session to start" message are gated on that link.

The reason no session starts: the spawned `claude` hangs at the workspace-trust dialog (*"Quick safety check: Is this a project you created or one you trust?"*), waiting for input that never comes. Confirmed by reading the agent's terminal buffer off the server — Claude was parked at the trust prompt.

`ClaudeLauncher.TrustWorktreeInClaudeConfig` pre-accepts trust by writing `projects["<worktree>"].hasTrustDialogAccepted = true` into `~/.claude.json`. But Claude Code normalises the `projects[]` lookup key, and on Windows that normalisation converts backslashes to forward slashes:

```js
function normalizeKey(e){ let t = path.normalize(e); if (os === "windows") return t.replaceAll("\\","/"); return t }
```

So Claude looks up `C:/Users/.../agent-xxx` (forward slashes) while the daemon wrote `C:\Users\...\agent-xxx` (backslashes). The keys never match, the pre-trust write is invisible, and the trust dialog blocks. POSIX paths already use `/`, which is why this only reproduces on Windows. (Verified the daemon *did* write `hasTrustDialogAccepted: true` under the backslash key and Claude still prompted.)

## Fix

Write the `projects[]` trust key in Claude's normalised (forward-slash on Windows) form via a small `NormalizeClaudeProjectKey` helper. AOT-safe (`Path.GetFullPath` + `string.Replace` + `OperatingSystem.IsWindows`).

## Verification

Reproduced locally on Windows (empty terminal + "Waiting for session to start"; terminal buffer showed the trust prompt). After the fix, a freshly launched agent linked its session within ~10s:

```
Updated agent c7f43525 status to Running (session=257badde…)
Linked session 257badde… to daemon agent c7f43525
```

and the Terminal now renders the live Claude UI ("Claude Code v2.1.200 / Welcome back …"). Pre-fix agents stayed `session=null` indefinitely.

## Related (not fixed here)

The same key-mismatch pattern affects `WriteMcpConfig`, which reads `projects[sourceRepoPath]` with the raw path — so a user's interactively-configured MCP servers are not copied into hosted worktrees on Windows. Worth a follow-up with the same normalisation.

### DIFF
--- a/src/Capacitor.Cli.Daemon/Services/ClaudeLauncher.cs
+++ b/src/Capacitor.Cli.Daemon/Services/ClaudeLauncher.cs
@@ -291,6 +291,17 @@ internal sealed partial class ClaudeLauncher(
         }
     }
 
+    /// <summary>
+    /// Matches Claude Code's <c>projects[]</c> key normalisation: absolute + collapsed, and
+    /// on Windows with forward slashes (Claude runs <c>path.normalize(p).replaceAll("\\","/")</c>).
+    /// Writing the raw Windows path instead makes our pre-trust write invisible to Claude.
+    /// </summary>
+    internal static string NormalizeClaudeProjectKey(string path) {
+        var full = Path.GetFullPath(path);
+
+        return OperatingSystem.IsWindows() ? full.Replace('\\', '/') : full;
+    }
+
     static void TrustWorktreeInClaudeConfig(string worktreePath) {
         // Serialize against concurrent agent launches. ~/.claude.json is shared
         // across the whole user and {worktree}/.claude/settings.local.json is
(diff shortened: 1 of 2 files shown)


stktung:
Addressed in b4f8b4c: `WriteMcpConfig` now looks up `projects[]` via `NormalizeClaudeProjectKey(sourceRepoPath)` first (matching the trust write), with a raw-path fallback for entries written by older builds or by hand. Added `ClaudeLauncherWriteMcpConfigTests` covering the normalized-key lookup (incl. env stripping), the raw-key fallback, and the no-entry case — on Windows these exercise the real backslash/forward-slash divergence. Full unit suite: 2240/2240 green.

PR #296, opened 2026-07-08
[AI-699] Auto-register kcap MCP servers for Cursor

realtonyyoung:
First per-harness rollout of the MCP auto-config epic (AI-1224) — wires the merged config-writer foundation (AI-1225) into the **Cursor** install path.

## What
- `kcap setup` and `kcap plugin install --cursor` now auto-register the 4 kcap MCP servers into `~/.cursor/mcp.json` (Standard shape: `mcpServers` map, `command: "kcap"` + `args`, no cwd — it's a user-global config). Uses `JsonMcpConfigWriter` + `KcapMcpServers.All` (Cursor is Claude-capable → includes `kcap-flows`) + `McpConfigShape.Standard` + `McpMarker("cursor")`.
- Non-destructive (preserves user-authored servers), idempotent, atomic; `kcap plugin remove --cursor` / `kcap uninstall` unregister the kcap entries + clear the sidecar marker.
- New `--skip-cursor-mcp` flag; wired through `CodingAgentsStep` (delegate + result flag + `HandleCursorMcp`), `SetupCommand`, `PluginCommand` (`InstallCursor`/`RemoveCursor`), `UninstallCommand`. Mirrors the existing Codex MCP-registration wiring.
- Docs: `--skip-cursor-mcp` added to help text; README's MCP auto-registration section now lists Cursor.

## Tests
- Unit: CodingAgentsStep (Cursor MCP registered / not-when-hooks-fail / skipped / failure-warning), PluginCommandCursor (install writes mcp.json + preserves a user server, `--if-installed` skips, remove unregisters), CursorPaths, UninstallCommand marker-clear. Full suite **2641/2641**, build clean (0 warnings, AOT-safe).
- **Live E2E (dev build vs real Cursor):** config write/idempotent/non-destructive/remove all verified against a real `~/.cursor`; stdio handshake confirmed AI-1233 behavior (version negotiation, `resources/list`/`prompts/list`/`ping` answered); and a real `search_sessions` call ("have we worked on ACP before") executed **through Cursor** and returned results.

Part of AI-1224. Depends on the merged foundations AI-1225 (#282) + AI-1233 (#284).

### DIFF
--- a/src/Capacitor.Cli/Commands/PluginCommand.cs
+++ b/src/Capacitor.Cli/Commands/PluginCommand.cs
@@ -8,6 +8,7 @@ using Capacitor.Cli.Core.Copilot;
 using Capacitor.Cli.Core.Cursor;
 using Capacitor.Cli.Core.Gemini;
 using Capacitor.Cli.Core.Kiro;
+using Capacitor.Cli.Core.Mcp;
 using Capacitor.Cli.Core.OpenCode;
 using Capacitor.Cli.Core.Pi;
 
@@ -694,33 +695,71 @@ public static class PluginCommand {
                 : $"Cursor hooks installed ({hooksPath})"
         );
 
+        // Register the kcap MCP servers in ~/.cursor/mcp.json so Cursor picks them up
+        // with no manual JSON edit. Non-destructive + idempotent. Never fails the
+        // install: a write error is a warning, not an error code (mirrors Codex).
+        if (!args.Contains("--skip-cursor-mcp"))
+            await RegisterCursorMcpServersAsync(env);
+
         return 0;
+    }
+
+    /// <summary>
+    /// Registers the kcap MCP servers in <c>~/.cursor/mcp.json</c> so Cursor loads them
+    /// without a manual JSON edit. Never fails the install: a write error is a warning,
+    /// not an error code.
+    /// </summary>
+    static async Task RegisterCursorMcpServersAsync(PluginEnvironment env) {
+        var change = JsonMcpConfigWriter.Register(
+            env.CursorMcpJson, KcapMcpServers.All, McpConfigShape.Standard, cwd: null, new McpMarker("cursor"));
+
+        switch (change) {
+            case JsonMcpConfigWriter.Change.Updated:
+                await env.Stdout.WriteLineAsync($"Cursor MCP servers registered ({env.CursorMcpJson}).");
+                break;
+            case JsonMcpConfigWriter.Change.Failed:
+                await env.Stderr.WriteLineAsync(
+                    $"Warning: could not update {env.CursorMcpJson} to register Cursor MCP servers.");
+                break;
+            // Unchanged: silent — same as Codex's already-registered case.
+        }
     }
 
     static async Task<int> RemoveCursor(string[] args, PluginEnvironment env) {
         var hooksPath = GetArg(args, "--cursor-hooks-path") ?? env.CursorUserHooksJson;
 
+        var hooksFailed = false;
+
         if (!File.Exists(hooksPath)) {
             await env.Stdout.WriteLineAsync("Nothing to remove — Cursor hooks file not found.");
+        } else {
+            try {
+                var removed = RemoveCursorHooks(hooksPath);
 
-            return 0;
+                await env.Stdout.WriteLineAsync(
+                    removed
+                        ? $"Cursor hooks removed ({hooksPath})"
+                        : "Cursor hooks were not installed."
+                );
+            } catch (Exception ex) {
+                await env.Stderr.WriteLineAsync($"Could not update Cursor hooks at {hooksPath}: {ex.Message}");
+                hooksFailed = true;
+            }
         }
 
-        try {
-            var removed = RemoveCursorHooks(hooksPath);
+        // Cursor is user-scope only (no --project split like Codex), so the kcap MCP
+        // entries are always unregistered here, independent of whether hooks.json
+        // existed — the two files are unrelated on disk.
+        var mcpChange = JsonMcpConfigWriter.Unregister(env.CursorMcpJson, McpConfigShape.Standard, new McpMarker("cursor"));
+        var mcpFailed = mcpChange == JsonMcpConfigWriter.Change.Failed;
 
-            await env.Stdout.WriteLineAsync(
-                removed
-                    ? $"Cursor hooks removed ({hooksPath})"
-                    : "Cursor hooks were not installed."
-            );
-
-            return 0;
-        } catch (Exception ex) {
-            await env.Stderr.WriteLineAsync($"Could not update Cursor hooks at {hooksPath}: {ex.Message}");
-
-            return 1;
+        if (mcpChange == JsonMcpConfigWriter.Change.Updated) {
+            await env.Stdout.WriteLineAsync($"Cursor MCP servers removed ({env.CursorMcpJson})");
+        } else if (mcpFailed) {
+            await env.Stderr.WriteLineAsync($"Could not update {env.CursorMcpJson} to remove Cursor MCP servers.");
         }
+
+        return hooksFailed || mcpFailed ? 1 : 0;
     }
 
     /// <summary>
--- a/src/Capacitor.Cli/Commands/UninstallCommand.cs
+++ b/src/Capacitor.Cli/Commands/UninstallCommand.cs
@@ -2,7 +2,6 @@ using Capacitor.Cli.Core;
 using Capacitor.Cli.Core.Cursor;
 using Capacitor.Cli.Core.Gemini;
 using Capacitor.Cli.Core.Kiro;
-using Capacitor.Cli.Core.Mcp;
 using Capacitor.Cli.Core.OpenCode;
 using Capacitor.Cli.Services;
 
@@ -133,17 +132,23 @@ public static class UninstallCommand {
         // hooks file).
         if (await PluginCommand.HandleAsync(["plugin", "remove", "--skills"]) != 0) hadFailures = true;
 
-        // Belt-and-braces marker cleanup. Plugin remove deletes the marker
-        // only when JSON entries changed; if the user manually pruned the
+        // Belt-and-braces marker cleanup. These hooks installers delete their
+        // marker only when JSON entries changed; if the user manually pruned the
         // entries earlier (or installed via a pre-marker build that later
         // wrote a marker on first refresh), the marker survives and IsInstalled
         // still reports kcap as installed. uninstall promises a full
         // wipe, so always nuke the markers regardless of what the JSON state
         // looked like going in.
+        //
+        // The Cursor MCP marker is intentionally NOT swept here: `plugin remove
+        // --cursor` above owns it via JsonMcpConfigWriter.Unregister, which clears
+        // it on any non-Failed outcome (including hand-pruned entries) but
+        // deliberately RETAINS it when the unregister fails, so a retry can still
+        // identify the kcap-owned entries. An unconditional clear here would
+        // defeat that recovery path.
         ClaudePluginInstaller.DeleteMarker(ClaudePaths.UserSettings);
         CodexHooksInstaller.DeleteMarker(CodexPaths.UserHooksJson);
         CursorHooksInstaller.DeleteMarker(CursorPaths.UserHooksJson());
-        new McpMarker("cursor").Clear(CursorPaths.UserMcpJson());
         GeminiHooksInstaller.DeleteMarker(GeminiPaths.SettingsJson());
         KiroHooksInstaller.DeleteMarker(KiroPaths.KcapAgentJson());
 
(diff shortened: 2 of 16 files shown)


src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -694,33 +695,71 @@ await env.Stdout.WriteLineAsync(
...
+            }
         }
 
-        try {
-            var removed = RemoveCursorHooks(hooksPath);
+        // Cursor is user-scope only (no --project split like Codex), so the kcap MCP
+        // entries are always unregistered here, independent of whether hooks.json
+        // existed — the two files are unrelated on disk.
+        var mcpChange = JsonMcpConfigWriter.Unregister(env.CursorMcpJson, McpConfigShape.Standard, new McpMarker("cursor"));
+        var mcpFailed = mcpChange == JsonMcpConfigWriter.Change.Failed;
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in `ef5425d` — centrally, not just in RemoveCursor: `JsonMcpConfigWriter.Unregister` now clears the ownership marker whenever unregister runs (not only when it removed JSON entries), skipping only on a hard `Failed`. So `remove --cursor` — and every future harness's unregister — clears an orphaned marker even after a user hand-edits the kcap entries out of the config. Covered by the new `Unregister_clears_marker_even_when_no_kcap_entries_to_remove` test.

test/Capacitor.Cli.Tests.Unit/PluginCommandCursorTests.cs
```diff
@@ -32,6 +34,120 @@ public async Task install_cursor_if_installed_short_circuits_on_same_version_mar
...
     }
 
+    // AI-699: `plugin install/remove --cursor` also (un)registers the kcap MCP
+    // servers in ~/.cursor/mcp.json. These use an explicit PluginEnvironment
+    // (not PluginEnvironment.FromProcess()) so env.CursorMcpJson resolves under
+    // a temp home instead of the real machine's ~/.cursor — mirrors
+    // PluginCommandCodexInstallIntegrationTests.TestEnv. The `--if-installed`
+    // refresh branch is used (pre-marker hooks.json seeded) rather than a bare
+    // `install --cursor`, so the AgentDetector "kcap on PATH" precheck (which
+    // only runs on the non-refresh path) never comes into play.
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in `ef5425d` — removed the `AI-699` references from the new test comments in `PluginCommandCursorTests.cs` and `UninstallCommandTests.cs` (rephrased without the Linear identifier).

src/Capacitor.Cli/Commands/UninstallCommand.cs
```diff
@@ -142,6 +143,7 @@ await Console.Out.WriteLineAsync(
         ClaudePluginInstaller.DeleteMarker(ClaudePaths.UserSettings);
         CodexHooksInstaller.DeleteMarker(CodexPaths.UserHooksJson);
         CursorHooksInstaller.DeleteMarker(CursorPaths.UserHooksJson());
+        new McpMarker("cursor").Clear(CursorPaths.UserMcpJson());
```
realtonyyoung: This unconditional marker clear can make a failed Cursor MCP unregister unrecoverable. `plugin remove --cursor` now intentionally keeps the ownership marker when `JsonMcpConfigWriter.Unregister` returns `Failed` (for example, `~/.cursor/mcp.json` is valid but cannot be rewritten, or is temporarily unreadable/malformed) so a retry can still identify the kcap-owned `kcap-*` entries. `uninstall` records that failure in `hadFailures`, but then this line deletes the marker anyway; after the user fixes the file/permissions and reruns uninstall/remove, the writer no longer knows those existing entries are kcap-owned and preserves them as user-authored. Please only clear the Cursor MCP marker when the unregister did not fail, or let `RemoveCursor` own this cleanup, and add a failure-path retry test.

realtonyyoung: Fixed in `eebe779`. Chose the second remediation — removed the unconditional sweep line and let `plugin remove --cursor` (called by `uninstall` just above) own the Cursor MCP marker cleanup via `JsonMcpConfigWriter.Unregister`, which clears the marker on any non-`Failed` outcome (including hand-pruned entries) but deliberately retains it on `Failed` so a retry can still identify the `kcap-*` entries. Also dropped the now-unused `using`. Added two regression tests: `UninstallCommandTests.User_level_uninstall_keeps_cursor_mcp_marker_when_unregister_fails` (uninstall retains the marker on a failed unregister) and `PluginCommandCursorTests.remove_cursor_retains_marker_on_failed_unregister_then_retry_removes_entries` (full fail → fix → retry-removes path). Full unit suite 2644/2644.

test/Capacitor.Cli.Tests.Unit/PluginCommandCursorTests.cs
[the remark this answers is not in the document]
realtonyyoung: Removed a stray `AI-699` I introduced in `eebe779` (`UninstallCommandTests.cs:471`) — rephrased to `(self-review)` in `aa660eb`. The originally-flagged comments were already cleaned in `ef5425d`; `grep -rn AI-699 src/ test/` is now empty.

test/Capacitor.Cli.Tests.Unit/PluginCommandCursorTests.cs
```diff
@@ -32,6 +34,151 @@ public async Task install_cursor_if_installed_short_circuits_on_same_version_mar
...
+        await Assert.That(retryExit).IsEqualTo(0);
+
+        var root    = JsonNode.Parse(await File.ReadAllTextAsync(mcpPath))!.AsObject();
+        var servers = root["mcpServers"] as JsonObject;
+        var keys    = servers?.Select(kv => kv.Key).ToArray() ?? [];
+        await Assert.That(keys).DoesNotContain("kcap-review");
+        await Assert.That(new McpMarker("cursor").Owned(mcpPath).ToArray()).IsEmpty();     // marker cleared after clean removal
+    }
+
+    static PluginEnvironment TestEnv(string fakeHome) => new(
```
realtonyyoung, line 175: `TestEnv` redirects `PluginEnvironment.HomeDirectory`, but `McpMarker("cursor")` does not use that value; it derives its storage root from `Environment.GetFolderPath(UserProfile)`. Because these fake homes are created under the temp directory, the marker treats `fakeHome/.cursor/mcp.json` as non-user-scope and writes ownership records to the real `~/.kcap/mcp-markers/...`. The install test above registers MCP servers and never unregisters them, so running this test leaves marker files in the developer/CI home. Please isolate the process HOME for these tests or inject a marker path so all sidecar state stays under `fakeHome` and is cleaned with the fixture.

realtonyyoung: Fixed in `800f053`. `TempDir` now pins the process `HOME` to itself for its lifetime (restored on `Dispose`), so `McpMarker` — which derives its storage from `Environment.GetFolderPath(UserProfile)` (`$HOME` on Unix), not `PluginEnvironment.HomeDirectory` — keeps all sidecar/central marker state under the temp home, for both the test's direct calls and the production `plugin --cursor` path. Verified: a full `PluginCommandCursorTests` run adds zero files to the real `~/.kcap/mcp-markers`.

test/Capacitor.Cli.Tests.Unit/PluginCommandCursorTests.cs
```diff
@@ -32,11 +34,174 @@ public async Task install_cursor_if_installed_short_circuits_on_same_version_mar
...
+    // PluginEnvironment.HomeDirectory. Without this, on Unix a temp-dir config is
+    // treated as non-user-scope and the ownership marker leaks into the real
+    // ~/.kcap/mcp-markers. Pinning HOME keeps all marker state under this dir (for
+    // both the test's direct McpMarker calls and the production plugin --cursor
+    // path) so it's cleaned on Dispose. Safe because the class is
+    // [NotInParallel("HomeEnvVarMutation")].
     sealed class TempDir : IDisposable {
+        readonly string? _originalHome;
         public string Path { get; } = System.IO.Path.Combine(
             System.IO.Path.GetTempPath(),
```
realtonyyoung: This still leaks marker state on Windows when the process temp directory is outside the user profile. `McpMarker` uses `Environment.GetFolderPath(UserProfile)`, which ignores the HOME variable on Windows, while this fixture puts `fakeHome` under `Path.GetTempPath()`. If a developer or CI sets `TMP`/`TEMP` to something like `D:\Temp` or `C:\Temp`, the config path is not under `UserProfile`, so the marker falls back to the real `%USERPROFILE%\.kcap\mcp-markers` again; the install test registers MCP servers and never unregisters them, leaving that file behind. Please create the fake home under `Environment.GetFolderPath(UserProfile)` on Windows (or inject a marker path) so the marker becomes a sidecar under the temp home and is deleted with the fixture.

realtonyyoung: Fixed in `26722f2`. You're right — pinning `$HOME` doesn't help on Windows (`GetFolderPath(UserProfile)` ignores it) and `Path.GetTempPath()` isn't guaranteed under the profile. Rooted the fake home under `Environment.GetFolderPath(UserProfile)` instead, so the config is always user-scope and `McpMarker` writes its marker as a sidecar under the test dir on every OS (both the test's direct calls and the production `plugin --cursor` path); it's removed with the dir on `Dispose`. Dropped the now-unnecessary `HOME` mutation. Verified on Unix: a full run adds zero files to the real `~/.kcap/mcp-markers` and leaves no test dirs under the home.

test/Capacitor.Cli.Tests.Unit/PluginCommandCursorTests.cs
```diff
@@ -32,9 +34,165 @@ public async Task install_cursor_if_installed_short_circuits_on_same_version_mar
...
+    // TEMP=D:\Temp), which would otherwise send the marker to the real
+    // ~/.kcap/mcp-markers. Rooting here keeps the marker a sidecar under this dir
+    // on every OS — for both the test's direct McpMarker calls and the production
+    // `plugin --cursor` path — so it's removed with the dir on Dispose. The class
+    // is [NotInParallel("HomeEnvVarMutation")] so the profile/$HOME McpMarker reads
+    // can't shift underneath it.
     sealed class TempDir : IDisposable {
         public string Path { get; } = System.IO.Path.Combine(
-            System.IO.Path.GetTempPath(),
+            Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
```
realtonyyoung: This still has one isolation hole: if the real user profile is itself a git worktree, `McpMarker.MarkerPath` will not classify this fake home as user-scope. The fake config is under `UserProfile`, but `IsInsideRepo(dir)` walks up to the profile directory and returns true when `~/.git` exists (a common dotfiles setup), so `JsonMcpConfigWriter.Register` writes the marker to the real `~/.kcap/mcp-markers/...` instead of the sidecar under this temp tree. The install test then leaves that central marker behind because it never unregisters. Please use an injected marker path for these tests, or otherwise create the fake home in a location that is under UserProfile but not under any ancestor with `.git`.

realtonyyoung: Fixed in `e789ce5`. You're right — profile-rooting doesn't help when `~` itself is a git repo (`IsInsideRepo` walks up to it → non-user-scope → central marker). Rather than keep chasing the classification, `TempDir.Dispose` now explicitly clears the marker for its cursor config: `McpMarker.Clear` resolves the identical path the production `plugin --cursor` path used to write it (sidecar or central), so the marker can't persist past the test on any OS or repo layout. Kept the UserProfile root so the common (non-repo) case stays a contained sidecar and never touches the real `~/.kcap/mcp-markers` at all. Verified on a non-repo home: zero central markers added, no leftover home dirs.

realtonyyoung:
NO FINDINGS

realtonyyoung:
@alexeyzimarev — ready for your review (you're already the requested reviewer). The automated review cycle is complete on `e789ce5`.

**What it does:** first per-harness consumer of the merged MCP config-writer foundation — wires kcap MCP auto-registration into the Cursor install path so `kcap setup` and `kcap plugin install/remove --cursor` write/remove the 4 kcap servers in `~/.cursor/mcp.json` (mirrors the Codex MCP wiring), with a `--skip-cursor-mcp` opt-out.

**Review status:** Qodo findings resolved; a full Codex review returned NO FINDINGS; all review threads resolved; CI green (AOT + ubuntu/windows build+test).

**Notable fixes made during review:**
- `JsonMcpConfigWriter.Unregister` clears the ownership marker on any non-`Failed` outcome but RETAINS it on `Failed`, so a retry can still identify the kcap-owned entries.
- `uninstall` no longer over-clears the Cursor MCP marker (cleanup rides on `plugin remove --cursor`).
- Test isolation hardened so the ownership marker never leaks into the real `~/.kcap/mcp-markers` (cleared in test teardown regardless of OS/home layout).

Thanks!

realtonyyoung:
NO FINDINGS

PR #309, opened 2026-07-10
[AI-1228] Auto-register kcap MCP servers for Kiro CLI

realtonyyoung:
Extends the MCP auto-config epic (AI-1224) to **AWS Kiro CLI** (AI-1228). Based on `main`.

## What
- **MCP** — registers the 4 kcap servers in Kiro's user-level `~/.kiro/settings/mcp.json` (Standard `mcpServers` shape, marker `"kiro"`). **Independent of the agent clone / default-agent flip** (per AI-1228): a plain JSON merge, so it still applies even when `kiro-cli` isn't on PATH to clone. Non-destructive + idempotent — preserves user servers and their `disabled`/`autoApprove` fields. Opt out `--skip-kiro-mcp`.
- **No trust** — kcap leaves `autoApprove` unset (Kiro's per-server `autoApprove` is unreliable); auto-approve stays the user's choice.
- **No steering** — AI-1228 is MCP-only; Kiro's steering is workspace-scoped, out of scope here.
- `remove --kiro` unregisters the servers independent of the agent restore.

## Design notes
- `InstallKiro` restructured to mirror the merged Copilot/OpenCode pattern: refresh no-ops if never opted in; `hooksCurrent` is `File.Exists`-guarded (a deleted `kcap.json` with a current marker is recreated); and — the Kiro-specific bit — the MCP file is registered **regardless of the clone outcome** (warn-and-continue if the `kiro-cli` clone fails).
- Setup flow: `HandleKiroHooks` gains an `out selected` signal (set once Kiro is opted-in with kcap on PATH); `HandleKiroMcp` is gated on that, so MCP registers even if the clone fails — matching the plugin path.

## Verification (dev binary, headless, temp HOME)
Fresh `install --kiro` with `kiro-cli` present but no default agent to clone → the clone failed **and MCP still registered** (independence); a pre-existing user server with `autoApprove`/`disabled` was fully preserved; kcap-* added in Standard shape (`command:"kcap"` + args, no `type`/`trust`).

## Tests
`KiroPathsTests` (+1), `PluginCommandKiroTests` (+4: shape + preserve `autoApprove`, skip flag, creates `settings/`, remove), `CodingAgentsStepTests` (+5: registered/gated/skipped/failure + **registers even when the agent clone fails**). Full unit suite **2698/2698**.

Linear: AI-1228


src/Capacitor.Cli/Commands/CodingAgentsStep.cs
```diff
@@ -199,6 +211,39 @@ Action<string>     writeLine
         return true;
     }
 
+    /// <summary>
+    /// Registers the kcap MCP servers in Kiro's <c>~/.kiro/settings/mcp.json</c> via
+    /// <see cref="Installers.RegisterKiroMcp"/>. Gated on Kiro being opted-in with kcap on PATH
+    /// (<paramref name="kiroSelected"/>) — independent of the agent-clone outcome, since the MCP
+    /// file is a plain JSON merge — and on <see cref="Options.SkipKiroMcp"/>. Non-destructive.
+    /// </summary>
```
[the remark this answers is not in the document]
realtonyyoung: Trimmed the flagged multi-line comments to concise forms — KiroPaths.SettingsMcpJson, HandleKiroMcp's doc + the `selected` note, InstallKiro's hooksCurrent comment, and the test-class header. Fixed in 043b927.

src/Capacitor.Cli/Commands/CodingAgentsStep.cs
```diff
@@ -185,10 +193,14 @@ Action<string>     writeLine
...
 
+        // Kiro is opted-in and kcap is on PATH: the MCP file (a separate settings/mcp.json) should be
+        // registered even if the agent clone below fails (it needs kiro-cli). Signal that to the MCP step.
+        selected = true;
+
         var ok = installers.InstallKiroHooks?.Invoke(paths.KiroHooksPath) ?? false;
 
         if (!ok) {
-            writeLine("  [yellow]⚠[/] Could not write Kiro agent hooks file.");
+            writeLine("  [yellow]⚠[/] Could not write Kiro agent hooks file (kiro-cli needed to clone your default agent).");
```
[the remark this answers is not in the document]
realtonyyoung: Good catch — generalized the message. It no longer asserts kiro-cli is THE cause; it now names kiro-cli as one possible cause alongside the clone / default-agent-flip failures that InstallKiroHooks can also return false on. Fixed in 043b927.

realtonyyoung (commented):
Found 2 issues in the Kiro MCP wiring. I did not run tests or builds, per request.

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -1215,50 +1215,95 @@ public static bool RemoveCopilotHooks(string hooksPath) {
...
-                  + "Re-install kcap via npm: npm install -g @kurrent/kcap"
-                );
-
-                return 1;
-        }
-
-        if (!InstallKiroHooks(agentPath)) {
-            if (refreshOnly) return 0;
+        // Refresh-only mode never touches a machine that never opted in.
+        if (refreshOnly && !KiroHooksInstaller.IsInstalled(agentPath)) return 0;
```
realtonyyoung: [P2] This refresh gate still treats Kiro as installed only when the hook/agent marker exists. With the new independent MCP path, users can have a real Kiro MCP install without hooks: `kcap setup --skip-kiro-hooks` now registers MCP-only, and a fresh `plugin install --kiro` can register MCP even when the agent clone fails because `kiro-cli` is missing. In both cases the `McpMarker("kiro")` state exists, but a future `--if-installed` refresh returns here and never heals/updates `~/.kiro/settings/mcp.json`. Please include the Kiro MCP ownership marker in the refresh gate, or let the MCP refresh run independently before returning for the no-hooks case.

src/Capacitor.Cli/Commands/CodingAgentsStep.cs
```diff
@@ -174,33 +182,43 @@ Action<string>     writeLine
 
         writeLine("  [green]✓[/] Kiro CLI detected");
 
+        // kcap must be on PATH for BOTH the agent hooks (`kcap hook`) and the MCP servers (`kcap mcp`) —
+        // same precheck as the Cursor/Copilot branches. No kcap → neither hooks nor MCP.
+        if (!installers.CapacitorOnPath()) {
```
realtonyyoung, line 229: [P3] This PATH precheck runs before the `--skip-kiro-hooks` branch, so `kcap setup --skip-kiro-hooks --skip-kiro-mcp` can still warn that Kiro was skipped because `kcap` is not on PATH and never reaches the accurate “skipped by flags (hooks + MCP)” message. Since both flags mean no Kiro hook or MCP command will be written, the both-skip case should short-circuit before requiring `CapacitorOnPath()`.

src/Capacitor.Cli/Commands/PluginCommand.cs
realtonyyoung: Fixed in 32bc4e9 — the `--if-installed` gate now also checks the kcap MCP ownership marker (`McpMarker("kiro").Owned`). An MCP-only install (no agent) heals **just** the mcp.json on refresh and returns before the agent-clone path, so we never install the hooks the user opted out of. Regression tests cover the MCP-only heal and the nothing-installed no-op.

src/Capacitor.Cli/Commands/CodingAgentsStep.cs
realtonyyoung: Fixed in 32bc4e9 — when both `--skip-kiro-hooks` and `--skip-kiro-mcp` are set, `HandleKiroHooks` now short-circuits with the accurate "skipped by flags (hooks + MCP)" message **before** the `CapacitorOnPath()` precheck (nothing is written, so PATH is moot). Added a regression test asserting the PATH check isn't reached and no "not on PATH" warning is emitted.

realtonyyoung:
NO FINDINGS

PR #314, opened 2026-07-10
[AI-1239] Auto-register kcap MCP servers for Pi (MCP-bridge extension)

realtonyyoung:
## Summary

Pi (`badlogic/pi-mono`) intentionally ships **no built-in MCP**, so there is no JSON `mcpServers` config to write the way the other harnesses have. This adds a kcap-owned **MCP-bridge extension** so Pi still gets the kcap MCP tools.

`kcap plugin install --pi` now installs, alongside the existing live-ingest extension:

- **`~/.pi/agent/extensions/kcap-mcp.ts`** — at load it spawns each `kcap mcp <name>` server (review, sessions, flows, memory) as a persistent stdio subprocess, performs the MCP handshake (`initialize` → `notifications/initialized` → `tools/list`), and re-registers every advertised tool as a native Pi tool named `kcap_<server>_<tool>` (all underscores — dodges the mixed `-`/`_` naming that trips weaker models, AI-1301) whose `execute()` proxies `tools/call` back to the subprocess.
- **`~/.pi/agent/AGENTS.md`** — kcap's marker-delimited steering block (reused across harnesses) nudging Pi toward the kcap tools.

## The bridge extension

Dependency-free (only `node:child_process`) and fail-safe:

- **Per-server isolation** — each server is spawned/handshaken/registered independently; one bad server never blocks the others or the pi session.
- **Bounded handshake** — 10s per server; a hung server is killed and skipped.
- **Pending-call rejection** — in-flight `tools/call`s are rejected (not hung) when a subprocess dies.
- **Teardown** — `session_shutdown` (fires on session switch *and* process exit) kills all subprocesses; a sync `process.once("exit")` backstop covers abrupt exits, and each instance drops its own exit listener on shutdown so per-session-switch reloads don't accumulate listeners. **No orphaned `kcap mcp *` processes survive a normal pi exit.**
- **Async factory** — pi awaits it before `session_start`, so the tools are registered before turn 1. Schema passthrough: the MCP `inputSchema` is passed verbatim as `parameters` (pi forwards it to the provider as JSON Schema; confirmed no runtime typebox validation on tool params).

## Wiring (mirrors the other harnesses)

- **Core:** `PiPaths.KcapMcpExtension` / `KcapMcpExtensionMarker` / `AgentsMd`; `PiMcpExtensionInstaller` (embedded const + distinct `.kcap-mcp-extension-version` marker).
- **PluginEnvironment:** `PiKcapMcpExtension`, `PiAgentsMd`.
- **PluginCommand:** `InstallPi` also writes `kcap-mcp.ts` (`--skip-pi-mcp`) + the AGENTS.md block (`--skip-pi-instructions`); `--if-installed` heals both when the ingest extension is present; `RemovePi` removes all three.
- **Setup flow:** `HandlePiExtension` gains an `out selected` signal; `HandlePiMcp` + `HandlePiInstructions` gate on it (so they install even if the ingest write fails); `SetupCommand` parses the two new flags.
- **Docs:** `help-plugin.txt`, `help-setup.txt`.

## Testing

Full unit suite green (the one intermittent failure is a pre-existing `LocalPermissionBridge` port-bind flake in an unrelated daemon test — passes in isolation). New/extended coverage:

- `PiMcpExtensionInstallerTests` — write/remove, marker, distinct-from-ingest marker, and bridge-content assertions (async factory, all 4 servers, `notifications/initialized`, `kcap_<server>_<tool>` naming, `session_shutdown` + `process.once("exit")` teardown).
- `PiPathsTests` — new path members + env-var derivation.
- `PluginCommandPiTests` — MCP + AGENTS.md install, `--skip-pi-mcp` / `--skip-pi-instructions`, AGENTS.md content preservation, remove strips all three.
- `CodingAgentsStepTests` — `HandlePiMcp` / `HandlePiInstructions` wiring, `selected`-independence (install even when the ingest write fails), skip flags, failure warnings.

Design: `docs/superpowers/specs/2026-07-10-pi-mcp-bridge-extension-design.md` (in the kcap-server repo).

Part of the AI-1224 MCP auto-config epic — Pi is the last harness.

### DIFF
--- a/src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
+++ b/src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
@@ -1,0 +1,356 @@
+namespace Capacitor.Cli.Core.Pi;
+
+/// <summary>
+/// Installs / removes kcap's <b>MCP-bridge</b> extension for Pi. Pi ships no
+/// built-in MCP (<c>badlogic/pi-mono</c> deliberately omits it), so there is no
+/// JSON <c>mcpServers</c> config to write the way the other harnesses have.
+/// Instead kcap ships a second TypeScript extension
+/// (<c>~/.pi/agent/extensions/kcap-mcp.ts</c>) that Pi auto-discovers: at load it
+/// spawns each <c>kcap mcp &lt;name&gt;</c> stdio server as a persistent
+/// subprocess, performs the MCP handshake, and re-registers every advertised tool
+/// as a native Pi tool (<c>kcap_&lt;server&gt;_&lt;tool&gt;</c>) whose
+/// <c>execute()</c> proxies <c>tools/call</c> back to that subprocess.
+///
+/// <para>This is a sibling of <see cref="PiExtensionInstaller"/> (the live-ingest
+/// extension) — same embedded-const + version-marker shape, but a distinct file
+/// and marker so the two install/refresh/remove independently. <see cref="ExtensionContent"/>
+/// is the single source of truth (embedding it as a const keeps NativeAOT happy —
+/// no manifest resource reflection).</para>
+/// </summary>
+public static class PiMcpExtensionInstaller {
+    public const string MarkerFileName = ".kcap-mcp-extension-version";
+
+    /// <summary>
+    /// The kcap Pi MCP-bridge extension. Untyped (<c>pi: any</c>) so it carries no
+    /// runtime dependency on the <c>@earendil-works/pi-coding-agent</c> types, and
+    /// fail-safe so a kcap/server hiccup never disrupts the pi session: each server
+    /// is spawned/handshaken independently and failures are logged and skipped.
+    /// </summary>
+    public const string ExtensionContent =
+        """
+        // kcap-mcp.ts — Kurrent Capacitor MCP-bridge extension for Pi (badlogic/pi-mono).
+        //
+        // Installed by `kcap plugin install --pi` into ~/.pi/agent/extensions/kcap-mcp.ts.
+        // Pi has no built-in MCP, so this extension bridges the kcap stdio MCP servers
+        // (`kcap mcp <name>`) into Pi as native tools: at load it spawns each server as a
+        // persistent subprocess, performs the MCP handshake, and re-registers every tool
+        // the server advertises as a Pi tool named `kcap_<server>_<tool>` whose execute()
+        // proxies `tools/call` back to that subprocess.
+        //
+        // Dependency-free (only the node:child_process builtin; no pi SDK import) and
+        // fail-safe: each server is spawned/handshaken/registered independently so one
+        // bad server never blocks the others or the pi session.
+
+        import { spawn } from "node:child_process";
+
+        // The kcap MCP servers to bridge, in `kcap mcp <name>` form. `flows` starts a
+        // *paid* hosted reviewer only on explicit invocation; its tool descriptions
+        // (surfaced verbatim from the server) name that cost so the model/user see it.
+        const KCAP_MCP_SERVERS = ["review", "sessions", "flows", "memory"];
+
+        // Bound the initialize + tools/list handshake per server so a hung server can
+        // never stall pi's startup indefinitely.
+        const HANDSHAKE_TIMEOUT_MS = 10000;
+
+        // A minimal line-delimited JSON-RPC 2.0 client over one `kcap mcp <name>`
+        // subprocess. No external MCP SDK.
+        class McpStdioClient {
+          server: string;
+          child: any = null;
+          nextId = 1;
+          pending = new Map<number, { resolve: (v: any) => void; reject: (e: any) => void }>();
+          buffer = "";
+          closed = false;
+
+          constructor(server: string) {
+            this.server = server;
+          }
+
+          // spawn + MCP handshake (initialize -> notifications/initialized).
+          async start(): Promise<void> {
+            const child = spawn("kcap", ["mcp", this.server], { stdio: ["pipe", "pipe", "ignore"] });
+            this.child = child;
+            child.on("exit", () => this.fail(new Error("kcap mcp " + this.server + " exited")));
+            child.on("error", (e: any) => this.fail(e instanceof Error ? e : new Error(String(e))));
+            child.stdout.setEncoding("utf8");
+            child.stdout.on("data", (chunk: string) => this.onData(chunk));
+
+            await this.request("initialize", {
+              protocolVersion: "2024-11-05",
+              capabilities: {},
+              clientInfo: { name: "kcap-pi-bridge", version: "1" },
+            });
+            // Per the MCP spec the client must send `notifications/initialized`
+            // after the initialize response and before any further requests.
+            this.notify("notifications/initialized", {});
+          }
+
+          onData(chunk: string) {
+            this.buffer += chunk;
+            let nl: number;
+            while ((nl = this.buffer.indexOf("\n")) >= 0) {
+              const line = this.buffer.slice(0, nl).trim();
+              this.buffer = this.buffer.slice(nl + 1);
+              if (!line) continue;
+              let msg: any;
+              try {
+                msg = JSON.parse(line);
+              } catch {
+                continue; // ignore non-JSON lines (defensive)
+              }
+              const id = msg && msg.id;
+              if (typeof id === "number" && this.pending.has(id)) {
+                const p = this.pending.get(id)!;
+                this.pending.delete(id);
+                if (msg.error) p.reject(new Error((msg.error && msg.error.message) || "MCP error"));
+                else p.resolve(msg.result);
+              }
+            }
+          }
+
+          // Reject every in-flight call when the subprocess dies so execute() never
+          // hangs; mark closed so later calls fail fast too.
+          fail(err: Error) {
+            if (this.closed) return;
+            this.closed = true;
+            for (const p of this.pending.values()) p.reject(err);
+            this.pending.clear();
+          }
+
+          send(obj: any) {
+            if (this.closed || !this.child || !this.child.stdin || !this.child.stdin.writable) {
+              throw new Error("kcap mcp " + this.server + " not available");
+            }
+            this.child.stdin.write(JSON.stringify(obj) + "\n");
+          }
+
+          notify(method: string, params: any) {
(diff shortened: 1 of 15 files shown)


src/Capacitor.Cli.Core/Resources/help-plugin.txt
```diff
@@ -81,6 +85,17 @@ Options:
...
+                      (~/.pi/agent/extensions/kcap-mcp.ts). By default `install
+                      --pi` writes it so Pi exposes the 4 kcap MCP servers as
+                      native tools (Pi has no JSON mcpServers config, so the
+                      bridge extension is the only way to surface them).
+  --skip-pi-instructions
+                      With --pi: skip installing kcap's steering-instructions
+                      block into Pi's global ~/.pi/agent/AGENTS.md. By default
+                      `install --pi` writes a small kcap-owned, marker-delimited
+                      block (non-destructive — preserves your own instructions)
+                      that nudges Pi to use the kcap tools for why/history questions.
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in 3937538 — added a dedicated "Pi extension" README section documenting the MCP-bridge extension + AGENTS.md steering + `--skip-pi-mcp`/`--skip-pi-instructions`, and updated the Pi callout and the uninstall description.

src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
```diff
@@ -0,0 +1,356 @@
...
+/// spawns each <c>kcap mcp &lt;name&gt;</c> stdio server as a persistent
+/// subprocess, performs the MCP handshake, and re-registers every advertised tool
+/// as a native Pi tool (<c>kcap_&lt;server&gt;_&lt;tool&gt;</c>) whose
+/// <c>execute()</c> proxies <c>tools/call</c> back to that subprocess.
+///
+/// <para>This is a sibling of <see cref="PiExtensionInstaller"/> (the live-ingest
+/// extension) — same embedded-const + version-marker shape, but a distinct file
+/// and marker so the two install/refresh/remove independently. <see cref="ExtensionContent"/>
+/// is the single source of truth (embedding it as a const keeps NativeAOT happy —
+/// no manifest resource reflection).</para>
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in 3937538 — trimmed the `PiMcpExtensionInstaller` doc comment and the embedded `kcap-mcp.ts` banner to concise intent-level notes.

src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
```diff
@@ -0,0 +1,356 @@
...
+          }
+
+          async listTools(): Promise<any[]> {
+            const res = await this.request("tools/list", {}, HANDSHAKE_TIMEOUT_MS);
+            return res && Array.isArray(res.tools) ? res.tools : [];
+          }
+
+          callTool(name: string, args: any): Promise<any> {
+            return this.request("tools/call", { name, arguments: args || {} });
+          }
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in 3937538 — `callTool()` now passes a generous `TOOL_CALL_TIMEOUT_MS` (20 min, well above the flows server's own round timeouts) to `request()`; on timeout the error surfaces to the model as tool-result content via `execute()`'s catch.

src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
```diff
@@ -0,0 +1,356 @@
...
+                    await client.start();
+                    return client.listTools();
+                  })(),
+                  HANDSHAKE_TIMEOUT_MS,
+                  "kcap mcp " + server + " handshake",
+                );
+              } catch {
+                client.stop(); // skip this server; leave the others registered
+                return;
+              }
```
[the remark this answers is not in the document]
realtonyyoung: Fixed in 3937538 — spawn now pipes stderr and forwards each line to `console.error` with a `[kcap-mcp <server>]` prefix, and the handshake-failure catch logs the server name + error before skipping (so a skipped server is explainable).

realtonyyoung (commented):
Reviewed the Pi MCP bridge and the C# install/setup wiring without running tests or builds, per request. I found a few correctness issues around live Pi failure handling and repo conventions: registration-time exceptions can abort the bridge after subprocesses are spawned, stdin pipe errors can crash instead of surfacing as tool-result content, MCP removal can silently succeed after delete failures, and new comments include Linear issue IDs.

src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
```diff
@@ -0,0 +1,356 @@
...
+            this.closed = true;
+            for (const p of this.pending.values()) p.reject(err);
+            this.pending.clear();
+          }
+
+          send(obj: any) {
+            if (this.closed || !this.child || !this.child.stdin || !this.child.stdin.writable) {
+              throw new Error("kcap mcp " + this.server + " not available");
+            }
+            this.child.stdin.write(JSON.stringify(obj) + "\n");
```
realtonyyoung, line 166: `send()` only checks `stdin.writable` and then calls `write()` without handling the pipe's own `error` path. If the MCP subprocess has already exited or closed stdin but the `exit` handler has not marked `closed` yet, this can surface as an unhandled `EPIPE`/stream error in the Pi extension process instead of rejecting this request and returning the tool-result content path. Please attach an error handler/callback for `child.stdin` and route it through `fail()`/request rejection so calls to a just-dead server fail fast without taking down Pi.

src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
```diff
@@ -0,0 +1,356 @@
...
+
+              clients.push(client);
+
+              for (const tool of tools) {
+                const mcpName = tool && tool.name;
+                if (!mcpName) continue;
+                const toolName = sanitizeToolName("kcap_" + server + "_" + mcpName);
+                if (!toolName) continue;
+                const description = String((tool && tool.description) || ("kcap " + server + " " + mcpName));
+                pi.registerTool({
```
realtonyyoung: This `registerTool` call is outside the per-server `try` block, and the session/process teardown hooks are not installed until after the whole `Promise.all` completes. If Pi rejects one advertised tool (duplicate sanitized name, unsupported schema, or any API error), the factory rejects, the Pi session can fail to load the extension, and any clients already pushed into `clients` are left without the shutdown/exit cleanup. To preserve the fail-safe/per-server isolation contract, wrap registration failures per server/tool and stop that client, or install teardown before registration and ensure one bad tool does not reject the whole factory.

src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
```diff
@@ -0,0 +1,356 @@
...
+        }
+    }
+
+    /// <summary>Removes kcap-mcp.ts + marker. Returns true if kcap-mcp.ts existed.</summary>
+    public static bool Remove(string extensionPath) {
+        var existed = File.Exists(extensionPath);
+        try {
+            if (existed) File.Delete(extensionPath);
+            DeleteMarker(extensionPath);
+        } catch {
```
realtonyyoung: `Remove()` collapses delete failures into the same `false` value used for "file was absent". `PluginCommand.RemovePi` then silently ignores a failed MCP-extension removal and exits 0, so a permission/IO failure can leave `kcap-mcp.ts` or its marker installed with no warning. Please let the exception propagate or return an explicit failure outcome so `plugin remove --pi` can report/remove failures separately from "not installed".

test/Capacitor.Cli.Tests.Unit/PiMcpExtensionInstallerTests.cs
```diff
@@ -0,0 +1,105 @@
+using Capacitor.Cli.Core;
+using Capacitor.Cli.Core.Pi;
+
+namespace Capacitor.Cli.Tests.Unit;
+
+/// <summary>
+/// Covers the Pi MCP-bridge extension installer (AI-1239): kcap-mcp.ts write/remove
```
realtonyyoung: The prompt notes that this repo forbids Linear issue IDs in code comments, but this PR adds `AI-1239` here and in the new Pi-related test section headers. Please remove the Linear IDs from comments while keeping the descriptive text.

src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
realtonyyoung: Fixed in 6f61698 — added `child.stdin.on("error", …)` in start() that routes EPIPE/stream errors through `fail()` (rejecting pending calls → Pi's tool-error path) instead of letting them crash pi.

src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
realtonyyoung: Fixed in 6f61698 — `pi.registerTool` is wrapped in try/catch (logged + skipped on failure) and `registered.add` happens only after a successful call, so a bad tool never rejects the Promise.all. The process-exit reaper is armed at first spawn via `ensureExitHook()` (not after startBridge), so children are killed even if the process dies mid-startup; session_shutdown can't fire before the awaited factory completes.

src/Capacitor.Cli.Core/Pi/PiMcpExtensionInstaller.cs
realtonyyoung: Fixed in 6f61698 — `Remove()` now lets file-delete IO/permission failures propagate (marker deletion stays best-effort), so `RemovePi`'s try/catch warns and returns non-zero instead of silently succeeding.

test/Capacitor.Cli.Tests.Unit/PiMcpExtensionInstallerTests.cs
realtonyyoung: Fixed in 6f61698 — removed the `AI-1239` references from the Pi test comments (PiMcpExtensionInstallerTests, PluginCommandPiTests, CodingAgentsStepTests).

PR #348, opened 2026-07-22
Add safe Cursor borrowed review snapshots

realtonyyoung:
## What changed

- Advertise structured borrowed-review capability metadata for every unattended vendor.
- Give Cursor current checkout context through a daemon-owned worktree snapshot instead of the live checkout.
- Mirror tracked, dirty, and non-ignored untracked files at launch, refresh the snapshot before follow-up rounds, and remove it during normal agent cleanup.
- Require every borrowed review launch to pass a runtime capability gate; Cursor's ACP factory also rejects any raw borrowed-cwd launch that bypasses snapshot materialization.
- Add gated real-CLI certification tests for Cursor and Claude, and remove Claude's obsolete `MultiEdit` deny token.

## Why

Cursor's zero-interaction modes can still write files, so direct borrowed-checkout execution cannot meet the unattended reviewer safety boundary. A disposable owned snapshot preserves the useful dirty/untracked context while ensuring mutations cannot reach the caller's checkout.

Claude remains fail-closed and does not advertise borrowed support. Its prospective read-only argv is covered by unit tests, but the live certification run is currently blocked by the org monthly spend limit.

## Validation

- `dotnet build test/Capacitor.Cli.Tests.Unit/Capacitor.Cli.Tests.Unit.csproj --no-restore --disable-build-servers -m:1 /nodeReuse:false`
- 56 focused descriptor, daemon-capability, launcher, ACP-boundary, and borrowed-orchestrator tests passed.
- Real Cursor ACP certification passed: zero interaction requests, result MCP completed, source checkout unchanged, explicit mutation confined to the owned snapshot.
- Claude probe confirmed the corrected deny-list syntax is accepted, then stopped at the org monthly spend limit before model/tool execution.


realtonyyoung:
Claude review flow: **NO FINDINGS** after the final portability fixes. Flow `15e0386d314b4f36a722a236bfee2876` completed clean and was closed.

PR #360, opened 2026-07-24
[AI-1439] Await import background tasks; de-flake retry timing tests

realtonyyoung:
Fixes two independent sources of kcap-cli unit-test flakiness surfaced while working [#335](https://github.com/kurrent-io/kcap-cli/pull/335). Linear: [AI-1439](https://linear.app/kurrent/issue/AI-1439).

## Part 1 — production bug: `kcap import` never awaited background title/summary work

`ImportCommand.HandleImport` computed the background-phase guard inverted:

```csharp
var ranBackground = backgroundTasks.IsEmpty;   // missing a '!'
```

The one boolean gates (a) the `if (…)` block that `await Task.WhenAll(backgroundTasks)`, and (b) the two Done-summary guards that print the Titles/Summaries rows. Because it was inverted, an import that **has** background work (title generation, or summaries under `--generate-summaries`):

- **skipped** the await block → the method returned while a title task could still be running → the shared `using var concurrencyLimit = new SemaphoreSlim(3)` was disposed → the straggler's `finally { concurrencyLimit.Release(); }` threw **`ObjectDisposedException`** (the intermittent unit failure — rare only because tasks usually finish before the method returns);
- dropped the Titles/Summaries rows from the Done summary.

Introduced in `fa2f013` (2026-04-26); #335's new tests merely exercised `HandleImport` enough to surface it.

**Fix:** `!backgroundTasks.IsEmpty`, renamed to `hasBackgroundWork` and extracted into a named `HadBackgroundWork` seam so the decision is unit-testable and can't silently re-invert — mirroring the existing `ComputePerSourceFinalCounts` "extract-and-test-the-helper-directly" convention in this file. The two `RanBackground: false` early-exit constructions stay correct.

This changes `kcap import`'s observable Done-summary output (the Titles/Summaries rows now appear when there was background work) — intended.

## Part 2 — parallel-load timing flakes

- **`SendWithRetry` timing tests** (the genuine remaining flaky-*failure* vector): widen the wall-clock margins while preserving each invariant.
  - `…_after_budget_exhausted`: `totalTimeout` 1200 ms → 5000 ms, so a scheduler stall during attempt 1 can't consume the whole budget before attempt 2 begins (`attempts > 1` is itself condition-based).
  - `…_enforces_total_timeout_…`: elapsed upper bound 2000 ms → 5000 ms (still ≪ the 30 s per-attempt cap it guards against).
  - These verify timeout enforcement, so they're inherently wall-clock; generous bounds is the correct hardening.
- **Port contention** (`LocalPermissionBridgeTests`, `AgentOrchestratorReviewerTokenTests`, `AgentOrchestratorVendorTests`): already fixed at the **production** level — `LocalPermissionBridge` binds an OS-assigned ephemeral loopback port behind a process-global claim set + 15-attempt transient-bind retry (#340), and the bridge tests carry `[NotInParallel]` grouping. No test change needed.
- **`AcpTranscriptAggregationTests` / `AgentOrchestratorVendorTests` `Task.Delay(200/100)`**: these are "settle, then assert a **negative**" waits — a short wait risks a false *pass* (missed regression), not a flaky *fail* — so they're not the parallel-failure vector and are left as-is.

## Tests

- New `ImportBackgroundWorkTests`: `HadBackgroundWork` empty → false, non-empty → true.
- `ImportDisplayGridTests`: `WriteDoneGrid` renders the Titles/Summaries rows iff `RanBackground`.
- Widened `HttpClientExtensionsRetryTests` margins.

Verified locally: full Import namespace (234), `SetupCommandTests`/`ImportVisibilityTests` (26, drive `HandleImport`), and `HttpClientExtensionsRetryTests` (5) all pass; build clean; `check-linear-ids.sh` passes.

## Out of scope / follow-ups
- Deeper process-global refactor of import background-task lifetime → AI-741.
- AI-1438 is a textual duplicate of AI-1439 — close as dup.

### DIFF
--- a/src/Capacitor.Cli/Commands/ImportCommand.cs
+++ b/src/Capacitor.Cli/Commands/ImportCommand.cs
@@ -467,11 +467,9 @@ static class ImportCommand {
         );
 
     /// <summary>
-    /// True when the import scheduled background title/summary work that the Done phase must
-    /// await (and whose completion the Titles/Summaries rows report). Extracted as a named seam
-    /// so the await-guard decision is unit-testable and can't silently re-invert: the whole class
-    /// of "background tasks never awaited → SemaphoreSlim disposed mid-flight → ObjectDisposedException"
-    /// bug was a single missing negation here.
+    /// True when the import scheduled background title/summary work the Done phase must await
+    /// (and whose results the Titles/Summaries rows report). A named seam so the await-guard
+    /// decision is unit-testable and can't silently re-invert.
     /// </summary>
     internal static bool HadBackgroundWork(ConcurrentBag<Task> backgroundTasks) => !backgroundTasks.IsEmpty;
 
--- a/src/Capacitor.Cli/Commands/ImportCommand.cs
+++ b/src/Capacitor.Cli/Commands/ImportCommand.cs
@@ -467,6 +467,15 @@ static class ImportCommand {
         );
 
     /// <summary>
+    /// True when the import scheduled background title/summary work that the Done phase must
+    /// await (and whose completion the Titles/Summaries rows report). Extracted as a named seam
+    /// so the await-guard decision is unit-testable and can't silently re-invert: the whole class
+    /// of "background tasks never awaited → SemaphoreSlim disposed mid-flight → ObjectDisposedException"
+    /// bug was a single missing negation here.
+    /// </summary>
+    internal static bool HadBackgroundWork(ConcurrentBag<Task> backgroundTasks) => !backgroundTasks.IsEmpty;
+
+    /// <summary>
     /// Compute the per-source Done-grid row for a single vendor.
     /// </summary>
     /// <param name="classifications">All classifications for this vendor (already filtered).</param>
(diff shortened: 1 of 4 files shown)


realtonyyoung:
Addressed both Qodo findings in fd4d109:

1. **Verbose comments** — trimmed the `HadBackgroundWork` XML doc and the `ImportBackgroundWork`/retry/display-grid test comments to concise intent (the bug narrative lives in the PR/Linear, not the code).
2. **Slow retry test** — the budget-exhausted test no longer uses a 5s `totalTimeout` (which the never-completing send ran to the full budget). It now uses a small per-attempt cap (50 ms) under a modest 1500 ms total, so retries accrue fast — `attempts > 1` stays robust and the test finishes in ~1.5 s (test 2's elapsed-bound widening, which costs no runtime, is kept).

Also for the record: the first ubuntu run failed on `HandleStopAgent_unknown_id_reaps_by_pid_record_where_env_is_readable` — an unrelated native `UnixSpawnerThread` PID-reaping test that FailFasts the process (windows passed the identical code). Not touched by this PR; a separate parallel-load flake candidate.

PR #364, opened 2026-07-24
[AI-910][AI-1516] Fix two kcap-cli parallel-load flakes

realtonyyoung:
Fixes two kcap-cli test flakes. Linear: [AI-910](https://linear.app/kurrent/issue/AI-910), [AI-1516](https://linear.app/kurrent/issue/AI-1516).

## AI-910 — `ClaudeHookStdoutTests` stdout-capture contamination

The three tests captured hook stdout by redirecting the **process-global `Console.Out`** to a `StringWriter`. A concurrently-running test whose SUT writes to `Console.Out` (e.g. `CodexHookCommand` emitting `{"continue":true}`) leaked a second JSON object into the capture → `JsonReaderException: '{' is invalid after a single JSON value`.

**Fix (issue's preferred option 1):** thread an optional `TextWriter? stdout` through `ClaudeHookCommand.Handle → HandleWithDeps → HandleCore` (default `Console.Out`). The SessionStart `hookSpecificOutput` envelope and the auth-lapsed `systemMessage` nudge now write to that writer. Each test passes its own `StringWriter` — no `Console.SetOut`, so nothing another test writes to `Console` can contaminate it. This is the durable fix for the class (a keyed `[NotInParallel]` couldn't prevent it; the writer does, regardless of scheduling).

`[NotInParallel]` is **kept**: the injection ends the contamination flake, but these tests still exercise process-global state (`KCAP_CONFIG_DIR` config + the in-process auth-provider cache) and real hook HTTP timing. (Verified that fully parallelizing them exposes an *unrelated* empty-envelope flake under load — out of scope.)

## AI-1516 — `HandleStopAgent` reap-by-pid-record confirmation race

Surfaced on PR #360's first ubuntu run: `HandleStopAgent_unknown_id_reaps_by_pid_record_where_env_is_readable` intermittently failed with the PID record still present after the process had exited.

**Root cause:** `ProcessReaper.KillConfirmAsync`'s `Classify` reads `IsAlive` (`kill(0)`) then the start token (`/proc/{pid}/stat`). A SIGTERM'd, proven-ours process in its death transition reads `alive=true` but its token momentarily unreadable → `Ambiguous`. The grace loop returned false on that transient `Ambiguous`, so the reaper reported *unconfirmed* even though the process was (about to be) dead → the record was stranded (`ReapByRecordAsync` deletes only on confirmed-gone). The single fixed 250ms post-SIGKILL check had the same fragility under OS-reflection lag.

**Fix:** both confirmation waits (grace + post-SIGKILL) poll via a shared `PollForConfirmedDeathAsync` for a *definitive* `Dead`/`Recycled`, treating a transient `Ambiguous` as **keep polling** (we've already proven ownership and signalled). The pre-signal and pre-SIGKILL **gates** still spare on `Ambiguous` — the safety invariant is unchanged: only a proven-ours leader is ever signalled, and a persistently-ambiguous target is still retained for the next sweep.

**Test correction (see the Linear comment).** The test's macOS branch asserted "always spares (`HasExited == false`)", but tracing showed that's a **race**, not a real spare: on macOS 26.5.2 `Classify = Ours` (identity is readable) so the reaper *does* SIGTERM the dummy — the old code only "passed" by returning on the first transient `Ambiguous` **before the SIGTERM landed**, leaving the process dead-but-record-retained (the exact bug). The macOS branch now asserts the real invariant — **record deleted IFF the process was confirmed gone** — making it a true regression guard (fails on the pre-fix dead-but-retained state, passes on the fix) and robust across macOS versions.

## Verification (local, macOS 26.5.2)
- `ClaudeHookStdoutTests` 3/3; full `Capacitor.Cli.Tests.Integration` **144/144**; `ClaudeHookCommandTests` 35/35; `SpoolOutageRecoveryTests` 2/2.
- Reaper test 4/4; `AgentOrchestratorVendorTests` **84/84**; `OrphanReaperTests` 9/9.
- Build clean; `check-linear-ids.sh` passes.
- The Linux reap+confirm path (the CI flake) is validated by the ubuntu CI leg.

### DIFF
--- a/src/Capacitor.Cli.Daemon/Services/ProcessReaper.cs
+++ b/src/Capacitor.Cli.Daemon/Services/ProcessReaper.cs
@@ -151,15 +151,18 @@ internal static class ProcessReaper {
     /// signal sent, so a transient Ambiguous keeps polling (never spares here); returns false at window
     /// expiry so the caller retains the record for the next sweep.</summary>
     static async Task<bool> PollForConfirmedDeathAsync(int pid, string expectedIdentity, TimeSpan window, CancellationToken ct) {
-        for (var waited = TimeSpan.Zero; waited < window; waited += TimeSpan.FromMilliseconds(250)) {
+        // Classify, then delay-and-reclassify until the window elapses — the final classify happens
+        // AFTER the last delay (when waited >= window), so a death reflected during that last interval
+        // is still confirmed rather than dropped.
+        for (var waited = TimeSpan.Zero; ; waited += TimeSpan.FromMilliseconds(250)) {
             switch (Classify(pid, expectedIdentity)) {
                 case LeaderState.Dead:     return true;
                 case LeaderState.Recycled: return true;
             }
+
+            if (waited >= window) return false;
             await Task.Delay(250, ct);
         }
-
-        return false;
     }
 
     /// <summary>SIGTERM/SIGKILL the target's process group (Unix; falls back to the bare pid when it
--- a/src/Capacitor.Cli.Daemon/Services/ProcessReaper.cs
+++ b/src/Capacitor.Cli.Daemon/Services/ProcessReaper.cs
@@ -114,17 +114,12 @@ internal static class ProcessReaper {
 
             SignalGroup(pid, hard: false); // SIGTERM the group (leader is proven ours → group is our lineage)
 
-            // Grace poll. The leader was proven ours immediately above, so a TRANSIENT Ambiguous here is
-            // the death transition — kill(pid,0) still answers for an instant after the process has begun
-            // exiting while /proc/{pid}/stat has already gone, so the start-token read is momentarily
-            // unreadable — NOT a reason to spare. Keep polling for the definitive Dead/Recycled; only
-            // those confirm within the window (see PollForConfirmedDeathAsync).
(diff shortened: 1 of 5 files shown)


realtonyyoung:
Addressed all three Qodo findings in 898246b:

1. **Windows hard-kill wait (Bug/Perf)** — good catch. On Windows `SignalGroup(hard:true)` is a no-op (the soft signal already `Process.Kill(entireProcessTree:true)` and the grace poll gave it the full window), so the post-SIGKILL confirmation poll is now capped at a single 250ms tick on Windows instead of the 5s `ConfirmAfterKill` window — no added latency for a non-terminating pid. Unix keeps the full poll (where SIGKILL is real and OS-reflection lag is expected).
2. **Verbose `KillConfirmAsync`/`PollForConfirmedDeathAsync` comments** — trimmed to concise intent.
3. **Verbose test rationale comments** — trimmed in both `ClaudeHookStdoutTests` and `StopAgentPidFallbackTests`.

Re-verified locally: `AgentOrchestratorVendorTests` 84/84, the reaper test 3/3, `ClaudeHookStdoutTests` 3/3; `check-linear-ids.sh` ✓.

PR #378, opened 2026-07-28
[AI-1526][AI-1528] Address Qodo findings from #376/#377

realtonyyoung:
Follow-up to #376 and #377, both of which merged with unaddressed Qodo review findings. No behavior change to the settlement protocol itself.

## Finding 1 — liveness read inside the sequenced-command lock (real bug/perf, from #377)

`SequencedCommandProcessor.BuildProcessedAckLocked()` invoked the `readLiveness` delegate while holding `_lock`. In production that delegate is `AgentOrchestrator.ReadLiveness()`, which called `AgentKillQuarantine.Snapshot()` and LINQ'd over the projected list — allocating inside the critical section. #377's proactive terminal ack moved that from "only on a duplicate replay" to **every settled command**, so lock-hold time grew and could delay concurrent `SubmitAsync` / `AckPrefix` callers.

Both halves fixed:

1. `AgentKillQuarantine.IsQuarantined(agentId)` — an allocation-free `ConcurrentDictionary.ContainsKey` membership test. `AgentOrchestrator.ReadLiveness` now uses it instead of `Snapshot().Any(q => q.Id == agentId)`.
2. `_readLiveness` is no longer invoked under `_lock` on **either** ack path:
   - **Proactive (lane) path** — the lock block records the outcome in the cache and advances the watermark; the ack is built and sent after the lock is released.
   - **Duplicate-replay path** — `HandleDuplicateLocked` now only *captures* the cached `CommandOutcome` (a struct copy) into an `out` parameter; `SubmitAsync` builds and sends the ack after releasing the lock. `SubmitAsync`'s body moved into `SubmitLocked` so the lock scope stays a single, obvious block.

`BuildProcessedAckLocked` is renamed `BuildProcessedAck` and its doc now states the "must not be called under `_lock`" contract.

**The deliberate ordering invariant is preserved**: both callers record the outcome in the cache *before* the ack is built, so an ack can never advertise an outcome the cache has not recorded. Only the ack *construction* moved out of the lock, not the cache write.

## Finding 2 — README described the old settlement-retry behavior (from #377)

The README said settlement-layer coded 409s were retried "a few times" and were "normally invisible to the caller." That stopped being true when the flows MCP server moved to a 3-minute elapsed deadline with an explicit coded timeout as a tool error. The section now documents:

- which two codes retry (`flow_settlement_busy`, `reviewer_launch_incarnation_superseded`) and that everything else surfaces on the first attempt;
- the backoff shape (250–500 ms first retry, 5–10 s steady state);
- that the budget is an **elapsed-time** deadline of 3 minutes from the first attempt, including each request's own duration (not an attempt count);
- that exhausting it returns an explicit `isError: true` tool result in the `Error (code): message` shape, carrying the last coded error plus attempt count and elapsed time — with an example — and that it is retryable;
- that the round poll loop has its own separate budget and message.

## Finding 3 — comment verbosity (both PRs, "review recommended")

Trimmed restated mechanics and repeated rationale at the three flagged sites, to roughly the density of the surrounding untouched code:

- `FlowRetryTiming.cs` — `FlowRetryClock` and `SettlementBackoff` class docs condensed. **Kept** the pinned backoff formula: it is the assertable contract the tests pin, not a restatement of the code.
- `SequencedCommandProcessor.cs` — dropped the "Task 13 / Task 15" plan-ticket call-site annotations and the duplicated §3.2 F rationale that appeared in three places; the ordering invariant is now stated once, on `BuildProcessedAck`.
- `AgentOrchestratorBorrowLaunchTests.cs` — dropped the historical "this path had never executed in production" paragraph. **Kept** the two-checkout requirement (a future simplification to one repo would silently make the test vacuous), the decoy-content note, and the non-vacuity guards ("the droppings must genuinely exist first", "re-basing the baseline cannot launder a corruption") — those record a non-obvious why and prevent regressions. Also kept the `ContentBaseline` / `GitState` helper docs explaining why `.git` is excluded and why `SnapshotTree` is not used.

## Tests

New regression coverage in `SequencedCommandProcessorTests`:

- `Liveness_is_never_read_while_the_processor_lock_is_held` — a test seam (`LockHeldByCurrentThreadForTest`, `Monitor.IsEntered`) lets the injected `readLiveness` delegate assert directly that it is not holding the lock. No sleeps, no thread choreography. **Mutation-verified**: reverting either ack path to build under the lock fails this test with `readsUnderLock == 2`.
- `Proactive_ack_is_built_only_after_the_outcome_is_recorded_in_the_cache` — guards the ordering invariant the lock move had to preserve.
- `AgentKillQuarantineTests.IsQuarantined_answers_by_id` — including the ordinal-comparison behavior of the backing dictionary.

Suite: `Capacitor.Cli.Tests.Unit` on macOS — baseline `origin/main` 4059 total / 44 failed (the known local tmp-symlink + parallel-load flakes: config-toml, MCP-registration, uninstall). Branch result recorded in a follow-up comment; failing sets diffed to confirm no new failures.

`scripts/check-linear-ids.sh` passes (no Linear IDs in `.cs`).


realtonyyoung:
### Local test results (macOS, `Capacitor.Cli.Tests.Unit`)

| | total | failed | skipped |
|---|---|---|---|
| baseline `origin/main` (4e7ee2c) | 4059 | 44 | 8 |
| this branch | 4062 | 42 | 8 |

`+3` total = the three new regression tests, all passing.

**Failing-set diff: zero new failures.** All 42 branch failures are a subset of the 44 baseline failures — the known local tmp-symlink / parallel-load flakes (`RegisterKcapMcpServers_*`, `UnregisterKcapMcpServers_*`, `InstallCodex_*`, `User_level_uninstall_*`, `EnableNetworkAccess_*`, …). Two baseline failures (`Dispose_flushes_the_partial_buffer_…`, `Transcript_channel_at_capacity_…`) did not reproduce on the branch — both are unrelated to these changes and are themselves flaky.

**Mutation check on the new guard** (so it is not vacuous): reverting both ack paths to build the `CommandAck` inside `_lock` makes `Liveness_is_never_read_while_the_processor_lock_is_held` fail with `readsUnderLock == 2` (one per path). Reverted before commit.

`scripts/check-linear-ids.sh` → exit 0.

realtonyyoung:
All three addressed at `b3a35df`.

**1 — `ai1526` in comments.** Removed. Both were design-doc paths, and they were doubly wrong for a public repo: they embed an internal issue slug *and* point at a file in the private `kcap-server` repo, so a public reader could never follow them. The API explanation stays; only the pointers went. (`check-linear-ids.sh` did pass on them — the script matches `AI-####`, not a lowercase `ai1526` inside a filename — so this was a real gap in the check, not just a lint miss.)

**3 — Error provenance.** Correct, and it was worth more than a README edit, so I fixed the message too. `LastCode` is null when the deadline cancels an in-flight request before any coded response is parsed — the very first attempt can time out — and the message still asserted *"the daemon is still settling a prior launch"* while printing the client's own default code in the `Error(...)` position, which reads as a server verdict. The code token stays stable, because agents match on it, but the narrative now names its source:

> `Error (flow_settlement_busy): gave up after 1 attempt over 3m — no coded response arrived before the deadline, so the code above is this client's default rather than something the server reported.`

Two tests pin both arms: no coded response must *not* say "still settling", and a real coded response must still say it. README updated to match.

**2 — Verbose concurrency comments.** Partly taken. I trimmed the duplication you pointed at: the two caller-side comments now point at `SendSettledAck` instead of restating its rationale, and `BuildProcessedAck`'s summary is shorter.

I did not remove the invariants themselves, and I want to be explicit about why rather than quietly declining. "Never invoke the liveness read under `_lock`" and "the cache is committed before the ack is built" are precisely what a five-round adversarial review kept finding violations of on this file — the second one caught a case where a committed `LaunchRejected(daemon_capacity)` could be replayed later as a contradictory `InternalError`. The code cannot express either constraint on its own; nothing stops the next edit from calling `Build(...)` at the call site again. That's the case where a comment earns its lines. Happy to cut further if you can point at a specific one that isn't load-bearing.

A codex review flow also ran on this PR to clean over five rounds; the fixes it produced are in the commits above `29c42be`.

Issue #379, opened 2026-07-28
kcap agent commands should be aware of flow participants


alexeyzimarev:
Follow-up to the `kcap agent start|ls|stop|attach` restructure (AI-1555).

## Problem

The local-attach path treats every daemon-hosted agent identically. `AgentOrchestrator.HandleLocalListAsync` returns all of `_agents` unfiltered, and `HandleLocalAttachAsync` will attach to any id in that map — including agents the daemon launched as **flow participants** (e.g. review-flow reviewers).

That lets a local user bypass an invariant that holds everywhere else: participants in a flow are addressed through the flow protocol (`send_to_participant`), not by typing directly into them. `kcap agent attach <reviewer-id>` hands the user raw PTY stdin on a participant, with no flow round, no routing, and no record of the interjection in the flow.

With `kcap agent stop` added by the same restructure, the same blind spot applies to termination: a local stop can kill a reviewer mid-round, and the flow has no idea why its participant vanished.

## What we want

`kcap agent` should know which of the daemon's agents are flow participants and treat them accordingly. Open questions for design:

- **Listing** — mark participants in `kcap agent ls` (flow id / role column) rather than showing them as ordinary agents.
- **Attach** — refuse by default with a message pointing at the flow tooling. Possibly allow a read-only attach (output stream, no stdin), which is genuinely useful for watching a reviewer work.
- **Stop** — refuse, or require an explicit override, and make sure the flow is notified rather than silently losing a participant.
- **Wire support** — the local IPC `AgentList` payload is `id\tstatus\tcwd`; carrying participant metadata needs an appended field (the frame contract is append-only).

## Not in scope

The `kcap agent` restructure itself (AI-1555) deliberately keeps today's behaviour here unchanged.

PR #383, opened 2026-07-28
Group the agent commands under `kcap agent`

### LINKED ISSUE #382: Group the agent commands under `kcap agent`
The local-attach feature shipped its commands at the top level of the CLI:

```
kcap run-agent <vendor> [flags] [-- <agent args>]
kcap attach <agent-id>
kcap ls
```

Three problems follow from that shape:

1. `attach` and `ls` are generic names occupying the top-level namespace. `kcap ls` says nothing about *what* it lists — a reader would guess sessions or repos before agents.
2. They are undocumented in `kcap --help`. `help-usage.txt` has no entry for any of the three; the only place they appear is the README, so they are undiscoverable from the CLI.
3. `--name` means the *daemon* name, not the agent name. Already confusing on `kcap attach --name`, and an outright trap on a `stop` subcommand.

There is also no way to stop a locally-started agent. The daemon only stops agents on a server-origin command, and that path explicitly refuses `--private` agents — so a `kcap run-agent --private` agent cannot be stopped from the CLI at all.

## Proposal

One coherent group covering the whole lifecycle of a daemon-hosted agent:

```
kcap agent                                          # same as `kcap agent ls`
kcap agent start <vendor> [flags] [-- <agent args>]
kcap agent ls                    [--daemon <name>]
kcap agent attach <id>           [--daemon <name>]
kcap agent stop   <id>           [--daemon <name>]
kcap agent stop   --all [-y]     [--daemon <name>]
```

Plus: `--name` becomes `--daemon`, `--detached` becomes `-d`/`--detach`, and `attach`/`stop` accept any unique agent-id prefix.

This is a **breaking change** — the old spellings are removed, not aliased.

alexeyzimarev:
Closes #382
AI-1555

Regroups the three top-level agent commands into one `kcap agent` group, and adds a `stop` subcommand that did not exist before.

## Command surface

```
kcap agent                                          # same as `kcap agent ls`
kcap agent start <vendor> [flags] [-- <agent args>]
kcap agent ls                    [--daemon <name>]
kcap agent attach <id>           [--daemon <name>]
kcap agent stop   <id>           [--daemon <name>]
kcap agent stop   --all [-y]     [--daemon <name>]
```

## Breaking change

`kcap run-agent`, `kcap attach`, and `kcap ls` are **removed, not aliased**. Two flags are respelled: `--name` → `--daemon` (it always meant the *daemon* name, which reads as an agent name once there is a `stop` subcommand) and `--detached` → `-d`/`--detach` (matching `kcap daemon start -d`). The old spellings are rejected as unknown flags rather than silently accepted.

## What else changed

- **`kcap agent stop`** is new end to end: two append-only local-IPC frames (`Stop = 8`, `StopAck = 70`), a `HandleLocalStopAsync` on the daemon, and the CLI subcommand. It reuses the existing graceful path — `/exit`, 15s wait, then terminate.
- **A local stop works on `--private` agents**, which nothing could stop before. `HandleStopAgent`’s body moved into `StopAgentCoreAsync`; the `IsPrivate` guard stays on the server-origin path (defence-in-depth against a leaked id) and only the 0600 local socket bypasses it. The core skips the two `_server.*` calls for private agents, since an unregistered agent has no server-side row.
- **Id prefixes:** `attach` and `stop` accept any unique prefix; an ambiguous one lists the candidates. A full 32-hex id is sent verbatim so the daemon’s PID-record fallback can still reap survivors of a prior daemon incarnation.
- **`kcap --help` now lists these commands at all** — previously all three were absent from `help-usage.txt`.

## Verification

- Unit suite: 42 failed / 4043 passed. Those 42 are a **pre-existing baseline** in `CodexHookCommandTests` and the uninstall/`config.toml` area — the merge base `7a4975a` fails the identical 42. Zero new failures.
- Integration suite: 144/144 passed.
- `dotnet publish -c Release` clean of IL3050/IL2026.
- `kcap agent start/ls/stop/attach` exercised live against a real daemon and the real `claude` CLI, including stop-by-prefix, `--all` with cancel/confirm/`-y`, unknown id, and the too-old-daemon path.

## Known follow-ups

- #379 — these commands are not flow-participant aware. `stop --all` stops **every** agent the daemon hosts, including server-launched agents and review-flow participants; that is now documented in `help-agent.txt` and the README, but the real fix is #379.
- Review suggested extracting a pure `AgentList` response classifier and a stop-arg validator as `internal static` seams (following `ResolveAgentId`’s precedent) so two fixed defects get regression pins.

## Design and plan

- `docs/superpowers/specs/2026-07-28-ai1555-agent-command-group-design.md`
- `docs/superpowers/plans/2026-07-28-ai1555-agent-command-group.md`


alexeyzimarev:
Triaged the bot reviews. Pushed `1c3c868` + `12e1a0a` addressing three of the five findings; two are deferred with reasoning, and one is a false positive.

## Fixed

**Qodo #3 / Codex P1 — valueless `--daemon` retargets a destructive command.** Confirmed and fixed. `NameFrom` returned `null` when `--daemon` was the final token, and `ResolveName(null)` resolves the *profile default* daemon — so `kcap agent stop --all -y --daemon` stopped every agent on the default daemon, with `-y` suppressing the confirmation that would have exposed it. A `--daemon` followed by a flag also crashed with an uncaught `ArgumentException`. Replaced with an `internal static` validating `DaemonNameFrom` that rejects both (and the empty string), checked in `ls`/`attach`/`stop` before any socket is resolved. `AgentStartArgs.Parse` had the same flag-as-value hole and now applies the identical rule, so the whole group agrees.

**Qodo #4 — failed stops reported success.** Confirmed and fixed. `StopAgentCoreAsync` now returns a success bool, `StopAck`’s payload is `id\tstatus` per line (`stopped`/`failed`), and the CLI prints per-agent results and exits 1 if any failed — matching `kcap daemon stop`’s existing exit-code behaviour. The frame is new in this unreleased PR, so the format change costs nothing.

Worth flagging: the first attempt at this fix was itself wrong, and review caught it. Reading `HasExited` immediately after `TerminateAsync` is *not* proof of death — `UnixPtyProcess.TerminateAsync` sends SIGKILL then issues a single non-blocking `waitpid` microseconds later, before the kernel reaps the child. A hung agent (the main reason anyone runs `agent stop`) would have been killed successfully and reported `Failed to stop`. `12e1a0a` polls briefly via `WaitForExitAsync` before calling a stop failed, with a test double modelling the reap landing just after the kill — it fails against the previous commit and passes now.

**Qodo #1 — `--yes` undocumented.** Fixed in `README.md`, which now also documents the failure line and non-zero exit.

## Deferred

**Qodo #5 / Codex P2 — `--all` confirmation TOCTOU.** Real: the client lists and prompts, then sends the empty sentinel and the daemon re-enumerates, so an agent launched during the prompt is stopped without appearing in the list. Binding execution to the confirmed ids changes the `Stop` frame’s payload contract, which I would rather not fold into this PR. Tracked alongside #379, which covers the related and larger problem that these commands cannot distinguish a flow participant from your own agent.

## Not a defect

**Qodo #2 — "Getting started remains unsynchronized".** The README’s `## Getting started` section covers install, setup, import, dashboard, and MCP servers; it has never introduced `run-agent`/`attach`/`ls`, so there is nothing there to re-point. The rule is to *check* both sections, which was done — the quick-start legitimately does not cover local agents, and the per-command section plus both cross-references were updated.

PR #392, opened 2026-07-29
Restore the `kcap agent` command group shadowed by the retired-verb tombstone

### LINKED ISSUE #391: `kcap agent` is dead on main: the retired-verb tombstone shadows the whole command group
Every `kcap agent` subcommand on `main` fails. Verified against a build of `6659557`:

```
$ kcap agent ls
Unknown command: agent — the daemon verb was renamed to 'daemon'.
Run `kcap daemon start|stop|status`, or `kcap --help`.
$ echo $?
2
```

Same for `agent start <vendor>`, `agent stop`, `agent attach`, and `agent --help`.

## Cause

Two changes landed a day apart and merged without a git conflict because they touch different regions of `Program.cs`:

- **#390 (AI-1569)** added a `if (command is "agent")` guard near the top of `Program.cs`, before any config resolution, printing a rename pointer and exiting 2. It was deliberately not an alias — the pre-rename daemon verb was meant to stay dead.
- **#383 (AI-1555)** added the `kcap agent start|ls|stop|attach` group at `case "agent":` in the switch further down.

The guard runs first, so the entire group is unreachable dead code.

## Why no test caught it

- #390's `RetiredAgentVerbTests` assert the tombstone fires — it still does.
- #383's tests exercise the arg parsers and daemon handlers directly and never go through `Program.cs`.

Nothing pinned "`kcap agent <sub>` actually reaches `AgentCommand`" end to end.

## Underlying question

Beyond the merge accident, `agent` now has two meanings in this codebase's history: it was the daemon group before the May 2026 rename to `daemon`, and #390 exists to catch users acting on that stale habit. `kcap agent start` used to mean "start the daemon" and now means "start a coding agent". The overlap on `start`/`stop` is real and worth a decision rather than an accident.

alexeyzimarev:
Closes #391
AI-1570

`kcap agent` is currently dead on `main` — every subcommand exits 2 with "Unknown command: agent". This restores it.

## What happened

#390 (AI-1569) added a `command is "agent"` guard at the top of `Program.cs` that exits 2 with a pointer to `kcap daemon`. #383 (AI-1555) added the `kcap agent start|ls|stop|attach` group at `case "agent":` in the switch below it. They merged without a git conflict — different regions of the file — and the guard runs first, so the whole group is unreachable.

Neither side's tests caught it. #390's assert the tombstone fires, which it does. #383's exercise the arg parsers and daemon handlers directly and never go through `Program.cs`. Nothing pinned that `kcap agent <sub>` reaches `AgentCommand` at all.

**This reverses a deliberate decision in #390** — that the pre-rename daemon verb stays dead so stale habits self-correct — so @realtonyyoung should weigh in. The call to keep `kcap agent` as the coding-agent group was made explicitly; I am not undoing it quietly.

## Changes

- Removed the tombstone guard from `Program.cs` and the two README paragraphs describing it.
- **Kept the signpost where it does not collide.** `kcap agent status`, `restart`, `logs`, `doctor`, and `service` only ever meant the daemon, so they now answer `` `status` manages the daemon — run `kcap daemon status`. `` This preserves what #390 was actually protecting against — a healthy daemon reading as dead mid-diagnosis — without keeping the verb dead. `start` and `stop` belong to both groups and dispatch to the agent group.
- `RetiredAgentVerbTests` → `AgentVerbDispatchTests`, inverted: it spawns the real binary and pins that `agent`, `agent ls`, `agent start`, `agent stop`, `agent attach`, and `agent --help` are not short-circuited before dispatch. That is the end-to-end guard whose absence let this through, and it fails against current `main`.

## Known limitation

The daemon signpost needs a configured server. The `agent` group resolves config before dispatching, so with no server set `kcap agent status` reports "No server configured" instead. #390's tombstone ran before config resolution and worked logged-out. I did not re-add a pre-config special case for `agent`, because that placement is exactly what shadowed the group — but it is a real regression in that one path and worth a second opinion.

## Residual ambiguity

`agent` has two meanings in this repo's history: the daemon group before the May 2026 rename, and now the coding-agent group. `kcap agent start` used to mean "start the daemon" and now starts a coding agent. The overlap on `start`/`stop` cannot be signposted away — see #391 for discussion.

## Verification

- Integration 157/157 pass, including the 8 new dispatch tests.
- Unit 4215 passed / 43 failed — all in the pre-existing `CodexHookCommandTests` / uninstall `config.toml` baseline (42-43 wobble, present at the merge base). `AgentCommandRoutingTests` 11/11, `AgentIdResolutionTests` 7/7, `AgentStartArgsTests` 12/12, `FrameCodecTests` 12/12.
- `dotnet publish -c Release` clean of IL3050/IL2026.
- Smoke: `kcap agent --help` exits 0 with the group help; `kcap agent status` prints the daemon pointer; `kcap agent frobnicate` prints usage.


alexeyzimarev:
Pushed `c17469d`. CI red was a second collision, not a flake — and the review found a real hole in the guard test I added.

## CI

Both platforms failed on `RetiredVerbScanTests.No_shipped_text_names_the_retired_daemon_verb`, from #389 (AI-1567). It bans the literal string `kcap agent` from every non-binary file under `kcap/`, `src/` and `npm/`. So the retirement of this verb was enforced in two places, not one: the runtime tombstone in #390 and this text ban in #389. #383 merged after the ban existed and simply never ran against it — CI was not re-run on the updated base.

Per the decision to keep `kcap agent` as the coding-agent group, the scan test is removed here. **That reverses deliberate work in both #389 and #390** — flagging it plainly rather than burying it in a diff. @realtonyyoung, this is worth your review; AI-1567 also mentions relaying the server's daemon remediation, so if the server still emits text naming `kcap agent` for the daemon, that half needs a decision too.

## Review findings

**@chatgpt-codex-connector was right that the dispatch test did not prove dispatch.** In a config-less environment every non-help case exited 1 at the missing-server gate and `--help` returned from the global help path, so no case actually reached `AgentCommand` — another pre-dispatch guard would have passed the very test written to catch pre-dispatch guards. Each case now asserts a string only `AgentCommand` emits, and `ls` targets a daemon name that cannot exist so the handler's own output is deterministic rather than depending on what happens to be running.

**Windows bypassed the daemon signpost** (both bots). `NotSupportedOnWindows` returned before the subcommand switch, so `kcap agent status` claimed coding agents are unsupported instead of pointing at `kcap daemon status`, which Windows does support. The signpost now runs ahead of the platform guard.

**The offline gate** (@qodo-code-review). `agent` joins `offlineCommands`, and `start` — the only subcommand needing a server, for the daemon to record to — reports that itself. `ls`/`attach`/`stop` only ever talk to the local socket, so requiring a server to list local agents was arbitrary. README updated.

## Two bugs the rewritten test then found

Worth noting because they were live on `main` from #383, not introduced here:

1. Any `argv[1]` was treated as a subcommand, so `kcap agent --daemon dev` failed with `unknown subcommand '--daemon'` instead of listing that daemon's agents. A leading flag is now an `ls` option.
2. `--no-update-check` is a global flag that `Program.cs` leaves in argv. Strict parsing read it as a subcommand — and for `start`, as a **vendor name**, so `kcap agent start --no-update-check` tried to launch a vendor called `--no-update-check`. Global flags are dropped before the split.

## Verification

Unit 4216 passed / 42 failed (the pre-existing `CodexHookCommandTests`/uninstall baseline, present at the merge base — the 43rd was the scan test). Integration 158/158, run three times consecutively after one transient failure. AOT publish clean.

PR #394, opened 2026-07-29
Copilot borrowed review: platform-resolved capability + a readable snapshot tool surface

realtonyyoung:
> **Landing state: the Copilot borrowed-review platform entry ships DISABLED.** Everything below is implemented, tested and live except the one line that consults the platform table. See "What is not closed" — a reviewer P1 that a tool-allowlist fix should not be the vehicle for.

Closes the CLI half of the Copilot borrowed-review defect: a borrowed Copilot reviewer could not read the snapshot it was handed, so it reviewed blind and truthfully reported the target file and branch as absent.

## The obvious fix does not work

The design's preferred option was to drop the exclusive `--available-tools` clamp and add `--deny-tool=write --deny-tool=shell`. **Probed live against Copilot CLI 1.0.75, and it does not contain writes.** Neither deny flag covers Copilot's file-**create** tool. A direct write to an absolute path outside the snapshot was not denied — it raised an inbound `session/request_permission` titled `Access paths outside trusted directories`, and this daemon's unattended policy for Copilot is `AutoApprove`, which grants exactly that shape. A deny-list here looks closed and is open; shipping it would have been strictly worse than the blind reviewer, which at least could not write.

## What ships instead

Keep the allowlist **exclusive** and widen it to the verified read ids (`glob`, `grep`, `view`) for a **borrowed-snapshot launch only**. Write and exec become *unrepresentable* rather than denied, so there is no permission request for anything to grant. Verified live: Copilot reports `bash`, `create`, `edit`, `web_fetch` among its disabled tools, raises zero permission frames, reads all three git-state sentinels (branch-only, tracked-modified, untracked) verbatim, and the decoy path outside the snapshot does not exist on disk afterwards.

The borrowed/non-borrowed split is load-bearing: it is what makes the server's shipped read-blind rejection correct rather than paranoid. Each direction is asserted separately.

## A readable allowlist is a write boundary, not a read one

Found by adversarial review, then reproduced. Widening the allowlist also widens what a path-taking read tool can be *pointed at*: Copilot answers an outside absolute path with the same permission request that defeated the deny-list, and `AutoApprove` grants it without inspecting the tool. Under the shipping allowlist a reviewer read a file outside the snapshot and echoed its contents back — which the still-enabled result channel would carry off the machine.

Two changes close that:

1. **A borrowed snapshot takes the `Fail` interaction policy**, overriding whatever the vendor declares. Structural rather than matched on the frame's title: under an exclusive read-only allowlist a correct launch raises **zero** interaction frames, so one arriving means the reviewer is reaching past its boundary and it is reaped. Cursor already runs under this contract, so it is a no-op there.
2. **The boundary moves below the vendor.** A borrowed launch is spawned under `sandbox-exec` with an inline `(deny default)` profile granting the snapshot root plus the minimum needed to start. A host without `sandbox-exec` resolves **unsupported**, not "supported without it" — the readable allowlist and the read boundary ship as one entry, because the read tools are what make an outside path reachable at all.

Proven three ways: an automated enforcement test running a real process under the *generated* profile; a live Copilot run reading all three sentinels and not reaching outside; and a live run with the outside permission **explicitly granted** where the read still failed — which is the property that matters, since it no longer depends on what the vendor asks.

## Capability is platform-resolved, not declared

The tool surface was verified on one platform, and tool identifiers and path-trust semantics are the security boundary, so an unverified platform advertises no borrowed review. This is **not** the version-keyed certification AI-1528 removed: that keyed on a vendor build, which auto-updates underneath a user and drifted by construction. An OS and architecture do not change under a running daemon.

One `ResolvedBorrowedReviewPolicy` feeds every consumer — advertisement, containment token, snapshot requirement, both pre-spawn validators, and the launch argv. Copilot's descriptor stops declaring borrowed review statically; a second, always-stale answer is worse than none, and the default fails closed.

## What is not closed, and why the entry is disabled

The profile must still grant recursive reads of `~/.copilot`, `~/Library/Keychains`, `/Library` and `/opt/homebrew` so the vendor can start and authenticate. All four are data-bearing. A build that silently accepted an out-of-bounds path could point the allowlisted read tools at them and exfiltrate with **no** interaction frame — `Fail` never fires and the sandbox permits it. The boundary is vendor-independent for arbitrary paths, not yet for those four.

Closing it means a per-launch HOME/state directory, authentication brokered through `COPILOT_GITHUB_TOKEN`/`GH_TOKEN` instead of the keychain grant, and runtime grants narrowed to executables and packages rather than whole config trees. That makes the daemon a credential-handling component — a decision in its own right, not a side effect of a tool-allowlist fix. Tracked as a follow-up.

So `Resolve` stays exactly as written and tested, and what ships disabled is the decision to consult it. Enabling costs one line; a tripwire test pins the disabled state so it cannot flip in a refactor. Until then the server answers a Copilot borrowed request with `vendor_containment_unreadable` plus the `context-only` remedy — honest, and strictly better than the silent blind review this issue was filed for.

## Verification

Four review rounds, every finding fixed at the production end rather than the test end:

- **The containment-mismatch test deadlocked** on a real ACP handshake once Copilot became independent-snapshot — caught by running the class, confirmed against an `origin/main` baseline worktree, moved to a synthetic descriptor.
- **A new test would have failed both CI legs**: `BuildProcessStartInfo` never called `ValidateBorrowedArtifact`. Fixed by adding it to the builder — the seam that decides whether the readable argv is emitted. Verified by forcing this host's entry to Linux/x64 and re-running green.
- **The enforcement test caught its own bug immediately**: resolving only a symlinked *leaf* is insufficient because `/tmp` and `/var` are themselves links, so the first implementation produced a reviewer that could not read its own snapshot — invisible to every profile-string assertion.
- **The sandbox was drawn at the wrong path** for a nested borrowed cwd, reintroducing blind review for a launch from `repo/src`.

Seven mutations killed: unverified-platform→supported (6 tests), read-tool leak into a non-borrowed launch (2), one extra allowlist id (3), descriptor-instead-of-policy resolution (1), call-site policy revert (1), snapshot-root→path (1).

Suites: `AcpHostedAgentRuntimeFactoryTests` 59/59, `CopilotBorrowedReviewPolicyTests` 17/17, `BorrowedReviewSandboxTests` 10/10, `AcpVendorDescriptorTests` 9/9, `AgentOrchestratorVendorTests` 103/103. Full `Tests.Unit`: 44 failures vs 53 on an `origin/main` baseline worktree run, my set a strict subset — zero regressions, all pre-existing macOS flakes.

Related: kcap-server AI-1527 PR 2 (merged) is the server-side read-blind rejection this pairs with.


realtonyyoung:
Qodo findings addressed.

**1. README omits the platform gate — fixed, and it was worse than stated.** The README claimed borrowed capability is advertised "on every platform where it is installed", which was already loose and is now simply wrong: Copilot advertises no borrowed review anywhere. Replaced with what actually happens — the `vendor_containment_unreadable` answer naming the daemon, and `mode="context-only"` as the working remedy — and corrected the two other places that described Copilot's borrowed posture (the flows-MCP section and the reviewer-setup section).

**3. Mutable tool allowlist — fixed, and a fair catch.** `ReadToolIds` was a shared mutable `string[]` reaching the argv builder by reference, so any code holding it could write `ReadToolIds[0] = "bash"` and silently hand every subsequent borrowed reviewer a shell. This array *is* the tool allowlist, i.e. the security boundary, and a security boundary should not be a writable static. Now an `ImmutableArray<string>`.

**2. Verbose policy comments — declining, deliberately.** Each of those paragraphs records something that was *reproduced*, not something inferred, and several exist because the obvious reading is wrong:

- `--deny-tool=write` does not cover Copilot's `create` tool — a reader who does not know that will "simplify" the exclusive allowlist back into a deny-list and reopen the hole.
- The platform key is not the version-keyed certification a previous issue removed, and the distinction is the entire justification for it existing.
- Resolving only a symlinked *leaf* is insufficient because `/tmp` and `/var` are themselves links — that comment marks a bug that shipped once in this very PR.

Moving them to docs would separate each rationale from the line it constrains, which is where it does its work. The repo's convention is heavy "why" comments on non-obvious invariants, and these are non-obvious by demonstration.

CI green, 4 codex review rounds to clean. Suites: factory 59/59, policy 17/17, sandbox 10/10, descriptor 9/9, orchestrator 103/103.

PR #395, opened 2026-07-29
[AI-1459] SessionStart memory index: Codex CLI adapter

realtonyyoung:
Wires the shared SessionStart memory subsystem into the Codex hook — the third harness onto that foundation after Claude and Cursor.

## What the foundation already gave us

Worth stating up front, because it shrank this change a lot: the shared subsystem **already had a `SessionStartHarness.Codex` output adapter** rendering exactly the required envelope, and its orchestrator already implements the monotonic budget (a `Remaining()` recomputed before every await), the lease state machine, and the fail-open catch-all. So this PR is the **call site**, not a renderer or a budget mechanism.

## Ordering

The fetch starts **before** the lifecycle POST so the two overlap, then is awaited under the budget remaining *at that instant* — `HookBudget.Remaining(processStart, "session-start") - HookBudget.Safety`, the headroom reserved for serialization plus the write — immediately before the stdout handshake.

Codex blocks on this hook's stdout, so:
- an unreachable/slow memory server degrades to the minimal handshake rather than delaying the write;
- the payload is fully serialized **before the first byte**, so a renderer fault can never emit a partial rich object followed by a second minimal one (Codex's parser accepts exactly one JSON value);
- post-stdout work (watcher-ensure, spool drain) still runs strictly after the write, through the existing ordering seam.

## Scope safety

Mirrors the Cursor adapter's guard: the git root discovered from the payload `cwd` is preferred, the payload `cwd` is the fallback, and a **blank scope root skips injection entirely** rather than letting the shared resolver fall back to the hook process's cwd and inject an unrelated repository's memories.

Eligibility comes for free from existing structure: nested `KCAP_SKIP=1` invocations return before this path, and excluded/disabled repos short-circuit above it — neither reaches the memory subsystem. Codex's SessionStart payload carries no lifecycle `source`, so the reason is reported as `New`; re-injection on a resume of the same session id is prevented by the shared lease keyed on (harness, session id) rather than by a signal we cannot observe.

## One deliberate asymmetry (please sanity-check this)

The **no-fragment** path keeps the pre-existing handshake constant instead of the shared adapter's rendering of the same envelope. Both encode `{"continue":true}`, but the adapter appends a trailing newline to *every* envelope it renders (`json + "\n"` — which Claude and Cursor already ship). Adopting it here would change the bytes Codex receives on **every** no-memory SessionStart (opt-out, exclusion, provider failure, budget exhaustion) for no gain, and byte-identity on that path is an acceptance criterion. So the constant wins for null; the adapter owns the fragment-bearing shape, which is the only genuinely new output. A test pins the asymmetry so it is a recorded decision, not an accident.

## Tests

`CodexSessionStartMemoryTests` (8, new) pins:
- byte-identical minimal handshake when no fragment exists (the load-bearing regression — it caught the trailing-newline difference above);
- one combined object carrying `continue` + `hookSpecificOutput.additionalContext`;
- a **single** JSON value with no trailing document;
- quote / backslash / newline / tab / non-BMP / empty-string escaping round-trips;
- the recorded newline asymmetry;
- Stop never carrying memory context.

Verified locally: CodexSessionStartMemoryTests 8/8 · CodexStdoutContractTests 2/2 (ordering contract intact) · ClaudeHookCommandTests 35/35 · CursorHookCommandTests 38/38 · CodexHooksInstallerTests 8/8. `check-linear-ids.sh` clean. AOT publish generates native code with no IL/trim warnings.

## Spec-vs-code note

The issue's rev-2 spec asserted that lifecycle POST and repo enrichment should be *post*-stdout. In the code as it stands they already run **before** the handshake write, and this PR did not restructure that — it inserts the memory fetch into the existing ordering rather than rewriting Codex's hook sequence, which would be a larger and riskier change than this issue's scope.

Closes AI-1459.


realtonyyoung:
NO FINDINGS

realtonyyoung:
## Review close-out

**Codex review flow: clean** (2 rounds on the original + 2 on the delta). **Qodo: both findings addressed.**

### Findings fixed

1. **[P1, found independently by both codex and Qodo] Unauthenticated memory client.** The production fallback was a bare `new HttpClient()`. Since `/api/memories/index` is bearer-authed and the shared provider hands the *rejected* bearer back to the factory after a 401, both the initial call and the refresh went out anonymous → retryable failure → Codex silently never received memory context on any authenticated deployment. Fixed with the same auth-aware refresh factory `ClaudeHookCommand` uses, extracted as `DefaultMemoryClientFactory` so it is assertable; `disposeClients` now only disposes clients we created.

2. **[P1, found by Qodo] Fail-open regression that fix #1 introduced.** The auth helper funnels through `EnsureAbsolute`, which calls `Environment.Exit(2)` on an unacceptable URL — from SessionStart that would kill the hook *before* the handshake, so Codex would receive no stdout and reject the session. `CanAttemptMemoryInjection` now gates auth discovery on the same `IsAcceptableUrl` predicate the validator uses.

3. **[Rule violation, Qodo] README.** Capability-matrix row for Codex CLI plus two prose mentions updated.

### Correction to this PR's original description

The description claimed fail-open as an end-to-end property. That is accurate **only for the memory branch**. A codex delta review correctly showed the SessionStart handshake can still be lost on an unacceptable base URL, because `AgentHookPoster.PostOrSpoolAsync` performs unguarded auth discovery before the write.

That hole is **pre-existing** — verified on `origin/main`, where `HandleSessionStart` already called `PostOrSpoolAsync` ahead of the handshake — so it is neither introduced nor widened here. It is not fixed in this PR because `PostOrSpoolAsync` is shared by all seven vendor hooks and the right disposition for an unpostable URL (skip-and-handshake vs. spool vs. exit) is a product decision affecting `ShouldSpawnAfter`/watcher sequencing. Tracked as **AI-1586** with the mechanism, provenance, and test guidance. Codex reviewed and accepted that scope split.

### Verification

New `CodexSessionStartMemoryTests` 16/16 · `CodexStdoutContractTests` 2/2 (ordering contract intact) · Claude 35/35 · Cursor 38/38 · installer 8/8. The auth guard is **mutation-tested** — reintroducing the bare client makes it fail. CI: both AOT publish checks, the linear-ID guard, and ubuntu build+test green; windows in flight.

PR #408, opened 2026-07-30
Make the `kcap agent` commands flow-participant aware

### LINKED ISSUE #379: kcap agent commands should be aware of flow participants

alexeyzimarev:
Closes #379
AI-1557

`kcap agent ls|attach|stop` treated every daemon-hosted agent identically, so a review-flow participant — a reviewer mid-round — was indistinguishable from an agent you started. `attach` handed you raw PTY stdin on it, and `stop --all` killed it silently. This teaches the commands the difference.

## What changed

`kcap agent ls` gains a `KIND` column:

```
AGENT                              STATUS     KIND         REPO
3f9acd34ef56ab78cd90ef12ab34cd56   Running    agent        ~/dev/api
ab99887766554433221100aabbccddee   Running    review-flow  ~/dev/api   [reviewer]
ff00112233445566778899aabbccddee   Running    review       ~/dev/web
```

For `review` and `review-flow` agents:

- **`attach` is read-only.** Output streams so you can watch a reviewer work; your keystrokes are not delivered, and your terminal size is not applied to it — a read-only viewer never enters `ClientDims`, so it cannot shrink the participants terminal through the min-clamp.
- **`stop` is refused** unless `--force`, with a message naming the flow.
- **`stop --all` skips them** and reports how many, rather than silently omitting them. `--force` includes them and lists them under their own heading in the confirmation, so the destructive variant shows *more* information than the safe one.

**Enforcement is daemon-side** for both, so a current client cannot bypass it.

## Wire changes

`FrameType` is append-only. `StopV2 = 10` carries a force flag; `AttachedReadOnly = 71` carries id + reason + snapshot. `AgentList` rows grow to `id⇥status⇥repo⇥kind⇥flowRunId⇥flowRole`; `StopAck` gains a `skipped` status.

`AttachedReadOnly` is a separate frame rather than a flag on `Attached` because `Attached`s payload ends with an unbounded snapshot — a trailing flag would be painted onto the users terminal instead of parsed. It also fails closed: an older CLI cannot decode frame 71, so it errors rather than pumping stdin at a participant.

## Version skew — please read

- **Older daemon:** it reports no kind, so `ls`/`attach` degrade silently to unprotected. `stop` does not degrade — it hard-fails, because the CLI sends `StopV2` which that daemon cannot decode, and tells you to restart it.
- **Older CLI:** it sends the legacy `Stop` frame, which has no force concept, so the daemon treats it as `--force`. An old client can therefore silently force-stop a protected agent. Frame 8 postdates v0.11.8, so no released client is affected — but it is a real hole and it is documented in the README rather than glossed.

## Known limitations

- `--force` still leaves the flow unaware its participant was stopped. Attributing that needs server-side work.
- Plain web-UI-launched agents (`LaunchKind.Default`) are deliberately not protected — they are your own work by another route.

## Verification

- Unit 4237 passed / 42 failed on this branch; those 42 are the pre-existing `CodexHookCommandTests`/uninstall-`config.toml` baseline confirmed at the merge base. Zero new.
- Integration 160/160 on the branch.
- **Trial-merged onto current `main` (13 commits ahead) and tested there:** plain `main` fails 45 unit tests, the merged result fails the same 45 with 21 more tests; integration 162/162 solo. So the merge is semantically clean, not merely conflict-free — a check worth doing explicitly after #383 auto-merged cleanly into a guard that shadowed it.
- Both AOT publishes clean of IL3050/IL2026.

## Design and plan

- `docs/superpowers/specs/2026-07-29-ai1557-flow-participant-aware-agent-commands-design.md`
- `docs/superpowers/plans/2026-07-29-ai1557-flow-participant-aware-agent-commands.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -961,7 +961,7 @@ Agent ids are long, so `attach` and `stop` accept **any unique prefix** — an a
 - `kcap agent attach` on one is **read-only** — you see its output, your keystrokes are not delivered, and your terminal size is not applied to it.
 - `kcap agent stop` on one is **refused** unless you pass `--force`, and `stop --all` skips them and says how many it skipped.
 
-Enforcement is in the daemon, so this holds regardless of client version. It does not apply when talking to a daemon older than this feature — that daemon reports no kind, every agent reads as `agent`, and the protections do not engage until it restarts onto the new binary.
+Enforcement lives in the daemon, so a current CLI can't bypass it by skipping the check or lying about `--force`. That guarantee has two version-skew exceptions: an old `kcap` sends the legacy `Stop` request, which has no `--force` concept — the daemon treats that as `--force`, so an old client can silently force-stop a review or review-flow agent. And against an old daemon, `ls`/`attach` degrade silently (no kind reported, every agent reads as `agent`), but `stop` doesn't degrade — it sends a newer request format the old daemon can't decode, so the connection closes and the CLI tells you to restart the daemon instead of stopping anything.
 
 `agent start` auto-starts the daemon if one isn't already running, and needs a configured server for the daemon to record to. `ls`, `attach`, and `stop` only talk to the local socket, so they work without one. A locally-started agent appears in **your own** web UI (owner-only until you share it from the web UI); use `--private` to opt out of registration entirely. Unix only for now.
 
--- a/README.md
+++ b/README.md
@@ -948,7 +948,7 @@ kcap agent start claude -d                    # start without attaching; prints 
 
 ```bash
 kcap agent                 # no subcommand — same as `kcap agent ls`
-kcap agent ls              # list daemon-hosted agents (id, status, repo)
+kcap agent ls              # list daemon-hosted agents (id, status, kind, repo)
 kcap agent attach ab12     # re-attach your terminal (any unique id prefix works)
 kcap agent stop ab12       # graceful /exit, then terminate
 kcap agent stop --all -y   # stop every agent this daemon hosts, no prompt
--- a/README.md
+++ b/README.md
@@ -954,7 +954,14 @@ kcap agent stop ab12       # graceful /exit, then terminate
 kcap agent stop --all -y   # stop every agent this daemon hosts, no prompt
 ```
 
-Agent ids are long, so `attach` and `stop` accept **any unique prefix** — an ambiguous one lists the candidates instead of guessing. `stop --all` includes `--private` agents and prompts for confirmation unless you pass `--yes`/`-y`. `--all` stops **every** agent this daemon hosts, not just the ones you started from this CLI — including agents launched from the web UI and review-flow participants (e.g. reviewers mid-round). Making these commands flow-participant aware is tracked in [#379](https://github.com/kurrent-io/kcap-cli/issues/379). If a stop can't be confirmed (e.g. an agent that never exits, even after SIGKILL), `stop` prints a per-agent "Failed to stop `<id>` — see `kcap daemon logs`." line and exits non-zero; `stop --all` still stops the rest and reports each agent's outcome individually.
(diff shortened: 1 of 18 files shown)


alexeyzimarev:
Fixed in `0501bf0`. @qodo-code-review is right, and the finding is slightly sharper than my own reviews had it.

Two earlier review passes rated this class Minor because the daemon computes eligibility itself and is authoritative for the stop — so a mis-parsed row cannot cause a wrong *stop*. That reasoning holds for the decision but not for the **confirmation prompt**, which is what the user actually consents to.

Traced concretely: with a tab in a repo path, `p[3]` becomes a path fragment, and since `IsProtectedKind` is deliberately fail-safe (`kind is not "agent"`), a **plain** agent reads as protected. The CLI drops it from `targets`, prints "Skipping 1 review agent(s)", prompts `Stop N?`, and sends `StopV2(force: false, "")`. The daemon then sees `Kind == Default` and stops it. The user confirms N and N+1 die — the prompt understates the blast radius, which is exactly the property it exists to convey. (The newline case is milder: a row fragment makes a protected agent read as stoppable, and the daemon correctly skips it.)

Closed at both layers, per the "preferable + minimal mitigation" split in the suggestion:

- **Daemon** (`AgentOrchestrator.LocalIpc.cs`): a `Cell()` helper replaces tab/CR/LF with a space in the three free-form fields (`RepoPath`, `FlowRunId`, `FlowRole`). `Id`, `Status` and `Kind` are all constrained values and need no escaping.
- **CLI** (`AgentCommand.FetchAgentsAsync`): accepts only **exactly 3** columns (older daemon) or **exactly 6** (current), and refuses the whole table with a clear error otherwise rather than acting on a guess. This replaces the previous `>= 3` floor, which admitted shifted rows.

I did not take the full structured/length-prefixed payload option. The daemon is the only producer, sanitisation makes every reachable case safe, and the strict column check fails closed on anything else — so the remaining benefit is bounded while the cost is a wire-format change to a frame in the same unreleased PR.

New test: `Local_list_neutralises_delimiters_inside_a_free_form_field` seeds a review-flow agent whose `flowRunId` and `flowRole` carry a tab and a newline, and asserts the emitted table is one row of exactly six columns with the kind column intact. I did not add a unit test for the CLI-side strictness — `FetchAgentsAsync` is socket-bound with no seam, and I would rather say so than add a test that asserts nothing.

Verification: unit 4238 passed / 42 failed (the pre-existing `CodexHookCommandTests`/uninstall baseline, zero new), integration 160/160, both AOT publishes clean.

PR #411, opened 2026-07-30
[AI-1505] Cursor subagent classification: correct the record, pin the contract, don't build the heuristic

realtonyyoung:
## What this is

AI-1505 asked for a **recency/mtime-filtered sibling-transcript signal** so an uncertain Cursor subagent classification could skip SessionStart memory injection. **This PR does not build that heuristic.** A controlled probe falsified the premise, and what it surfaced instead is a real defect in the same code.

**No runtime behaviour changes.** Comments, tests, and docs only.

## The probe

Four `cursor-agent 2026.07.23-e383d2b` runs in a throwaway workspace with project-level hooks (the developer's global `~/.cursor/hooks.json` was never touched), each confirmed to have spawned a real subagent — three via `Task`, one via a named custom subagent. Sanitized artifacts are committed under `docs/probes/2026-07-30-cursor-subagent-hooks/`.

**F1 — a Cursor subagent child never fires `sessionStart` or `sessionEnd`.** Children fire five other hooks. Corroborated server-side from a *different* hook installation: `model` (carried only on `sessionStart`) is populated for both parents and NULL for both children.

Memory injection is reached only from `sessionStart`, so a child cannot receive the memory index. The harm AI-1505 exists to prevent does not occur on this surface.

**F3 — live prompt-hash correlation is unavailable for the child's entire hook window.** AI-1461 modelled this as a race a later moment resolves. The run-4 timeline shows the parent's `Task` tool_use stays unflushed (1 line, 263 B) through *every* child hook and only lands after the child's last one. So no hook relocation and no mtime filter can recover a link.

**F4/F7 — the live linker cannot revive on its own, three times over.** Its guard needs `sessionStart` **and** a non-empty `transcript_path`; `sessionStart` always carries null. And even if Cursor dispatched its native `subagentStart` (which *is* fully implemented in the CLI bundle, carrying `parent_conversation_id`), `kcap` would drop it — the event is in neither `CursorHooksParser.CursorHookEvents` nor `CursorHookEventMap`.

## What changed

**Comments corrected (3).** Each asserted something the probe disproves:
- `CursorLiveSubagentLinker.SaveLink` claimed a marker loss only means a top-level fallback healed by import. True only if the write fails *before* a start side effect; after one, the child can be routed twice.
- The ack gate called the loss "diagnosable"; it returns silently with no log or metric.
- The memory-classification note asserted a residual misclassification risk and "no cheap signal exists".

**The inert path is annotated, not deleted.** Retained as the landing site if Cursor ships `subagentStart` dispatch. The comments record that a revival needs a *different trigger* (the parent's hooks — the current builders key off events a child never fires) **and** registration/mapping changes.

**Tests.** Four scenarios asserting child `sessionStart` → `subagent-start` / child `sessionEnd` → `subagent-stop` are removed rather than relabelled; two dispatcher tests that used a child `sessionStart` as a vehicle now seed the spool directly. Added pins for the measured payload contract and for the decided stale-state behaviour.

## Two things worth a reviewer's attention

1. **One new test is deliberately not mutation-sensitive, and says so.** Deleting the `transcript_path` conjunct leaves `SessionStart_with_null_transcript_path_…` passing, because a null path also fails downstream. Rather than overstate it, the limitation and the mutation result are recorded in the test body. The other pin *is* mutation-verified.
2. **A known corrupt-state risk is documented, not fixed.** A `SaveLink` write failure is swallowed *after* the caller has already set `subagentParentId`, so the divert still runs and the child's transcript can be routed both under the parent and as its own top-level session. Two characterization tests record this and are explicitly excluded from the mutation rule — the leading remedy is *expected* to break them. Fixing it is a runtime change, deliberately out of scope. It has no producer on the measured contract, and an audit of the developer's machine found zero marker/spool/ack state (weak, single-machine evidence — recorded as such).

## Open gate (not merge-blocking)

The Cursor **IDE** Agent Window was not exercised; it isn't scriptable from here. If an IDE child fires `sessionStart`, that would be a real correctness gap rather than a docs gap, and D1 would need reopening for that surface. §7 of the spec has the exact re-probe procedure and the three outcomes. This is not merge-blocking because the change alters no runtime behaviour and every claim is scoped to the tested CLI version — an IDE result can only widen the scope, not invalidate what ships. It needs a named owner with IDE access.

## Testing

- Cursor suites: `CursorHookCommandTests` 40/40, `CursorWatcherSpawnTests` 17/17, `CursorLiveSubagentIntegrationTests` 3/3, `CursorSubagentStaleStateTests` 6/6.
- Full `Capacitor.Cli.Tests.Unit`: 20 failures, all in the known macOS baseline (MCP registration, config file, stderr flood). **Zero Cursor/subagent failures**; none of the failing tests touch Cursor code.
- Spec is codex spec-review clean (flow `6e7a88cd`, 9 rounds, rev 10).


realtonyyoung:
## Qodo findings — all four addressed

**🐞 3. Ack marker not cleaned — VALID, fixed in `ca5e76b`.**
Correct and a real leak: I seeded `MarkSubagentStartAcked` but the fixture cleanup only deleted the link marker, and the ack marker lives in a separate directory (`cursor-subagent-start-ack`). Both `CursorLiveSubagentIntegrationTests`' fixture `Dispose` and `CursorSubagentStaleStateTests.TryDeleteMarker` now clear it, with a comment explaining why the second directory exists. This mattered more than it looks: `HasSubagentStartAck` is exactly what the fail-closed gate consults, so a leaked ack could have silently un-gated a later test.

**🐞 4. Test comment contradicts code — VALID, already fixed in `e5e1476`.**
Same finding the codex reviewer raised as a P1. `live_agent_id_matches_the_dashless_id_the_import_path_would_use` was still driving a child `sessionStart` and requiring a `subagent-start` back, contradicting the class comment I had just written. It now exercises `BuildSubagentStartPayload` directly — the parity property lives in the builder, not in how it is reached. Your review ran against `02f1732`, before that commit.

**📘 1. Linear ID in `probe_hook.py` docstring — fixed in `ca5e76b`.**
Removed. CI's `check-linear-ids.sh` only scans `.cs`, so this wasn't failing, but there's no reason for the ID to be there.

**📘 2. Verbose comment block in `RunMemoryOrchestrationAsync` — respectfully declining.**
This one I'd push back on. The block *replaces* a longer (~20-line) comment and is shorter than what it removes, so the change reduces verbosity rather than adding it. Its length is also load-bearing: it draws the distinction between what the **source** constructs (this method has one call site behind the `sessionStart` guard — an internal invariant a test can pin) and what makes `ClassificationAuthoritative: true` **warranted** (an external vendor behaviour, measured over four probe runs, that a Cursor update could change). Collapsing that to a one-liner is what produced the original misleading note this PR exists to correct — the previous version read as a proof and sent a reviewer down a two-round detour. It ends with a spec pointer rather than restating the analysis, which is the single-sourcing rule the design asks for.

Happy to trim if you disagree, but I'd rather over-explain this particular flag than under-explain it again.

---

Also worth flagging, since it's the kind of thing your **High-Level Assessment** weighed: your recommendation matched the design's own conclusion on all three alternatives — keep the inert path annotated rather than delete it (Cursor already ships `subagentStart` in its bundles and would need a landing site), defer runtime diagnostics for the silent ack-gate drop, and don't build the heuristic. Those are recorded as D2, D2a and D1 respectively in the spec.

PR #416, opened 2026-07-31
[AI-1463] Finalise the Gemini SessionStart memory certification

realtonyyoung:
Follow-up to #414, which merged with all four of its definition-of-done items still open. This runs the measurement they were gating on, writes the result back into the spec, and fixes the harness defects that running it exposed.

**Test and docs only — no production code changes.**

## §3.3 is settled — live, row 1, option (a)

Against **`gemini 0.53.0`**, `kcap 0.11.10-alpha.0.10+998ec34` (the tree merged as `e2b2821`), macOS 26.5.2 / osx-arm64.

The failure is real rather than simulated: a WireMock instance proxies the whole API to the configured server and fails **only** `POST /hooks/session-start/gemini`. The index fetch on the same base URL is proxied untouched — the one arrangement that produces the shape under test.

| Link | Observed |
|---|---|
| lifecycle POST | `400` from the forced mapping, matched on its response-body marker |
| index fetch | `GET /api/memories/index?machine=… → 200`, same base URL, same invocation |
| hook exit code | **non-zero**, read from the hook Gemini itself ran |
| hook stdout | `hookSpecificOutput.additionalContext` carrying the nonce |
| gemini turn | exit **0** — completed |
| model receipt | the nonce reproduced in the answer |

The turn continues **and** the context is consumed, so option **(a)** — no commit gate — is the design and option (b) is withdrawn.

### Getting the measurement honest took four attempts

Each rejected version looked sufficient, and each was holed in review. Worth recording, because the same shapes will recur in the next adapter's cert:

1. **Substitution.** Proving the non-zero exit from a *separate* direct `kcap hook --gemini` run — Gemini's own hook could have returned 0 through a session- or source-dependent branch.
2. **Turn-global.** "Some forced 400 happened during the turn" — with two invocations, one could carry the nonce while a *different* one took the 400.
3. **Shared identifier.** Correlating those two on session id — but a session id is reused across invocations (`startup` and `resume` share one).
4. **Per-invocation evidence** (this one): a recording shim named `kcap` on the PATH Gemini inherits captures each invocation's exit code, stdout and stderr. The invocation whose `additionalContext` carried the nonce must *itself* have exited non-zero and written the failed-POST diagnostic to *its own* stderr.

## Harness defects the run exposed — all fixed

None is a product defect. The first three made the cert **lie**, which is worse than a cert that fails.

1. **The trusted-folder gate silently voided the run.** Certs run in a fresh throwaway worktree, which 0.53.0 refuses as untrusted (`exit 55`, before any model call). The positive case failed honestly; the negative control **passed vacuously** — a turn that never ran trivially contains no nonce. Fixed with `--skip-trust`; the negative control now asserts turn completion too.
2. **The recorded `kcap` version described the wrong binary.** `Process.Start` resolves a separator-free filename against the working directory before PATH, and this assembly's output folder holds a `kcap` copied there by the project reference. `RecordCertEnvironmentAsync` — which exists to stop a stale-binary misdiagnosis — recorded the test build while its own sibling `which kcap` line reported the PATH build. Resolution is now explicit, asks `access(X_OK)` rather than `File.Exists`, and **throws** rather than falling back to a bare name the working directory would satisfy.
3. **An `int.MinValue` sentinel satisfied the property under test.** An unreadable exit record is `!= 0`. Now nullable: a parse failure fails the assertion.
4. **A PATH shim must not wrap the MCP servers.** The recording shim is a `kcap` on PATH, and so are the four long-lived `kcap mcp` stdio servers Gemini launches. Buffering a stdio JSON-RPC server's stdout traps its handshake until it exits, which it never does — every run hung at exactly 120s with four wrapped processes alive and zero HTTP traffic. The shim now `exec`s anything that is not `kcap hook`.
5. **Cleanup was a log line, not a gate.** A leaked nonce memory makes a *later* run's positive case pass on this run's evidence. Now: `ArchiveMemoryAsync` reports whether the archive was confirmed; an ambiguous save (no id **or** a throw from the call) sweeps for the slug and marks the run dirty unconditionally; a prospective gate refuses to start further cases and an `[After(Assembly)]` teardown fails the run — because a cleanup failure in the *last* case is otherwise observed by nobody.

## New coverage

- `GeminiMemoryIndexLiveCertTests` — the non-zero-exit case (gated), plus turn-completion in the negative control.
- `GeminiSessionStartHandshakeOnPostFailureTests` — the integration half the test plan called for, carrying the branch-specific lease assertion (a) requires: the failed-POST invocation *completes* the lease, and a later `resume` emits no context.
- `MemoryIndexLiveCertHarnessTests` — seven hermetic cases pinning PATH resolution.
- `LiveCertIndexCleanlinessGuard` — the assembly teardown.

## Verification

| Suite | Result |
|---|---|
| `GeminiMemoryIndexLiveCertTests` (live, gated) | 3/3 |
| `MemoryIndexLiveCertHarnessTests` | 30/30 |
| `GeminiSessionStartMemoryTests` | 30/30 |
| `SessionStartMemoryFoundationTests` | 28/28 |
| `GeminiSessionStartHandshakeOnPostFailureTests` | 1/1 |

**Mutation-proven, not asserted:** reverting the execute-bit check to `File.Exists` fails exactly the two resolution tests; resuming a different session id fails the lease assertion; dropping the dash normalisation failed the (since-replaced) id correlation; pointing the stderr diagnostic at a route that does not fail here fails exactly that assertion.

**Review:** six rounds with a hosted codex reviewer, clean at round 6; Qodo's `PATH fallback runs cwd` fixed in `cbeaf65` — it caught the working-directory fallback that the review flow did not reach.

Also pinned, because it cost a wrong conclusion mid-run: `session_id` is validated with `Guid.TryParse`, and a failure takes the same emit-and-return-0 path as a suppressed session — a decorated id produces a plausible exit 0 with the allow object and **no HTTP traffic at all**.

### DIFF
--- a/docs/superpowers/specs/2026-07-30-ai1463-gemini-memory-adapter-design.md
+++ b/docs/superpowers/specs/2026-07-30-ai1463-gemini-memory-adapter-design.md
@@ -434,27 +434,38 @@ real index and then failed its POST.
 a live alternative. Parsing was not assumed to be consuming: the model reproducing the nonce is what
 distinguishes them, and it is the assertion the cert makes.
 
-**Every link is measured on the same invocation; nothing is inferred from a stand-in or from a
-turn-global count.** Two review rounds were needed to get here, and both objections were right:
+**Every link is measured on the same invocation; nothing is inferred from a stand-in, a turn-global
+count, or a shared identifier.** Three review rounds were needed to get here and each objection was
+right. They are worth recording, because each rejected version looks sufficient:
 
-* A first draft proved the non-zero exit from a SEPARATE direct `kcap hook --gemini` run. That is a
-  substitution — every assertion could hold while Gemini's own hook returned 0 through some session- or
-  source-dependent branch. Fixed by prepending a recording shim named `kcap` to the PATH Gemini inherits,
-  capturing each hook invocation's stdin, stdout and exit code.
-* The second draft then asserted "some forced 400 happened during the turn", which is turn-global: with
-  two hook invocations, one could carry the nonce and exit non-zero for an unrelated reason while the
-  *other* hit the forced failure, and the cert would still be green. Fixed by correlating on session id —
-  the invocation whose `additionalContext` carried the nonce must be the invocation whose POST the forced
-  mapping rejected.
+1. Proving the non-zero exit from a SEPARATE direct `kcap hook --gemini` run is a **substitution** —
+   every assertion could hold while Gemini's own hook returned 0 through some session- or
+   source-dependent branch. Fixed by a recording shim named `kcap`, prepended to the PATH Gemini
+   inherits, that captures each hook invocation's exit code, stdout and stderr.
+2. "Some forced 400 happened during the turn" is **turn-global** — with two hook invocations, one could
+   carry the nonce while a *different* one took the 400.
+3. Correlating those two on **session id** narrows it but does not close it: a session id is reused
+   across invocations (`startup` and `resume` share one), so the same counterexample survives with equal
+   ids.
+
+The closing evidence is per-invocation: the invocation whose `additionalContext` carried the nonce must
+itself have written the failed-POST diagnostic (`… session-start/gemini: HTTP 400`) to *its own* stderr.
 
 So the assertion chain is: the recorded invocation that delivered the nonce **is** the one that exited
-non-zero, **and** is the one whose session-start POST the forced mapping (identified by its response-body
-marker, so an upstream 400 relayed by the catch-all proxy cannot stand in) rejected, **and** the turn
-completed, **and** the model reproduced the nonce.
+non-zero, **and** is the one whose own stderr reports the rejected session-start POST, **and** that 400
+came from this test's forced mapping rather than from upstream (matched on the response-body marker),
+**and** the turn completed, **and** the model reproduced the nonce.
 
 The recorded exit code is nullable by design: an unreadable or half-written record fails the assertion
 rather than satisfying it. An earlier version used an `int.MinValue` sentinel, which is non-zero and
 therefore passed the exact property under test — the shape of a false-pass, caught in review.
+
+**Cleanup is now a gate, not a log line.** A leaked nonce memory makes a later positive case pass on the
+previous run's evidence, so any cert that cannot *confirm* its nonce memory was removed — an
+unconfirmed archive, a throwing archive, or a save whose id could not be recovered — marks the run, and
+`SkipUnlessLiveGateReady` then refuses to start any further live case in that process. Recovery-by-slug
+is polled rather than asked once, and distinguishes "confirmed absent" from "could not tell", because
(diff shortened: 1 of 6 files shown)


realtonyyoung:
## Qodo review — both items addressed

### 🐞 "PATH fallback runs cwd" — FIXED in `cbeaf65`

Two distinct problems in this finding, and Qodo is the only reviewer that caught the second:

**Executability** — already fixed in `fc8781d`: resolution now asks `access(path, X_OK)` through libc rather than testing `File.Exists`, so a non-executable shadow earlier in PATH is skipped and the walk continues, exactly as `which`/`execvp` do. Two mutation-proven tests cover it.

**The bare-name fallback** — fixed now. This one was live and the code-review flow never reached it. The comment claimed returning the bare name lets "Process.Start produce its own, clearer error", which is wrong for the one command that matters: `Process.Start` consults the **working directory** before PATH, and this assembly's output folder holds a `kcap` copied there by the `Capacitor.Cli` project reference. So the fallback produced no error at all — it silently ran the test build, reintroducing the wrong-binary failure on precisely the path where resolution had already failed.

It now throws `FileNotFoundException` naming the PATH it searched. The two affected tests assert the throw rather than the passthrough:

- `An_unresolvable_command_throws_rather_than_falling_back_to_the_working_directory`
- `A_non_executable_only_match_does_not_resolve`

### 📘 "Overly verbose test comments" — not changing these, and the reason matters

I'd normally take this. Here I think the comments are load-bearing rather than decorative, on two grounds:

**They match the file's established convention.** `MemoryIndexLiveCertHarness` and the sibling cert classes already document *why* each guard exists in multi-paragraph XML docs — the `archive_memory` doc explains that `id` vs `memory_id` leaked 13 memories into the live index; the `[NotInParallel]` doc explains that it prevents concurrency but restores nothing. Terse comments here would be the outlier.

**Each paragraph records a defect that was actually shipped or nearly shipped**, and every one of them was a *test that passed while proving nothing*:

- the negative control passed vacuously because gemini exited 55 on the trust gate and a turn that never ran trivially contains no nonce;
- the recorded `kcap` version described the test build while the sibling `which kcap` line reported the PATH build;
- an `int.MinValue` sentinel for an unreadable exit code is `!= 0` and satisfied the exact property under test;
- a session id is reused across `startup`/`resume`, so correlating on it does not prove per-invocation causation;
- wrapping the PATH `kcap` broke the four long-lived `kcap mcp` stdio servers and hung every run at exactly 120s.

Every one of those is invisible from the code and re-introducible by a well-meaning simplification. Deleting the explanation is how the next person re-derives it the expensive way — this cert has already cost two sessions to that failure mode. Happy to be overruled if there's a specific paragraph that reads as noise rather than as a trap marker.

---

Six rounds of code review (codex, clean at round 6) plus this. Live cert 3/3 against `gemini 0.53.0`, harness 30/30, integration 1/1.

PR #418, opened 2026-07-31
[AI-1571][AI-1620] Fix two Windows-only CI flakes: concurrent Console capture + shared config dir

realtonyyoung:
Fixes the two known Windows-only kcap-cli CI flakes. Linear: [AI-1571](https://linear.app/kurrent/issue/AI-1571), [AI-1620](https://linear.app/kurrent/issue/AI-1620).

Both are the same underlying shape: **process-global state shared between test classes whose `NotInParallel` keys don't overlap**, so they run concurrently.

## AI-1571 — `ImportVisibilityTests` auto-skip: "Auto-skipping" vs "unknown auth provider"

**The issue's hypothesis was wrong** — this is not an AI-1504 provider-validation regression, and nothing rejects the seeded provider. `'NonsenseProvider'` appears in exactly one place in the repo: `LoginDiscoverTests.Exchange_rejects_unknown_provider`, which writes `Error: unknown auth provider 'NonsenseProvider'` straight to **ambient `Console.Error`** (`OAuthLoginFlow.cs:300/361`).

The import test captures stderr by swapping the **process-global `Console.Error`**, while keyed to `"ResolvedStateMutation"` — a different group from every other capturing test. So when two capturers overlap:

- the other capturer saves *our* writer as its "original" and restores it when it finishes — after which our own `Auto-skipping` line is written to **that test's** buffer, not ours;
- meanwhile an unrelated stderr write (the login-flow error) lands in ours.

That produces precisely the observed pair: buffer contains the foreign error and *not* `Auto-skipping`. Windows-only is a timing artifact, not a platform behaviour — the clobber is pure .NET static state.

**Fix:** bare `[Test, NotInParallel]` (globally exclusive), matching every other Console-redirecting test in this suite — the remedy `CodexHookCommandTests` already documents after hitting the identical "stale Console.Out writer" problem. Bare also subsumes the `ResolvedStateMutation` serialization the `SetResolvedState` call needs, since nothing runs alongside a bare-`NotInParallel` test.

**Two latent instances of the same defect, fixed while here** (they could clobber the above, and each other):
- `UnusableUrlGuardTests.InlineDrain_emits_its_own_guard_diagnostic` — captured `Console.Error` with **no `NotInParallel` at all**.
- `UninstallCommandTests.Project_flag_errors_when_not_inside_git_tree` — keyed to `"ConsoleStreams"`, a group with **no other member**, so it serialized against nothing. (Note `ConsoleSerialGroup`/`ConsoleStreams` are referenced in comments but never defined as a shared constant; bare `NotInParallel` is what the working tests actually use.)

## AI-1620 — `CrossProcessRefreshTests` `BeforeTest` hook: `config.json` sharing violation

The `[Before(Test)]` hook deletes `config.json` and the tokens dir inside the **assembly-wide shared** `KCAP_CONFIG_DIR` (`RepoPathStoreGlobalSetup`), but the class was keyed only to `TokenStoreProfileTests` — leaving 20+ other `config.json` readers/writers free to hold a handle concurrently. Windows rejects the delete; because it throws out of a `Before` hook, the three tests are reported failed **before they run**. Unix unlinks an open file silently — hence Windows-only.

Note the issue's suggested "give each class its own config dir" isn't available: `PathHelpers.ConfigDir` is `static readonly`, captured once per process, so there is exactly one config dir for the whole assembly.

**Fix:**
- Class-level bare `[NotInParallel]`, and the four **per-method** keyed attributes removed — method-level takes precedence over class-level in TUnit, so leaving them would have silently re-weakened the class-level guarantee.
- The deletes are now best-effort with a short bounded retry (20 × 25 ms) that swallows `IOException`/`UnauthorizedAccessException` and never throws from the hook — covering a handle held from **outside** the test host (an unreaped watcher/daemon child), which exclusivity alone can't prevent.

## Verification

Windows CI is the real oracle for both (neither reproduces on macOS: the AI-1620 delete succeeds silently on Unix, and AI-1571 needs an unlucky interleave). Locally, full `Tests.Unit` on this branch vs pristine main on the same machine:

| | main | this branch |
|---|---|---|
| total | 4770 | 4770 |
| failed | 49 | **45** |

- **Zero new failures** (set-diff: nothing appears on the branch that didn't fail on main).
- Every remaining failure is in the known pre-existing macOS home-collision family (`RegisterKcapMcpServers_*`, `User_level_uninstall_*`, `EnableNetworkAccess_*`, Codex `config_toml`) — green in CI's clean home.
- None of the four touched classes fail: `CrossProcessRefreshTests` 4/4, `ImportVisibilityTests` 48/48, `UnusableUrlGuardTests` 17/17, `LoginDiscoverTests` 8/8.
- The 4-test delta is suite noise, not a claimed effect — only one of those four captures Console.
- `check-linear-ids.sh` passes (IDs kept out of `.cs`).

Since AI-1571 manifests ~50% of Windows runs, a single green Windows leg isn't proof; worth a couple of reruns before trusting it.

### DIFF
--- a/test/Capacitor.Cli.Tests.Unit/CrossProcessRefreshTests.cs
+++ b/test/Capacitor.Cli.Tests.Unit/CrossProcessRefreshTests.cs
@@ -16,7 +16,14 @@ namespace Capacitor.Cli.Tests.Unit;
 /// A fake refresh delegate is injected so no network or real provider endpoint is touched. Shares
 /// the KCAP_CONFIG_DIR token store, so it cleans up and runs non-parallel like TokenStoreProfileTests.
 /// </summary>
-[NotInParallel(nameof(TokenStoreProfileTests))]
+// Globally sequential (NO group key). The Before(Test) hook below deletes config.json and the
+// tokens directory inside the ASSEMBLY-WIDE shared KCAP_CONFIG_DIR (see RepoPathStoreGlobalSetup),
+// so it must not overlap ANY test that touches those files — a keyed group only serialized this
+// class against TokenStoreProfileTests, leaving every other config.json reader/writer free to hold
+// a handle open. On Windows that sharing violation throws out of the hook (BeforeTestException),
+// failing these tests before they run; on Unix the unlink silently succeeds, which is why it was
+// Windows-only. Bare NotInParallel subsumes the TokenStoreProfileTests serialization.
+[NotInParallel]
 public class CrossProcessRefreshTests {
     static readonly TimeSpan Window = TimeSpan.FromMinutes(5);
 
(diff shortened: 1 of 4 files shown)


realtonyyoung:
Codex static review: **no blocking issues**. It independently confirmed the load-bearing claims — bare `[NotInParallel]` is globally exclusive and therefore subsumes the keys it replaced; method-level attributes override class-level in TUnit (so removing the four per-method keyed attributes was required, not cosmetic); `ResolvedStateMutation` has no remaining referent; and *all* `Console.SetOut`/`Console.SetError` captures under `test/` are now protected by bare `NotInParallel`.

Its one non-blocking finding is addressed in 9c8b98b, and I think it was the right call: `DeleteBestEffort` swallowed a **persistent** sharing violation as success, which for these tests risks a **false pass** rather than a red — they assert on token state, and a leftover tokens directory already satisfies the "a peer already refreshed it" assertion. It's now `ClearWithRetry`: same bounded retry (40 × 25 ms), but after the budget it re-checks and throws with the stuck file named and the last IO exception as inner. The wasted trailing sleep is gone too.

That is a deliberate deviation from AI-1620's "tolerate rather than throw" suggestion — the reported symptom is a *transient* holder, which exclusivity + retry fully absorb; a file still held after a second is a different problem and worth surfacing. Noted on the issue.

Also for the record on the earlier ubuntu red: that was `WorktreeManagerTests.CreateAsync_ConcurrentBaseRefs_EachWorktreePinnedToCorrectSha` (1 of 4770) — concurrent `git worktree add` on one clone hitting a git metadata race (`fatal: failed to read .git/worktrees/review-2/commondir: Success`). Unrelated to this diff; a separate third flake.

PR #419, opened 2026-07-31
[AI-1624] Fix daemon termination on unhandled dispose exception in LocalPermissionBridge

realtonyyoung:
Fixes [AI-1624](https://linear.app/kurrent/issue/AI-1624/kcap-daemon-terminates-on-shutdown-unhandled-objectdisposedexception).

## The bug

The daemon doesn't shut down — it **crashes**, and launchd restarts it. Five occurrences in `daemon-tony.log` across two days:

```
crit: kcap.Daemon  AppDomain.UnhandledException (terminating=True)
System.ObjectDisposedException: The CancellationTokenSource has been disposed.
   at LocalPermissionBridge.StopAsync()
   at LocalPermissionBridge.DisposeAsync()
   at ServiceProviderEngineScope.DisposeAsync()
   at Host.DisposeAsync()
   at DaemonRunner.DisposeHostAsync()
```

`LocalPermissionBridge` is both an `IHostedService` and an `IAsyncDisposable`, so the host stops it and the DI container then disposes it — and a SIGTERM-driven host shutdown can race the daemon's own shutdown sequence. `DisposeAsync` disposed `_cts` but left the field non-null, so a second or concurrent pass reached `StopAsync`'s `_cts.CancelAsync()` on a disposed CTS. Because the throw surfaces inside DI teardown, nothing catches it and the process dies.

## Why it's worth fixing now

**It kills in-flight review-flow reviewers.** The daemon dying mid-flow orphans the reviewer, and the symptoms then cascade into something that looks unrelated: `participant_died`, then `Timed out waiting for reviewer agent to register with the daemon`, then `reviewer_recovery_in_progress` while the server reconciles — with every retry extending the backlog it's clearing. During AI-1505 this killed three review flows and sent me chasing machine load for a while, because the box genuinely was loaded. The daemon log is what settles it.

## The fix

- `DisposeAsync` runs its body at most once (interlocked guard) and never lets an exception escape into the host's dispose path.
- `_cts` is nulled via `Interlocked.Exchange` **before** disposal, and `StopAsync` reads the field once and treats `ObjectDisposedException` as "already stopped".

Both halves are needed. The dispose guard alone doesn't cover a `StopAsync` arriving after a completed dispose — that's the reverse interleaving, and it's one of the tests.

## Verification

Three regression tests for the interleavings: `stop → dispose → dispose`, `stop → dispose`-then-`stop`, and concurrent `dispose ‖ stop`.

**Proven to catch the bug rather than assumed to.** Reverted the fix and re-ran: all three fail with `ObjectDisposedException: The CancellationTokenSource has been disposed` — the exact production exception — and **nothing else in the 46-test suite fails**. Restored, 46/46 green; `DaemonHostDisposalTests` 2/2.

## Scope

Two files, no behavioural change beyond making teardown survivable. Nothing about the bridge's request handling, token model or port claiming is touched.

### DIFF
--- a/src/Capacitor.Cli.Daemon/Services/LocalPermissionBridge.cs
+++ b/src/Capacitor.Cli.Daemon/Services/LocalPermissionBridge.cs
@@ -146,11 +146,13 @@ internal sealed partial class LocalPermissionBridge(
     }
 
     public async Task StopAsync(CancellationToken cancellationToken) {
-        // Read the field ONCE: DisposeAsync nulls it after disposing, so re-reading could see a
-        // non-null reference here and null moments later. Cancelling an already-disposed CTS
-        // throws, and this method runs on the host's dispose path where a throw is fatal — so
-        // treat that as an already-stopped bridge rather than an error.
-        var cts = Interlocked.CompareExchange(ref _cts, null, null);
+        // Read the field ONCE. DisposeAsync exchanges it to null BEFORE disposing, so a plain read
+        // is enough: reference reads are atomic, and the only bad outcome — capturing a non-null
+        // reference that is disposed a moment later — is handled by the catch below. (An
+        // interlocked read would add nothing here.) Cancelling an already-disposed CTS throws, and
+        // this runs on the host's dispose path where a throw is fatal, so treat it as an
+        // already-stopped bridge rather than an error.
+        var cts = _cts;
         if (cts is not null) {
             try { await cts.CancelAsync(); }
             catch (ObjectDisposedException) { /* already stopped and disposed — nothing to cancel */ }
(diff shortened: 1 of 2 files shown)


realtonyyoung:
## Qodo: 0 bugs, 1 rule violation — partially accepted

**📘 Verbose shutdown comment block — accepting one half, declining the other.**

**Accepted:** the *operational* detail doesn't belong in source. "5 occurrences in `daemon-tony.log` over two days" is incident metadata that will be stale within a release and is already recorded on [AI-1624](https://linear.app/kurrent/issue/AI-1624/kcap-daemon-terminates-on-shutdown-unhandled-objectdisposedexception). I'll trim the frequency and log references from both files.

**Declined:** the *mechanism* stays. This isn't background colour — it's the thing two reviewers had to correct me on, twice:

- I first attributed the double disposal to a SIGTERM-driven shutdown racing the daemon's own sequence, and shipped that explanation in a comment.
- The real cause is that `DaemonRunner` registers this type through **two** DI descriptors (`AddSingleton<LocalPermissionBridge>()` plus an `AddHostedService` factory resolving the same instance), and Microsoft DI tracks disposables *per descriptor* without de-duplicating by reference — so the scope disposes one instance twice, deterministically.

A reader who doesn't know that looks at an interlocked `_disposed` guard on a service that is registered once in the file they're reading, concludes it's defensive cruft, and deletes it. The comment exists to prevent exactly that deletion. Same for the two `SYNTHETIC` labels: without them, a maintainer reasonably assumes those tests model production ordering and "fixes" them to match — which is how the wrong narrative got into this file in the first place.

The verbosity here is the residue of a correction, not decoration. I'd rather carry ten lines than have someone re-derive a cause that took a review round to establish.

Happy to be overruled on the mechanism too if you'd prefer it lives only in the Linear issue — but the deletion risk is real and the issue isn't in the reader's field of view.

---

**Gate status:** codex code review **clean**; CI green on `4abcaa4` including Windows; 0 Qodo bugs.

PR #420, opened 2026-07-31
[AI-1625] Serialise .git/worktrees-mutating git commands per repository

realtonyyoung:
Fixes the concurrent-`git worktree add` race. Linear: [AI-1625](https://linear.app/kurrent/issue/AI-1625).

## The bug

`git worktree add` enumerates the existing `.git/worktrees/*` entries while creating its own, so two concurrent adds on one repository can have the later one read a sibling whose metadata is still being written:

```
fatal: failed to read .git/worktrees/review-2/commondir: Success
```

The `: Success` is the tell — errno 0 means the read didn't fail for a filesystem reason, it came back **empty**, and only a concurrent writer can leave a sibling's `commondir` zero-length mid-creation. `worktree remove` mutates the same tree.

**This is reachable in production, not just under test.** The daemon creates worktrees concurrently when a flow launches several reviewers against one source repo — precisely the concurrency `CreateAsync`'s per-worktree fetch ref was introduced for ("so concurrent review launches in the same source repo can't race on each other's fetches"). The fetch was guarded; the `worktree add` immediately after it was not, and neither was `RemoveAsync`'s `worktree remove`.

Surfaced as a CI flake: `WorktreeManagerTests.CreateAsync_ConcurrentBaseRefs_EachWorktreePinnedToCorrectSha`, 1 failure of 4770 on [run 30627851370](https://github.com/kurrent-io/kcap-cli/actions/runs/30627851370/job/91147118706) (on PR #418, whose diff was provably unrelated).

## The fix

A static per-repository `SemaphoreSlim` gate around all three mutating call sites — the two `worktree add`s in `CreateAsync` and the `worktree remove` in `RemoveAsync`. Static because `RemoveAsync` is static, so the gate must be shared across instances; keyed on the normalised full path under the platform's path-case rules (mirroring the existing `FileSystemPathComparison`).

The slow network `fetch` deliberately stays **outside** the gate: it's already collision-free via the per-worktree ref, and holding a lock across a download would serialise concurrent launches for no benefit.

## Regression coverage — and why it's a direct test

The race window is too narrow to reproduce on demand, so an end-to-end repro would be a flaky guard. Instead `WorktreeMetadataGateTests` pins the guarantee itself (the helper is `internal` for this reason):

1. a second mutation on the **same** repo waits for the first;
2. mutations on **different** repos are not serialised — catches a global-instead-of-per-repo gate;
3. **equivalent spellings** of one path (trailing separator, `.` segment) share a gate — catches mis-keying;
4. the permit is **released when the mutation throws** — a leak would hang every later launch on that repo.

**Mutation-tested:** stubbing the gate to a pass-through turns exactly tests 1 and 3 red (2 and 4 correctly stay green — they cover different properties, and both fail by timeout if broken). So the negative assertions aren't vacuous.

## Verification and its limits

- `WorktreeMetadataGateTests` 4/4, `WorktreeManagerTests` 10/10; daemon builds 0 warnings / 0 errors; `check-linear-ids.sh` passes.
- **I could not reproduce the original race locally:** 12/12 runs of the concurrent test pass on `main`, and local git 2.49 tolerated three hand-built partial sibling states (empty entry dir, `gitdir` without `commondir`, empty `commondir`) — all exit 0. So I'm not claiming the precise git-internal trigger, only the errno-0 signature above. CI remains the oracle; the change is strictly *less* concurrency, so it cannot regress correctness.
- **In-process only.** Two separate daemon processes operating on one repo would still race; that would need a cross-process file lock. Out of scope here (one daemon per machine is the norm) and called out on the issue as the follow-up if it ever bites.

### DIFF
--- a/src/Capacitor.Cli.Daemon/Services/WorktreeManager.cs
+++ b/src/Capacitor.Cli.Daemon/Services/WorktreeManager.cs
@@ -144,7 +144,12 @@ public partial class WorktreeManager(DaemonConfig config, ILogger<WorktreeManage
                 current = next;
             }
 
-            return current.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+            // Trim a trailing separator so "/repo" and "/repo/" agree — but never down to "", which a
+            // POSIX root ("/") or a bare drive root would otherwise become. An empty key is the one
+            // dangerous direction: it MERGES every degenerate input onto one gate instead of splitting.
+            var trimmed = current.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+
+            return trimmed.Length > 0 ? trimmed : current;
         } catch {
             return path; // too long / invalid chars: worst case a split gate, never a wrong one
         }
--- a/src/Capacitor.Cli.Daemon/Services/WorktreeManager.cs
+++ b/src/Capacitor.Cli.Daemon/Services/WorktreeManager.cs
@@ -37,16 +37,24 @@ public partial class WorktreeManager(DaemonConfig config, ILogger<WorktreeManage
     /// being read while another add is still writing it, since <c>worktree add</c> does enumerate the
     /// existing entries while creating its own) is a HYPOTHESIS, not something this change proves; the
     /// nonsensical <c>Success</c> errno is suggestive but not proof. The remedy does not depend on
-    /// pinning it down: git takes no lock here, so serialising our own concurrent access is correct
-    /// regardless of which interleaving produced the message. <c>worktree remove</c> and
-    /// <c>branch -D</c> also mutate or read the same tree. The daemon really does run these
-    /// concurrently (a flow launching several reviewers against one source repo) — the same concurrency
-    /// the per-worktree fetch ref below was introduced for.</para>
+    /// pinning it down: whatever synchronisation git does here plainly did not prevent this, so
+    /// serialising our own concurrent access is correct regardless of which interleaving produced the
+    /// message. <c>worktree remove</c> and <c>branch -D</c> also mutate or read the same tree. The
+    /// daemon really does run these concurrently (a flow launching several reviewers against one source
+    /// repo) — the same concurrency the per-worktree fetch ref below was introduced for.</para>
     /// <para>Static because <see cref="RemoveAsync"/> is static, so the gate is shared across
     /// instances. Growth is one semaphore per distinct repo for process lifetime, accepted rather than
     /// evicted: a daemon sees a handful of repos, and removing an entry while a waiter holds it would
-    /// split the gate. In-process only — two daemon processes on one repo would still race, which
-    /// needs a cross-process file lock if it ever bites.</para></summary>
+    /// split the gate.</para>
+    /// <para><b>Known limits.</b> (1) In-process only — two daemon processes on one repo would still
+    /// race; that needs a cross-process file lock. (2) Identity is a canonicalised path, so aliases the
+    /// path layer cannot see through — bind mounts, Windows SUBST/mapped drives, 8.3 short names — still
+    /// split the gate (safe direction: a missed exclusion, never a wrong merge), and the
+    /// case-insensitive comparison can merge <c>Foo</c>/<c>foo</c> on a case-sensitive volume (harmless
+    /// over-serialisation). Filesystem identity rather than a path would close both, and .NET exposes
+    /// none portably. (3) <c>worktree add</c> runs the repo's <c>post-checkout</c> hook while the gate is
+    /// held, so a hook that synchronously asks this daemon for another worktree operation on the SAME
+    /// repo would deadlock. No shipped hook does that; the gated calls must stay non-reentrant.</para></summary>
     static readonly System.Collections.Concurrent.ConcurrentDictionary<string, SemaphoreSlim> WorktreeMetadataGates =
         new(OperatingSystem.IsWindows() || OperatingSystem.IsMacOS()
             ? StringComparer.OrdinalIgnoreCase
@@ -88,16 +96,22 @@ public partial class WorktreeManager(DaemonConfig config, ILogger<WorktreeManage
     /// operation, alongside the <c>worktree</c> command it guards.</remarks>
     static async Task<string> ResolveGateKeyAsync(string repoPath) {
         try {
+            // NOT sourceReadOnly: that sets GIT_CONFIG_NOSYSTEM=1, so a repository trusted only through
+            // a SYSTEM-scoped safe.directory would fail this probe while the mutation it guards
+            // succeeds — silently demoting us to a checkout-path key and re-splitting the very gate this
+            // resolution exists to unify. rev-parse is a read, so it needs neither the maintenance
+            // suppression nor the lock avoidance that flag also carries.
+            //
             // --path-format=absolute needs git 2.31+; on older git this exits non-zero and we resolve
             // the (possibly relative) plain form against the checkout instead.
             var absolute = await RunGitCaptureResult(
-                repoPath, GitTimeout, sourceReadOnly: true, "rev-parse", "--path-format=absolute", "--git-common-dir");
+                repoPath, GitTimeout, sourceReadOnly: false, "rev-parse", "--path-format=absolute", "--git-common-dir");
 
             if (absolute.ExitCode == 0 && absolute.Stdout.Trim() is { Length: > 0 } fromAbsolute)
                 return NormalizePathKey(fromAbsolute);
 
             var plain = await RunGitCaptureResult(
-                repoPath, GitTimeout, sourceReadOnly: true, "rev-parse", "--git-common-dir");
+                repoPath, GitTimeout, sourceReadOnly: false, "rev-parse", "--git-common-dir");
 
             if (plain.ExitCode == 0 && plain.Stdout.Trim() is { Length: > 0 } fromPlain)
                 return NormalizePathKey(Path.IsPathRooted(fromPlain) ? fromPlain : Path.Combine(repoPath, fromPlain));
--- a/src/Capacitor.Cli.Daemon/Services/WorktreeManager.cs
+++ b/src/Capacitor.Cli.Daemon/Services/WorktreeManager.cs
@@ -30,6 +30,49 @@ public partial class WorktreeManager(DaemonConfig config, ILogger<WorktreeManage
             ? StringComparison.OrdinalIgnoreCase
             : StringComparison.Ordinal;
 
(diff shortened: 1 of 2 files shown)


realtonyyoung:
Codex static review found **two blocking issues, both real** — my first attempt was incomplete. Addressed in 4afebe6.

**1. The gate keyed the wrong identity.** I keyed on the *checkout path*, but the metadata being protected is the **shared** `.git/worktrees`. A main checkout and each of its linked worktrees have different paths and one common git dir — so a launch from a linked worktree ran *completely unguarded* against one from the main checkout, i.e. exactly the unguarded add this PR exists to prevent. Now keyed on `rev-parse --git-common-dir` (absolute where git supports `--path-format`, else resolved against the checkout), falling back to the normalised path for a non-repo (the standalone-snapshot case) or a git that refuses both.

**2. The fetch I deliberately left outside the gate could mutate the gated metadata.** A plain `git fetch` ends by running automatic maintenance, and current git's maintenance task list includes `worktree-prune` — so it could touch `.git/worktrees` while another launch was mid-add. Now `-c maintenance.auto=false -c gc.auto=0` on that call (both spellings for older git; `-c` rather than `--no-auto-maintenance`, which needs a newer git). Worth noting I checked whether this was already covered: the existing `maintenance.auto=false` in `NewGitPsi` only applies under `sourceReadOnly`, which a fetch doesn't use — so the finding stood.

Also taken:
- **`branch -D` moved inside the gate** (non-blocking finding). It *reads* worktree metadata — git refuses to delete a branch checked out in any worktree, so it enumerates them — and it's best-effort here, so a concurrent add made it fail silently and leak the branch. One acquisition now covers remove + branch delete.
- **Dictionary growth** documented as deliberate rather than left implicit (a handful of repos per daemon; evicting an entry under a waiter would split the gate).
- **Comment wording** softened: the fatal error is observed, the half-written-`commondir` interleaving is *inferred*. It no longer reads as established fact.

### Both blocking fixes are mutation-proven

| Mutant | Tests that should go red | Result |
|---|---|---|
| gate removed entirely | same-repo exclusion + path-spelling | exactly those 2 of 4 |
| path-based key (pre-fix) | linked-worktree shares a gate | exactly that 1 of 5 |

Each mutant reddens only its own target, so the suite discriminates rather than passing by accident.

Local: gate tests 5/5, `WorktreeManagerTests` 10/10, build clean, `check-linear-ids.sh` ✓. CI was already fully green on the previous commit (including the ubuntu leg where this flake actually manifested) and is re-running now.

Codex also positively confirmed: no permit leak on sync/async exceptions or cancellation, no production path nests the same gate or inverts lock order, holding the gate across the whole add/remove is necessary, and the negative assertions are sound.

realtonyyoung:
Codex re-review: **no blocking issues** — both blocking findings from round 1 resolved. It also positively confirmed no daemon path nests gate acquisition, no other `git worktree` mutations exist in the tree, and common-git-dir resolution covers linked worktrees and path aliases.

Took three of the four remaining items in e033e6a; two were substantive rather than polish:

**1. A real hole in the gate's guarantee.** `RunGitCaptureResult`'s timeout path called `Kill(true)` and threw *without waiting for the process to die* — so a caller holding the gate would release it while a killed-but-still-running git could still be mutating the very metadata the gate protects. Now bounded-waits (`KillGrace`) for the kill to land before unwinding.

**2. My tests could have passed for the wrong reason** — the negative window could elapse while the *second* caller was still resolving its gate key (which spawns git). Fixed at both ends: added a checkout-path → key cache (also removes redundant process spawns, since the gate is taken up to three times per launch), and rewrote the exclusion tests to assert **ordering** — second entered strictly after the release — which holds however slow resolution is. Timeouts are now hang guards only.

**3. Comment wording** — the fatal error is now stated as the observed fact, the half-written-metadata interleaving is explicitly labelled a hypothesis, and the remedy is justified without it (git takes no lock there, so serialising our own access is correct regardless of which interleaving produced the message).

**Not changed:** macOS path comparison stays case-insensitive, matching this file's existing `FileSystemPathComparison`. On a case-sensitive volume that can only over-serialise two distinct repos — never under-protect one.

### The rewrite was vacuous, and mutation testing caught it

Worth recording, because it nearly shipped. My first version of the ordering rewrite used `IsEquivalentTo`, which **ignores ordering by default** — so all five tests passed with the gate deleted entirely. Replacing previously-proven assertions had silently voided the guarantee. Now positional:

| Mutant | Should redden | Actual |
|---|---|---|
| gate removed (original tests) | same-repo + spelling | 2 of 4 ✓ |
| path-based key (pre-fix) | linked-worktree | 1 of 5 ✓ |
| gate removed (rewrite, `IsEquivalentTo`) | 3 exclusion tests | **0 of 5 ✗ — vacuous, caught** |
| gate removed (rewrite, positional) | 3 exclusion tests | 3 of 5 ✓ |

Local: gate 5/5, `WorktreeManagerTests` 10/10, build clean, `check-linear-ids.sh` ✓. CI was fully green on the previous commit (both platforms) and is re-running.

realtonyyoung:
Round 3 addressed in 31d5c4c. The blocking finding was correct, and my comment claiming the daemon doesn't produce path aliases was simply wrong.

**Gate key is now physically canonicalised.** It resolves symlinks/junctions component by component, **re-walking after each substitution** (bounded against cycles) because a link's recorded target can itself sit under a symlinked ancestor. The new test caught precisely that on its first run: on macOS the alias keyed to `/var/…/real` while the real path keyed to `/private/var/…/real`, since `/var` is itself a link to `/private/var`. A single pass would have left the gate split — so the aliasing risk you flagged wasn't theoretical, it reproduced immediately. New test: *a symlinked spelling of one repo shares its gate* (skipped on Windows, where symlinks need elevation).

**Dropped the key cache** rather than documenting around the staleness you flagged. A cache has to assume a path's repository identity never changes, and it can change — a linked-worktree directory reused for another repo, a retargeted symlink, `git worktree repair` moving a common dir. A stale entry hands two callers on one repository *different* gates, i.e. exactly the failure this PR exists to prevent. Cost of dropping it is one extra local `rev-parse` per gated operation, next to the `worktree` command it guards.

**Both overstatements corrected.** The removal comment no longer asserts the "half-torn-down sibling" mechanism as fact. And you were right that the test's "holds however slow" claim was false — it now names the residual assumption (the second acquisition must reach the semaphore inside the settle window; there's no seam to observe "reached the gate") and points at mutation testing as the real evidence.

Local: gate tests **6/6**, `WorktreeManagerTests` **10/10**, build clean, `check-linear-ids.sh` ✓. CI was fully green on `e033e6a` and is re-running.

For the record on the previous round's ordering rewrite: mutation testing caught that my first attempt was **vacuous** — `IsEquivalentTo` ignores ordering, so all five tests passed with the gate deleted entirely. Now positional, and the gate-removed mutant fails exactly the three exclusion tests.

realtonyyoung:
Codex round 4: **no blocking issues.** Addressed in 4f68429, and I'm treating this as the last review cycle — the remaining items are documented trade-offs rather than defects.

**Fixed (the one that mattered):** the common-dir probe ran with `sourceReadOnly`, which sets `GIT_CONFIG_NOSYSTEM=1`. A repository trusted only through a **system-scoped** `safe.directory` would fail the probe while the mutation it guards succeeds — silently demoting us to a checkout-path key and re-splitting the exact gate this resolution exists to unify. `rev-parse` is a read, so it needs neither the maintenance suppression nor the lock avoidance that flag carries.

**Re-verified rather than assumed:** the review correctly noted that dropping the key cache means `WarmGateKey` no longer keeps git out of the timed region, which could have made my ordering assertions vacuous again. Re-ran the gate-removed mutant against the uncached code: **4 of 6 fail — all four exclusion tests — and the other two pass.** So the assertions still bite; the `rev-parse` probe is tens of ms against a 500 ms window. My earlier claim on this PR was gathered with the cache present, so it needed re-earning.

**Documented, not half-solved** — each has no portable fix:
- in-process scope only (two daemons on one repo still race → cross-process file lock);
- identity is a canonicalised **path**, so bind mounts / Windows SUBST / 8.3 names still split a gate (safe direction: a missed exclusion, never a wrong merge), and the case-insensitive compare can merge `Foo`/`foo` on a case-sensitive volume (harmless over-serialisation). Filesystem identity would close both; .NET exposes none portably;
- `worktree add` runs the repo's `post-checkout` hook **while the gate is held**, so a hook synchronously asking this daemon for another worktree op on the same repo would deadlock. No shipped hook does that, and a bounded gate wait would trade an exotic hang for plausible spurious failures under legitimate queueing — so it's recorded as a non-reentrancy constraint instead.

**Nit taken:** "git takes no lock here" overstated the evidence → now "whatever synchronisation git does plainly did not prevent this".

Local: gate **6/6**, `WorktreeManagerTests` **10/10**, build clean, `check-linear-ids.sh` ✓. CI was fully green on `31d5c4c` (both platforms, 4th consecutive green ubuntu leg — the one this flake actually appeared on) and is re-running on `4f68429`.

### Review arc, for whoever picks this up

Rounds 1–3 found four real defects in my own fix, two of which meant it didn't actually close the hole: gate keyed on the checkout path instead of the shared git dir; the fetch I'd excluded could itself run `worktree-prune`; a timed-out git was killed but not awaited, letting a live process escape the gate; and the key was lexical-only, so symlink aliases split it. Separately, mutation testing caught that my own test rewrite was **vacuous** (`IsEquivalentTo` ignores ordering — all five passed with the gate deleted). Nothing here was found by a green test run.

realtonyyoung:
Qodo findings — one taken, one declined with reasons (d8c0ebe).

**Taken: empty gate key.** `"/".TrimEnd('/')` is `""`, so a POSIX root (or a bare drive root) produced an empty key. Worth fixing precisely because it's the one **dangerous** direction in this design: everything else fails toward *splitting* a gate — a missed exclusion — whereas an empty key **merges** every degenerate input onto one. Now keeps the untrimmed form when trimming would empty the string.

**Declined: gate dictionary never evicts.** Codex raised the same growth independently and reached the opposite conclusion — *"an acceptable small tradeoff here; safe eviction would require waiter-aware reference counting"* — and I agree with that read. Evicting an entry that a waiter is about to acquire would **re-split the gate**, i.e. reintroduce the exact failure this PR exists to prevent, so refcount+TTL eviction would add a race to a change whose whole purpose is removing one. What's retained is one small managed `SemaphoreSlim` per *distinct repository* over the daemon's lifetime — a handful in practice, not per launch. It's documented as a deliberate choice on the field itself.

Happy to implement eviction as a **separate** change with its own concurrency tests if it's wanted, but I don't think it belongs inside a concurrency fix.

Local after both: gate **6/6**, `WorktreeManagerTests` **10/10**, build clean, `check-linear-ids.sh` ✓.

PR #423, opened 2026-07-31
[AI-1629] Read transcripts with FileShare.ReadWrite on the final-drain poll

realtonyyoung:
Fixes the top remaining kcap-cli Windows CI flake — and the production hazard behind it. Linear: [AI-1629](https://linear.app/kurrent/issue/AI-1629).

## The bug is in production, not the test

`WatchCommand.WaitForFinalLineCompletionAsync` read via `File.ReadAllTextAsync`, which opens **`FileShare.Read`** — denying `Write` to every other handle for the duration of the read.

All **seven** other transcript reads in `WatchCommand` already pass `FileShare.ReadWrite`, and the file states the reason at line 3233:

> `Opened by the caller with FileShare.ReadWrite (Qodo #291 #1) so the writing agent is never blocked.`

This was the one read that missed it. On Windows file sharing is mandatory, and this poll runs on the **shutdown final-drain path**, re-reading the transcript up to `attempts` times *while the vendor CLI is flushing its last records* — so each read can deny the vendor a write to its own transcript. That is exactly the harm the existing invariant exists to prevent, on the worst possible path.

The tests only surfaced it indirectly: the SUT swallows its own read failure and retries (`catch { }`), so what fails is the tests' **awaited concurrent writer**, which is why it looked like test infrastructure rather than a product defect.

## The filed diagnosis was wrong — corrected before implementing

Reported as `Path.GetTempFileName()` name-pool collisions. The census refutes that: of **28** `GetTempFileName` call sites, **26 never flaked in 14 days**, and all 4 failures landed on the only **2** that write concurrently while the SUT reads. A 65,535-name-pool collision would scatter across all 28, not concentrate on exactly the concurrent ones. Issue title/description corrected; the `GetTempFileName` hygiene point is left there as optional and explicitly **not** bundled here.

## Changes

- `ReadAllTextSharedAsync` / `ReadAllTextShared` helpers (`FileShare.ReadWrite`), used by the final-drain poll.
- Same fix for the **two Kiro sidecar reads** in this file — the CLI writes that `{id}.json` as it runs, so it is the same invariant violation. Found by inspection; unlike the poll, these have **no CI evidence** behind them.

## Tests — and an honest split on what is proven

| Test | Mutant = revert to `File.ReadAllTextAsync` | |
|---|---|---|
| `completes_while_another_handle_holds_the_file_open_for_writing` | **still passes on macOS** | real coverage, but only discriminates on Windows |
| `WatchCommand_never_reads_a_transcript_with_a_write_denying_open` | **fails** ✅ | mutation-proven locally |

The behavioural test is deterministic (it *holds* a writer handle for the whole call rather than racing it), but I verified it cannot discriminate on macOS — Unix has no mandatory sharing, so it passes with and without the fix. That is written into the test so a local green is never mistaken for proof; its real proof is the Windows leg.

Because of that I added the **source-level invariant guard**: no `File.ReadAllText*` call may exist in `WatchCommand.cs` (doc comments excluded — they name the banned API to explain the rule). It fails on every platform if the fix is reverted, and it turns the rule into something enforced rather than remembered. It also asserts the source file exists first, so a layout change fails loudly instead of passing vacuously.

## Verification

`TranscriptReadSharingInvariantTests` 1/1 · `WaitForFinalLineCompletionAsyncTests` 5/5 · `LiveWatchHardeningAcceptanceTests` 5/5 · `WatchCommandTests` 57/57 · build clean · `check-linear-ids.sh` ✓.


realtonyyoung:
Codex review flow: **clean at round 3** (`code-review`, vendor codex, 3 rounds). Summary of what it changed:

**Round 1 — two P2s, both on my source-guard test, both right:**
1. `[CallerFilePath]` is not a reliable checkout locator (deterministic-build path mapping rewrites it; a test assembly can run without its source). I verified it wasn't live here — no `ContinuousIntegrationBuild`/`DeterministicSourcePaths`/`PathMap` in the repo or CI — but fixed the class anyway by embedding the guarded source as a resource.
2. The scan didn't enforce what it claimed: a line-wise `Contains` missed `File.ReadAllTextAsync (path)`, line-split calls, aliases and static imports. Replaced with a whitespace-tolerant whole-text regex over comment-stripped source, renamed off "never reads", and documented the residual bypasses.

**Round 2 — one P2, the sharpest of the three:** `StripComments` can't distinguish a comment from a string literal, so `var scheme = "https://"; var t = File.ReadAllText(path);` would strip the rest of the line and **delete the real violating call, leaving the guard green**. My spelling self-test proved the matcher but not the preprocessing, so it couldn't have caught it.

**I removed the guard rather than iterate a third time.** Roslyn is the correct tool but means adding ~10 MB of `Microsoft.CodeAnalysis` to central package management for a belt-and-braces test; a hand-rolled C#-aware lexer is where subtle bugs live. Three rounds of findings on a *supporting* test, each exposing a new vacuity vector, is the signal to simplify.

**What covers the regression:** the behavioural test — which discriminates only on Windows, exactly where the bug manifests, and Windows is a required check — plus the invariant recorded in `CLAUDE.md` under "Common mistakes to avoid", including the key detail that violations are *invisible on macOS/Linux*. Docs can't pass vacuously.

Codex confirmed the production change correct in round 1 and left it untouched thereafter: `StreamReader(Stream)` keeps UTF-8 + BOM detection, disposal order is safe, the enclosing `catch` still covers open failures, and `FileShare.ReadWrite` matches the file's existing invariant.

**Still open for a maintainer call:** the two Kiro sidecar reads are inspection-only — I can't prove the Kiro CLI writes `{id}.json` concurrently, and they carry no CI evidence (read-only share widening inside existing best-effort try/catch). Happy to split them into a separate PR if you'd prefer this one be strictly the evidenced fix.

PR #440, opened 2026-08-03
[AI-1709] Dedupe MCP registrations and retire the resident node wrapper for kcap mcp

realtonyyoung:
## Problem

Every Claude Code session spawns one stdio pair (resident node wrapper + NativeAOT binary) per kcap MCP server — ~36 MB/pair unswapped, ~2 GB across a heavy multi-session day — a primary driver of the swap exhaustion that degraded the daemon and killed review flows on 2026-08-02. The node half exists only because `kcap.js`'s "exec" is actually spawn-and-wait; a user-scope duplicate registration multiplied it further.

## Fix (per the reviewed spec/plan on AI-1709 — Phases 1–2; merge-tier and idle-exit deliberately deferred)

- **Structural duplicate audit** (`McpRegistrationAudit`, both Claude scopes — top-level and `projects[…].mcpServers`): a user-scope entry is a removable duplicate only when semantically canonical; divergent same-name entries are reported as conflicts and preserved (repo policy: same name ≠ ownership). `ClaudeLauncher` stops propagating plugin-shadowed canonical entries into agent worktrees (plugin-installed-gated; mutation-checked both ways).
- **`kcap daemon doctor` gains an MCP-registrations section** (runs before the daemon-file early return; honors `CLAUDE_CONFIG_DIR`): reports duplicates/conflicts, `--clean` removes only canonical duplicates via atomic rewrite, and a stale-path scan covers the six JSON harness configs + Codex TOML + Claude's config, distinguishing missing-binary ("re-run kcap setup") from differs-from-current-resolution.
- **Generated configs register the native binary path** (`Environment.ProcessPath` with seam + `"kcap"` fallback) — threaded through the JSON writers AND `CodexConfigToml`. `McpMarker` evolved to **v2 per-entry fingerprints** (SHA-256 over canonical JSON; v1 read with legacy semantics, migrates on next record; marker recorded only after the config write commits) so refresh-healing across binary relayouts and uninstall both keep working with absolute paths, while genuine user edits are never clobbered (mutation-checked).
- **npm wrapper retired for MCP entries**: new side-effect-free `bin/resolve.js` (no require-cycle with `kcap.js` — pinned by a require-cache assertion); `refresh.js` patches the shipped plugin `.mcp.json` validate-first with sibling-temp + atomic rename, exactly one warning on failure, six canonical pairs only, stale-absolute re-patch. Exercised via a fake release layout (incl. a space in the path); node tests for all three JS files wired into CI on both matrix legs.

## Tests

Targeted classes green: McpRegistrationAuditTests 14, ClaudeLauncherWriteMcpConfigTests 5, JsonMcpConfigWriterTests 23, McpMarkerTests 13, SetupCommandTests 41, CodingAgentsStepTests 131, six PluginCommand harness classes, McpDoctorSectionTests 6, plus the three node test files (`ok`×3). Mutation-checked: launcher skip (both the skip and its structural gate), refresh patch guards (command + args), Codex fingerprint-mismatch heal guard. Pre-existing macOS failure clusters (CodexConfigTomlTests /var-symlink guard ×20, UninstallCommandTests ×11) proven **identical failing sets on a detached origin/main baseline** — not regressions. AOT publish clean for both csprojs (one IDE0059 in an untouched file, 0-line diff). Full suites delegated to CI.

Not done here (needs the live system): the before/after RSS measurement from the issue's recipe — run it post-merge on a fresh session.

Spec + plan + codex spec-review findings: https://linear.app/kurrent/issue/AI-1709

## Accepted limitation (review finding 5)

`ResolveCliSibling` intentionally has no fallback for daemon-without-sibling-`kcap` layouts: every official layout (npm platform packages, release archives) ships both binaries together, so those layouts are unsupported by design. The failure direction is safe — the registration is merely not recognized as canonical, so the agent worktree keeps the wrapper duplicate; nothing is lost or suppressed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,16 @@ jobs:
       - name: Build solution
         run: dotnet build Capacitor.slnx
 
+      - name: Run npm wrapper tests
+        # Plain-Node assertion scripts for the npm launcher/refresh/resolve
+        # modules (no test framework, no npm install). `shell: bash` gives
+        # fail-fast -e semantics on both matrix legs.
+        shell: bash
+        run: |
+          node npm/kcap/bin/kcap.test.js
+          node npm/kcap/bin/resolve.test.js
+          node npm/kcap/bin/refresh.test.js
+
       - name: Run unit tests
         # --maximum-parallel-tests 1 serializes the entire unit suite.
         # The codebase has many tests that mutate process-global state
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -14,6 +14,7 @@
         <PackageVersion Include="SQLitePCLRaw.bundle_e_sqlite3" Version="3.0.3" />
         <PackageVersion Include="NSubstitute" Version="5.3.0" />
         <PackageVersion Include="Spectre.Console" Version="0.55.2" />
+        <PackageVersion Include="System.Threading.AccessControl" Version="10.0.0" />
         <PackageVersion Include="Tomlyn" Version="2.4.0" />
         <PackageVersion Include="TUnit" Version="1.18.37" />
         <PackageVersion Include="WireMock.Net" Version="1.7.0" />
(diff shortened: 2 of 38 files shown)


realtonyyoung:
NO FINDINGS

PR #470, opened 2026-08-05
Signal the PTY child's whole process group on terminate

### LINKED ISSUE #469: PTY TerminateAsync signals only the leader — a SIGKILL'd codex orphans its code-mode-host children
`UnixPtyProcess.TerminateAsync` sends SIGTERM/SIGKILL to the top-level PID only. The child is a forkpty **session leader** (pgid == pid), and helpers it spawns — codex's `codex-code-mode-host`, MCP servers — live in that group. When the leader dies without forwarding the signal (SIGKILL always; SIGTERM vendor-dependent), the helpers are orphaned to init and keep running.

`ProcessReaper` already solved this for the PID-record fallback path (`kill(-pid, sig)` with pid fallback, leader-anchored); the primary live-stop path never got the same treatment — its doc header explicitly notes the "descendant that outlives a leader" as the deferred containment concern.

Observed while investigating the flow-reviewer teardown report (Linear AI-1742): orphaned `codex-code-mode-host` processes after reviewer stops.

Fix: signal the process group in `TerminateAsync` (group first, pid fallback — the established `ProcessReaper` pattern), only while the leader is unreaped so a recycled pgid is never signalled.

realtonyyoung:
Closes #469. Linear: AI-1742.

## Problem

`UnixPtyProcess.TerminateAsync` sends SIGTERM/SIGKILL to the leader pid alone. The child is a forkpty **session leader** (pgid == pid) and helpers it spawns — codex's `codex-code-mode-host`, MCP servers — live in its group, so a leader that dies without forwarding the signal (SIGKILL always; SIGTERM vendor-dependent) orphans them to init, still running. `ProcessReaper` already signals the group on the PID-record fallback path and documents the descendant-outliving-the-leader case as the deferred containment concern; the primary live-stop path never got the same treatment.

## Fix

`SignalGroup(sig)`: `kill(-pid, sig)` with a `kill(pid, sig)` fallback — the exact `ProcessReaper` pattern — used for both the SIGTERM and the SIGKILL escalation. Signalled only while `HasExited` is false: the leader is unreaped until `CheckExited`'s `waitpid`, so the pgid still names our lineage; once the leader is gone the group id proves nothing and is never signalled (the record/scan reap layers own that case).

## Test

`Terminate_kills_the_leaders_whole_process_group`: a shell leader backgrounds a helper, reports its pid over the PTY, and waits; after `TerminateAsync` the helper must be gone (`kill(pid, 0)` → ESRCH), with a positive control that it was alive before.

**Mutation-checked, with a trap worth recording:** the first version used a plain background `sleep`, and the leader-only-kill mutant *passed* — the kernel sends SIGHUP to the controlling terminal's foreground group when the session leader exits, so the helper died anyway. The shipped helper ignores SIGHUP (`trap '' HUP` surviving an `exec`), which both defeats that accidental cover and matches the shape that actually leaks in production. Leader-only mutant now fails the test; the fix passes it.

Local: targeted unit tests green, integration suite green, AOT publish clean of IL2026/IL3050, Linear-ID guard clean.

### DIFF
--- a/src/Capacitor.Cli.Daemon/Pty/Unix/UnixPtyProcess.cs
+++ b/src/Capacitor.Cli.Daemon/Pty/Unix/UnixPtyProcess.cs
@@ -264,6 +264,12 @@ public sealed class UnixPtyProcess : IPtyProcess {
     /// group id proves nothing and is never signalled; a descendant that outlived the group kill
     /// is the record/scan reap layers' job.</summary>
     void SignalGroup(int sig) {
+        // Same guard as ProcessReaper.SignalGroup: kill(2) gives non-positive pids special
+        // meanings — kill(0) signals the CALLER's own group and kill(-1) broadcasts — so a
+        // zero/negative Pid must never reach either call (this repo has a recorded incident of
+        // exactly that SIGKILLing the test host's group).
+        if (Pid <= 0) return;
+
         lock (_reapSignalGate) {
             if (HasExited) return; // reaped since the caller's check — the ids may be recycled
 
--- a/src/Capacitor.Cli.Daemon/Pty/Unix/UnixPtyProcess.cs
+++ b/src/Capacitor.Cli.Daemon/Pty/Unix/UnixPtyProcess.cs
@@ -231,7 +231,7 @@ public sealed class UnixPtyProcess : IPtyProcess {
             return;
         }
 
-        UnixPtyInterop.kill(Pid, UnixPtyInterop.SIGTERM);
+        SignalGroup(UnixPtyInterop.SIGTERM);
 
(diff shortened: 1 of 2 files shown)


realtonyyoung:
NO FINDINGS

realtonyyoung:
**Qodo finding 1 (missing pid<=0 guard) — fixed in 3bc2e33.** `SignalGroup` now carries the same non-positive-pid guard as `ProcessReaper.SignalGroup`, for the same reason (kill(0) targets the caller's own group, kill(-1) broadcasts; this repo has a recorded Pid=0 incident). Codex round 1's atomicity finding (reap vs group signal) was fixed in a3bfd0e with a shared gate serializing CheckExited's waitpid against signalling; codex re-review verdict on the final state: clean.

PR #477, opened 2026-08-07
Treat the recorded reviewer version as a minimum, not an exact match

realtonyyoung:
Kiro and Gemini are the only version-gated reviewers in the product, and both failed closed the moment the installed build **changed** — in either direction. A routine vendor patch release took the reviewer offline until an operator re-ran `kcap daemon reviewer affirm`.

This reinterprets the recorded value as **the oldest build the daemon will run**. An upgrade needs no action from anyone; a downgrade below the record is still refused; a build later found to be bad is excluded by raising the floor with the same verb.

**No migration.** `DaemonRunner.SeedReviewerAffirmation` already recorded the installed build the first time a reviewer was enabled, so every existing record already holds "the version we certified" — only its interpretation changes. A daemon currently blocked by a vendor upgrade unblocks itself.

### What this gives up, stated rather than hidden

A future vendor build that weakens its own containment is now admitted silently — for Gemini the MCP-allowlist semantics, for Kiro `KIRO_HOME` suppression. `GeminiReviewerCapability`'s class doc argued against exactly this design; that paragraph is **kept rather than deleted**, because it records what the boundary cost and what to reach for if a Gemini release is ever found to have weakened the allowlist. Both risks were already accepted at the operator's first affirmation — what changed is that the acceptance carries forward across upgrades instead of being re-taken every patch.

It is also the second time the strict shape has bitten: a maintainer-curated certified set took the Gemini reviewer offline at `0.54.0`, one patch ahead of certified `0.53.0`, recoverable only by a kcap release. Exact-match affirmation removed the release coupling but kept the treadmill, just moved it onto the operator.

### Three things review turned up

- **Ordinal equality is only sound when both values are in the same domain.** An earlier revision ordinal-compared whenever *either* side failed to parse, which refused a genuine upgrade while **labelling it "below minimum"**. A one-sided parse failure is now its own `Incomparable` outcome, and the `affirm` remedy provably terminates it in both directions (tested).
- **The vendor switches ended in `_ => Allowed`.** A new refusing arm would have been silently **admitted** — a fail-closed gate whose safe direction was whatever nobody thought about. Every arm is now explicit with no discard, and `CS8509` is an error repo-wide so the next arm is a build failure instead. Measured before enabling: **zero** CS8509 warnings across the daemon, Core, the CLI and the tests.
- **The version parse is shared** with `DaemonRunner.CliVersionAllowed` rather than duplicated, because both gates must classify orderability identically — and that shared classification is what bounds this change's scope, since widening what parses here would widen what the Claude certification gate admits.

### Verification

- `ReviewerVersionAffirmationsTests` 46/46 · `KiroReviewerCapabilityTests` 32/32 · `GeminiReviewerCapabilityTests` 31/31 · `ReviewerVersionStoreVendorKeyingTests` 3/3 · `KiroReviewerVersionStoreTests` 8/8 · `DaemonRunnerVersionProbeTests` 11/11
- **Mutation-checked**: reverting the floor to an equality compare reddens 7 tests; restored with the inverse edit, green again.
- AOT publish clean (no IL2026/IL3050) on both the CLI and the daemon.
- Monotonicity is pinned as a property, not just per-arm: every pair the old ordinal rule allowed is still allowed, including values no version parser accepts.

Spec: `docs/superpowers/specs/2026-08-06-ai1776-reviewer-version-floor-design.md` (Copilot spec review, clean after 5 rounds).

Fixes AI-1776

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -7,7 +7,7 @@
              maps to the permissive outcome makes the safe direction of a fail-closed gate whatever
              nobody thought about. The reviewer version gate shipped exactly that bug. Measured before
              enabling: zero CS8509 warnings across the daemon, Core, the CLI and the unit tests. -->
-        <WarningsAsErrors>CS8509</WarningsAsErrors>
+        <WarningsAsErrors>$(WarningsAsErrors);CS8509</WarningsAsErrors>
         <MinVerTagPrefix>v</MinVerTagPrefix>
     </PropertyGroup>
     <ItemGroup>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,6 +2,12 @@
     <PropertyGroup>
         <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
         <NoWarn>IDE0055;IDE1006</NoWarn>
+        <!-- CS8509: a switch expression that does not cover every named enum value. An error, not a
+             warning, because the alternative to exhaustiveness is a discard arm, and a discard that
+             maps to the permissive outcome makes the safe direction of a fail-closed gate whatever
+             nobody thought about. The reviewer version gate shipped exactly that bug. Measured before
+             enabling: zero CS8509 warnings across the daemon, Core, the CLI and the unit tests. -->
+        <WarningsAsErrors>CS8509</WarningsAsErrors>
         <MinVerTagPrefix>v</MinVerTagPrefix>
     </PropertyGroup>
     <ItemGroup>
--- a/README.md
+++ b/README.md
@@ -780,13 +780,19 @@ review is not necessarily you) and defaults off.
 
 Two further things it does *not* do:
 
-- it does not bypass the **build affirmation**. The reviewer's only containment is Gemini's exact-name MCP
-  allowlist, which is a behaviour of the installed build — so a `gemini` version other than the one this
-  daemon affirmed is refused even with the flag on, with a coded error naming both builds. Enabling the
-  reviewer affirms whatever is installed at that moment, so you only meet this after an upgrade; clear it
-  with `kcap daemon reviewer affirm --vendor gemini`. (This replaced a maintainer-curated *certified version
-  set*, which took the reviewer offline on every Gemini release until a new kcap shipped — a build one patch
-  ahead of the certified one made the feature unreachable.)
+- it does not bypass the **minimum version**. The reviewer's only containment is Gemini's exact-name MCP
+  allowlist, which is a behaviour of the installed build, so the daemon records a minimum `gemini` version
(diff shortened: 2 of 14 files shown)


realtonyyoung:
Thanks — going through all four. Two fixed in `22e6a8e`, one was already fixed before the review ran, one I'm declining with evidence.

### 4. `WarningsAsErrors` not compositional — **fixed**, good catch

Right, and it's the finding with the longest tail: a later project-level `<WarningsAsErrors>` would have silently replaced this one and dropped CS8509 enforcement, which is exactly the guard the rest of this PR leans on. Now `$(WarningsAsErrors);CS8509`. Verified the build still treats it as an error afterwards.

### 3. Affirm direction misreported — **already fixed**, before this review ran

Agreed entirely, and it was independently raised in the Copilot code-review flow. Fixed in `3f44d5a`; this review ran against `fa55768`. It's now three-way rather than two, with an explicit unknown branch:

```
Set the minimum for daemon 'X' to kiro-cli Y (was Z); the two do not order as
version numbers, so this may have raised or lowered it.
```

Same reasoning you give: claiming a direction we didn't compute is the across-domains mistake the `Incomparable` arm exists to avoid in the gate itself.

### 2. Overly verbose comments — **partially accepted**

Fair on the clearest instance: the no-discard comment was five lines duplicated verbatim across both vendor files, while the enum's own doc already carried the detail. Cut to two lines in both.

Declining the broader trim. The repo's convention in these files is explicitly *why*-oriented prose — `GeminiReviewerCapability` and `ReviewerVersionStore` both carry multi-paragraph class docs today, and CLAUDE.md's "write self-explanatory code instead" is about restating *what* the code does. The comments kept here record decisions the code cannot express: which of two earlier models was abandoned and why, and — in `TryParseVersion`'s case — a specific instruction not to re-add a `.Trim()`, because doing so silently widens a different vendor's security gate. That one is load-bearing: the reviewer flow caught me making exactly that mistake in this PR.

### 1. `AI-1776` referenced in docs — **declining**

The rule is narrower than applied here. CLAUDE.md says *"DO NOT use Linear issue numbers in **comments**"*, and the enforcing CI check is scoped to C# source:

```
rg -n "AI-[0-9]+" src/ test/ --type cs
```

which passes on this PR. No `.cs` file added here carries an identifier.

For docs the repo convention is the opposite — `docs/superpowers/specs/` already contains **14** spec documents named this way, including `ai1706`, `ai1650`, `ai1703`, `ai1325`, `ai1413`, `ai1505`, `ai899`. Renaming this one would make it the only inconsistent file in the directory. The spec is also the durable design record for a change that deliberately reverses an earlier documented decision, so the traceable link is the point rather than an oversight.

Happy to revisit if the intent is that the docs convention should change — but that's a repo-wide call, not something to do to one file in this PR.

PR #488, opened 2026-08-07
Deliver a borrowed reviewer's result through a daemon-brokered capability

realtonyyoung:
Fixes AI-1797

## The bug

A borrowed Copilot review does all its work correctly and then throws the answer away.

Reviewer session `01144c0060e04a7b8fa6ecbab1ac2752`, verbatim:

```
+00:02  Viewing …/borrowed-d40cc491/src/Capacitor.Cli.Core/ProcessUrlPolicy.cs
+00:02  {"content":"namespace Capacitor.Cli.Core;\n\n/// <summary>What a process should do…
+00:06  submit_review_result → {"message":"MCP server 'kcap-flow-result': Error:
                                internal error handling the request.","code":"failure"}
+00:11  submit_review_result → same
+00:15  submit_review_result → same
+00:19  "The tool is returning an internal error on every attempt. Here are my results directly: …"
```

The sandbox admitted the read, containment held (`Disabled tools: bash, create, edit, …`), and the brokered vendor token authenticated. The reviewer produced a correct answer at +00:19 — then the driver waited out the round's full 12.5-minute timeout, because the result had nowhere to go and the result contract ignores reply text by design.

## Mechanism

`AcpHostedAgentRuntimeFactory` redirects the sandboxed vendor's `HOME` to a per-launch state dir so it cannot reach `~/.copilot` or the keychain. MCP servers are spawned as the vendor's **children**, so they inherit that `HOME` — but `AcpReviewFlowMcp` handed the result channel only `KCAP_URL`, which routes it to `CreateAuthenticatedClientAsync` and a token store that is no longer reachable. The exception was caught and reported as a generic error, with the real reason written to stderr and discarded.

The vendor's own credential was explicitly brokered in one line later. kcap's own MCP children never got the same treatment.

## Approach

A daemon-brokered loopback capability, mirroring `kcap-review-context` — which sits six lines below the bug and already solves exactly this by taking a capability URL instead of ambient credentials.

- `LocalPermissionBridge` serves `POST /{token}/flow-result` and `/{token}/flow-message` on a reviewer grant minted **with** a forwarder, and relays under the daemon's credential (the daemon runs unsandboxed, so its token store resolves).
- The upstream API path is chosen from a **fixed table in the bridge**, never echoed from the request — the caller is a sandboxed vendor child, and letting it name the path would turn an authenticated relay into an open one.
- The forwarder lives on the **grant**, so revoking the reviewer closes the submit path in the same operation that closes the read path. A submit path outliving its reviewer would let a lingering child report into a flow whose participant the server already reaped.
- A grant minted **without** a forwarder 404s rather than exposing an endpoint that could only fail, so the capability is scoped to launches that need it.
- On a borrowed launch the channel gets `KCAP_FLOW_CAPABILITY_URL` **instead of** `KCAP_URL`, mutually exclusive by construction — passing both would leave the broken path reachable as a silent fallback. In capability mode it never constructs an authenticated client at all.
- A borrowed launch with no capability now fails **before launch**, mirroring the existing `kcap-review-context` guard, rather than at delivery after the reviewer has done all its work.

`send_flow_message` is covered too — it shares the channel and would have broken identically.

## Why it wasn't caught

The sandbox's enforcement tests are genuinely thorough: real `cat`/`ls` processes under the production profile, per closed tree, no model-layer refusal accepted as evidence. Every one passed. But they assert only what the reviewer **cannot read** — nothing asserted it could still **report**. Denial coverage and function coverage are independent axes.

It was also unreachable until now: every previous Copilot review ran non-borrowed, where `HOME` is real and the token resolves. This is the first borrowed Copilot run, which only became possible once a daemon token source was configured.

## Testing

TDD throughout — each behavioural test watched failing first, for the stated reason.

- `Borrowed_snapshot_result_channel_uses_capability_not_ambient_credential` — the **absence** of `KCAP_URL` is the load-bearing half
- `Direct_review_result_channel_keeps_server_url` — regression guard pinning the fix to the sandboxed case (passes without the change, labelled as a guard)
- `Snapshot_without_result_capability_fails_before_launch`
- `Reviewer_flow_result_capability_forwards_the_body_then_dies_with_the_token` — includes the post-revoke 404 and asserts the upstream path is the bridge's constant
- `Reviewer_without_a_submit_forwarder_has_no_flow_result_capability`
- `Capability_url_overrides_the_api_root_for_a_borrowed_reviewer` — the deliberately unusable `apiRoot` is what makes it non-vacuous

The new guard initially broke 9 existing borrowed-snapshot containment tests, which is the design working — a delivery-less borrowed launch is now unconstructable. Fixed at the shared fixture helper.

Local: `AcpReviewFlowMcpContextTests` 6/6, `LocalPermissionBridgeTests` 58/58, `McpFlowResultServerTests` 16/16, `AcpHostedAgentRuntimeFactory*` 85 total / 0 failed. `AgentOrchestratorVendorTests` 207 total / 5 failed — **the same 5-failure count on a clean `origin/main` detached baseline**, all from the slow PTY/timeout family with membership varying per run.

## Not in this PR

Two sub-issues noted on AI-1797, both separable: the uninformative error text, and fast-failing a round after terminal channel failures instead of burning the full timeout.

### DIFF
--- a/src/Capacitor.Cli.Daemon/Services/AcpReviewFlowMcp.cs
+++ b/src/Capacitor.Cli.Daemon/Services/AcpReviewFlowMcp.cs
@@ -36,7 +36,7 @@ internal static class AcpReviewFlowMcp {
                 "Borrowed-snapshot review cannot inject kcap-flow-result (missing result capability URL).");
 
         var resultEnv = ctx.IsBorrowedSnapshot
-            ? new AcpMcpServerEnvVar[] { new("KCAP_FLOW_RESULT_URL", ctx.FlowResultCapabilityUrl!),
+            ? new AcpMcpServerEnvVar[] { new("KCAP_FLOW_CAPABILITY_URL", ctx.FlowResultCapabilityUrl!),
                                          new("KCAP_FLOW_AGENT_ID", ctx.AgentId) }
             : [new("KCAP_URL", ctx.ServerUrl!), new("KCAP_FLOW_AGENT_ID", ctx.AgentId)];
 
--- a/src/Capacitor.Cli.Daemon/Services/AcpReviewFlowMcp.cs
+++ b/src/Capacitor.Cli.Daemon/Services/AcpReviewFlowMcp.cs
@@ -24,9 +24,24 @@ internal static class AcpReviewFlowMcp {
         var channelName = ctx.LaunchIdentity?.ResultChannelWireName
                        ?? KcapMcpRegistry.ReservedResultChannelId;
 
+        // A borrowed launch reports through a daemon-minted loopback capability, NOT through
+        // KCAP_URL. The two are mutually exclusive on purpose. KCAP_URL is what sends the channel to
+        // HttpClientExtensions.CreateAuthenticatedClientAsync, which loads the token store out of
+        // HOME — and a borrowed launch's HOME is the per-launch state dir, so that load cannot
+        // succeed. Passing both would leave the broken path reachable as a silent fallback; the
+        // observable was a generic "internal error handling the request" with the real reason on
+        // stderr, after the reviewer had already done all its work.
+        if (ctx.IsBorrowedSnapshot && string.IsNullOrWhiteSpace(ctx.FlowResultCapabilityUrl))
+            throw new InvalidOperationException(
+                "Borrowed-snapshot review cannot inject kcap-flow-result (missing result capability URL).");
+
+        var resultEnv = ctx.IsBorrowedSnapshot
+            ? new AcpMcpServerEnvVar[] { new("KCAP_FLOW_RESULT_URL", ctx.FlowResultCapabilityUrl!),
+                                         new("KCAP_FLOW_AGENT_ID", ctx.AgentId) }
+            : [new("KCAP_URL", ctx.ServerUrl!), new("KCAP_FLOW_AGENT_ID", ctx.AgentId)];
+
         var servers = new List<AcpMcpServerSpec> {
-            new(channelName, ctx.CapacitorPath, ["mcp", "flow-result"],
-                [new("KCAP_URL", ctx.ServerUrl!), new("KCAP_FLOW_AGENT_ID", ctx.AgentId)])
+            new(channelName, ctx.CapacitorPath, ["mcp", "flow-result"], resultEnv)
         };
 
         foreach (var id in allowlistServerIds) {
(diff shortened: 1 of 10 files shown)


realtonyyoung:
Both Qodo findings addressed in `98f4fd89`.

**🐞 Unbounded submit body read — real, fixed.** The POSTing process is a sandboxed vendor child, so an unbounded `ReadToEndAsync` is a memory-exhaustion lever on the daemon. I'd flagged this as a question to reviewers myself and Qodo confirmed it.

The cap is enforced on **bytes actually read**, deliberately not on `Content-Length` — a chunked or hostile client need not send that truthfully, so a header-only check would be a guard the attacker controls. One byte past `MaxSubmitBodyBytes` (1 MiB) rejects with 413, so an oversized body is never fully buffered and the forwarder never runs for it.

Two tests, both watched failing first:
- `Oversized_submit_body_is_rejected_without_reaching_the_forwarder` — asserts 413 **and** that the forwarder was not invoked
- `Submit_body_just_under_the_cap_still_reaches_the_forwarder` — so the guard can't pass by rejecting everything, which is the failure mode a bare size check invites

**📘 Verbose comments — fair, trimmed.** All three cited blocks cut back to the non-obvious rationale. I've kept the security "why" (why the upstream path is bridge-chosen, why the forwarder lives on the grant, why `KCAP_URL` and the capability are mutually exclusive) since those are the parts a future edit could silently break, and dropped the restatement of what the code plainly does.

`LocalPermissionBridgeTests` 60/60, `AcpReviewFlowMcpContextTests` 6/6, `McpFlowResultServerTests` 16/16.

realtonyyoung:
**Copilot review flow: clean at round 2** (2 rounds, `applied_reviewer_vendor: copilot`, context-only). Round 1 raised three actionable items; all resolved without code changes, and the reviewer accepted each on round 2.

**1. Unverified outer catch in `HandleAsync` — verified, no hang.** This was my own open question in the PR description. `HandleAsync`'s body is wrapped in a `try` whose outer `catch` sets 500 and closes the response, itself wrapped in a `try/catch` for the already-committed case. The new POST branch is inside it, so a throwing `forward(...)` yields 500 + Close rather than an abandoned context. The PR description's caveat is now discharged.

**2. `new HttpClient()` never disposed — misreading, no leak.** The declaration is at `RunAsync` scope (line 68), outside the dispatch closure, and `client ??=` (line 114) constructs exactly one instance per process, not one per submit. That's a per-launch stdio server, so it's a single long-lived reused client — the recommended pattern. The pre-existing non-borrowed arm has the identical shape and lifetime. The reviewer accepted this on being shown the scope.

**3. Percent-encoded paths 404 — correctness-only, left as-is.** Traced for the record: the switch arm is built from the *same* substring the lookup keys on, so no encoding can make the extracted segment collide with a registered 32-char-hex token. Extra segments fall through the same way.

Also acknowledged and deliberately not fixed: a TOCTOU window between the grant lookup and the awaited relay. Closing it client-side would require revocation to share a lock with an in-flight upstream call, making reaping block on an arbitrarily slow request — worse for a path whose job is prompt reaping. The upstream API validates run state on every submit, so a late submission is rejected there.

Prior round already addressed both Qodo findings (unbounded body read → byte cap; verbose comments → trimmed).

PR #489, opened 2026-08-07
Snapshot submodule contents as plain files instead of refusing the repo

realtonyyoung:
Fixes AI-1794

## The problem

The `IndependentSnapshot` borrowed-review lane refused any repository whose index carried a gitlink:

```csharp
if (SplitNulRecords(initialIndex).Any(entry => entry.Span.StartsWith("160000 "u8)))
    throw new InvalidOperationException("borrowed_snapshot_submodules_unsupported");
```

`kcap-server` pins `src/cli` as a submodule, so **our own primary repo could not be borrowed-reviewed by any snapshot vendor**. Measured effect across vendors on a kcap-server changeset:

| vendor | containment | outcome |
|---|---|---|
| codex | borrows in place | ✅ sees the real tree |
| cursor | `IndependentSnapshot` | ❌ this guard |
| copilot | `IndependentSnapshot` | ❌ this guard |
| claude | none | ⚠️ falls back to an owned worktree — a fresh clone, so the submodule is at the **stale pin**. It does not error; it reviews code the branch never wrote and returns clean. |

## Approach

Submodules are snapshotted as **plain content**. Borrowing means the reviewer sees the developer's working tree, and for a submodule that is the checked-out files — not the pinned commit a fresh clone gives. Dirty and untracked files inside a submodule are carried for the same reason they are in the superproject.

Its `.git` is never copied: that is a gitlink *file* pointing into the superproject's `.git/modules`, which the snapshot deliberately does not have, so copying it would leave a dangling reference. The submodule gains no git identity inside the snapshot.

## Security-relevant details

This lane has history (AI-1675 was a credential-exfiltration-via-symlinks finding), so the new path source deliberately gets no shortcuts:

- **Every submodule path is prefixed to be relative to the SUPERPROJECT before it enters the manifest loop**, so it flows through the identical containment, symlink, reserved-path, encoding and capacity guards. A submodule is not a second, weaker lane. Notably `EnsureNoLinkedComponents` now walks the submodule directory component too, so a symlinked submodule dir is caught.
- **Prefixes stay raw bytes** until that loop's own decode, preserving its classify-before-decode guarantee — a non-UTF8 submodule path can't slip past `ClassifyReservedPath`.
- **The submodule directory is resolved through `ContainedPath` before its git is invoked**, so an index-supplied path cannot steer `git -C` outside the source tree.
- An **uninitialized** submodule (nothing checked out) is skipped rather than failing the snapshot.

## Testing

`BorrowedSnapshot_CarriesSubmoduleFilesAsPlainContentWithoutItsGit` — watched failing first with exactly `borrowed_snapshot_submodules_unsupported`. It asserts the submodule's **dirty** tracked file and its **untracked** file both arrive (so it can't pass on a pinned-commit checkout), and that no `.git` exists at the submodule path in either file or directory form, while the superproject's own independent `.git` still does.

Worth noting: the guard being removed had **no test at all** — `rg submodules_unsupported test/` returned nothing.

Local: `WorktreeManagerTests` 22 total / 0 failed, `BorrowedReviewContextTests` 19/0, `AgentOrchestratorVendorTests` 207/0.

## Out of scope

Option 4 from the issue — making the **claude** owned-worktree fallback refuse or warn on a submodule repo instead of silently reviewing at the stale pin — is deliberately not in this PR. It's a different mechanism (no snapshot involved) and separable, but it is the more dangerous half of the original report, since it fails silently rather than loudly. Worth its own issue.

### DIFF
--- a/src/Capacitor.Cli.Daemon/Services/WorktreeManager.cs
+++ b/src/Capacitor.Cli.Daemon/Services/WorktreeManager.cs
@@ -839,11 +839,9 @@ public partial class WorktreeManager(DaemonConfig config, ILogger<WorktreeManage
                 generation = await CreateReviewContextGenerationAsync(
                     source, reviewContextRoot, sourceHead, initialIndex, caseSensitive, plan, ct);
 
-            // Submodules are snapshotted as PLAIN CONTENT. Borrowing means the reviewer sees the
-            // developer's working tree, and for a submodule that is the checked-out files — not the
-            // pinned commit, which a fresh clone would give instead. Their .git is never copied: it
-            // is a gitlink file into the superproject's .git/modules, which this snapshot does not
-            // have, so carrying it would leave a dangling reference.
+            // Submodules snapshot as plain content: borrowing shows the developer's checked-out
+            // files, not the pinned commit. Their .git is never copied — it is a gitlink into the
+            // superproject's .git/modules, which this snapshot does not have.
             var submodulePrefixes = ReadSubmodulePrefixes(initialIndex);
 
             var manifest = await ReadSourceManifestAsync(source, plan, caseSensitive, submodulePrefixes, ct);
@@ -945,7 +943,10 @@ public partial class WorktreeManager(DaemonConfig config, ILogger<WorktreeManage
             // checked BEFORE git is invoked here.
             EnsureNoLinkedComponents(source, subDir, subRel);
 
-            if (!Directory.Exists(subDir)) continue;   // uninitialized submodule: nothing checked out
+            // Gate on the submodule's OWN git, not on the directory: `git submodule` leaves an empty
+            // directory for an uninitialized one, so Directory.Exists is true there and `git -C`
+            // would fail and abort the entire snapshot — the opposite of skipping it.
+            if (!Path.Exists(Path.Combine(subDir, ".git"))) continue;
 
             // Nested submodules are NOT walked. Failing is deliberate: a silently incomplete snapshot
             // is invisible to the reviewer, who would report on source it never saw.
@@ -960,8 +961,6 @@ public partial class WorktreeManager(DaemonConfig config, ILogger<WorktreeManage
         // A stage-only addition has no working-tree bytes to mirror. Skip those exact raw paths
         // before decoding so an unrelated, absent non-UTF8 index entry cannot interfere with the
         // review-context extractor's classify-before-decode guarantee.
-        var deleted = await RunGitCaptureBytes(source, GitTimeout, true, ct,
-            "ls-files", "--deleted", "-z");
         var deletedPaths = deletedRecords
             .Select(static path => Convert.ToBase64String(path))
             .ToHashSet(StringComparer.Ordinal);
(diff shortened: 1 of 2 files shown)


realtonyyoung:
**Copilot review flow: clean at round 2** (context-only). Round 1 found a **real security escape** — the best finding of the run — plus two correctness gaps. All three fixed in `c9e8...` (see latest commit).

## The escape (reviewer's #1/#6)

`ContainedPath` is **lexical** — `Path.GetFullPath` does not resolve symlinks — so a gitlink whose working-tree directory is a symlink passed containment while `git -C` followed it and enumerated a tree outside the source. The per-file symlink guards can't undo a directory-level traversal that already happened.

The test was written first, and its RED output **proves the traversal** rather than merely asserting a message:

```
git ls-files -co --exclude-standard -z failed: fatal: not a git repository (or any of the parent directories): .git
```

git had already been invoked inside the symlink target. Had that directory been a git repo, its files would have been enumerated and copied into the snapshot.

Fix: `EnsureNoLinkedComponents` (the existing per-component reparse-point walk) now runs on the submodule directory **before** git is invoked there.

## Also fixed

- **Non-UTF-8 submodule path** now raises the coded `borrowed_snapshot_invalid_path_encoding` instead of a bare `DecoderFallbackException`.
- **Nested submodules now fail** with `borrowed_snapshot_nested_submodules_unsupported` rather than being silently omitted. The reviewer's argument was the stronger one: silent incompleteness is invisible to a reviewer, who would then report on source it never saw.

## Deliberately unchanged

An **uninitialized** submodule still skips silently. The asymmetry with the nested case is intentional: an uninitialized submodule has no checked-out content in the developer's own tree either, so the snapshot faithfully mirrors what they have — whereas nested content exists and would be dropped. Reviewer agreed on round 2.

`WorktreeManagerTests` 23 total / 0 failed.

realtonyyoung:
All Qodo findings addressed.

**🐞 #1 Uninitialized submodule aborts the snapshot — real, and my bug.** `git submodule` leaves an **empty directory** for an uninitialized submodule, so `Directory.Exists(subDir)` is true, `git -C` there fails, and `RunGitCaptureBytes`'s clean-exit requirement aborts the entire snapshot — the exact opposite of the skip that guard was written to perform. Now gated on the submodule's own `.git` (file or dir) instead of on directory existence. Good catch; the guard read as correct and wasn't.

**🐞 #5 Redundant `ls-files --deleted` — real, removed.** Its output stopped being read when `deletedRecords` replaced it; I left the call behind. Dead git process per snapshot, now gone.

**🐞 #6 Test leaked temp repos — real, fixed.** The submodule test now removes both helper-created upstream/clone pairs alongside the worktree root.

**🐞 #4 Unguarded submodule path decode — already fixed** before this run. Qodo reviewed `190a7a5`, which predates the copilot-review commit; that commit wrapped the decode in the coded `borrowed_snapshot_invalid_path_encoding`. No further change needed.

**📘 #2, #3 Verbose comments — trimmed**, in both the production block and the test's doc comment.

`WorktreeManagerTests` 23 total / 0 failed.

PR #491, opened 2026-08-08
Write down the manual Antigravity IDE certification procedure

realtonyyoung:
## Problem

The README already said Antigravity memory-index certification "needs the **manual IDE procedure**", and AI-1467's acceptance asks for that procedure to live in the repo docs. It was never written anywhere. The single remaining step to certifying this harness was a procedure nobody could follow.

## Why it's still manual

The automated cert exists and is not the gap. It drives a real `agy -p` turn and fails — but not on our side: print mode fires the `PreInvocation` hook, reads our stdout, and **discards the returned `injectSteps`**. Our half is verified by that same run (hook fires, correct payload, inside budget), so the cert is retained as the regression test that will notice upstream honouring it.

Re-ran it today against **`agy` 1.1.11** (previously recorded on 1.1.10): unchanged, still discarded. That version observation is now in the doc.

The IDE is a different runtime sharing the same `~/.gemini` plugin config. Nothing measured says it behaves like print mode, and nothing says it doesn't — only a human driving the GUI can answer it.

## Every command in the doc was executed while writing it

That caught three things a plausible-looking draft had wrong:

- **The nonce must go in the memory's `description`, not its `content`.** The injected index is one `slug: description` line per memory; content is reachable only through a `get_memory` call — so a nonce in `content` cannot be echoed without a tool call, which defeats the question the certification asks. Verified both ways against the live hook: with the nonce in `content` it is absent from the 2.6 KB payload; moved to `description` it appears. The automated harness does the same thing for the same reason.
- **`kcap mcp memory save` / `kcap mcp memory index` do not exist.** `kcap mcp memory` starts the MCP *server*; it is not a command group. Saving goes through the `save_memory` MCP tool.
- **It is `kcap config show`, not `kcap config get`.**

The doc also records a trap worth knowing before anyone debugs an empty result: the `PreInvocation` handler returns **zero bytes with exit 0** when `conversationId` or `transcriptPath` is missing, *before* it consults the memory index. An empty result therefore does not mean "no index" and looks identical to one — it cost a debugging cycle here.

Cleanup is a step rather than a footnote: the harness refuses to start when a stale nonce may be present, so a forgotten one blocks the next cert run. The nonce memory used to validate this doc was archived, and its disappearance from the index confirmed.

## Scope

Docs only — no source, no test, no CLI surface change. Links the README matrix row to the procedure; the row itself stays `pending`, which is still accurate.

Part of AI-1467 — deliberately not a closing keyword, since the IDE observation itself still needs a human at the GUI.


realtonyyoung:
Superseded — the premise was wrong. This documented a manual GUI certification on the theory that the CLI couldn't be certified; a real interactive `agy` session then carried the injected index as a transcript event, certifying the CLI directly. The corrected state (interactive ✅, `-p` ❌ upstream, app untested) lands via the README/test-relabel PR instead. The reusable mechanics found while writing this doc (nonce-in-description, the zero-bytes-on-malformed-payload trap) are recorded on AI-1467.

PR #493, opened 2026-08-08
Unattended reviewers work out of the box; the opt-in gate becomes an opt-out

realtonyyoung:
Gemini, Kiro, OpenCode and Antigravity each required an explicit
`KCAP_<VENDOR>_UNATTENDED_REVIEWER=1` before a review flow could use them. That
opt-in is removed: **unset now means enabled**, and a recognised falsey value
disables.

## Why the gate had to go

**It was bypassable by the person it was supposedly protecting against.** The
reviewer vendor is a caller-chosen parameter. If the concern is "a requester uses
a reviewer to read files under the daemon's uid and gets them back in the
findings," gating `opencode` accomplishes nothing — they request `vendor:
"claude"`, which was never gated and has *more* capability. A control every
caller can route around by picking a different value from a menu is friction, not
security.

| Reviewer | Gated before? | Tool surface |
|---|---|---|
| claude | no | full — `bypassPermissions` |
| codex | no | full — `--ask-for-approval never` |
| cursor | no | full — `--force --approve-mcps --trust` |
| copilot | no | full — `--allow-all-tools` |
| gemini | **yes** | full — `--approval-mode yolo` |
| kiro | **yes** | `fs_read` + `thinking` only |
| opencode | **yes** | read/grep/glob/list only — no shell, write or network |

Two of the four gated vendors run **read-only** reviewers. The strictest policy
was attached to the least dangerous configurations while four full-tool reviewers
ran ungated.

**And it taxed the honest path hardest.** A supervised daemon inherits nothing
from the installing shell, so turning a reviewer on meant editing a service unit
and restarting — to use a feature the operator had explicitly requested. That is
the mechanism behind AI-1755, where Gemini's non-advertisement was diagnosed
through six ruled-out hypotheses before the consent gate turned out to be the
cause.

## What still protects a launch

The per-vendor **version floor** — a different mechanism answering a different
question. Several of these reviewers' containment is a behaviour of the installed
build, and OpenCode's is environment-based, which fails **silently** if a build
stops honouring it (measured: with no `OPENCODE_PERMISSION`, a review-shaped turn
writes files with no prompt). The floor is seeded unconditionally at startup so
it can never block a first launch, and `kcap daemon reviewer affirm` raises it
past a build found to be bad. Remediation, not permission.

## Why an opt-out rather than deleting the variables

Silently **enabling** something an operator had deliberately set to `0` is the
one surprise this must not spring. An unrecognised value resolves to the default
and is reported by name at startup, so a typo cannot quietly take a reviewer
offline — the direction worth being loud about, since the parse can no longer
distinguish "meant to disable" from "meant to enable".

## Verified

On a real daemon with **no consent configuration at all**: all eight vendors
advertised, **none withheld**. Before this, four were withheld.

Targeted suites, all green on the merged tree (main moved under this branch —
`#490` landed in Antigravity's capability files and its tests are passing here
too): `KiroReviewerCapabilityTests` 44/44, `AntigravityReviewerCapabilityTests`
43/43, `GeminiReviewerCapabilityTests` 31/31, `OpenCodeReviewerCapabilityTests`
14/14, `AcpHostedAgentRuntimeFactoryTests` 83/83, `AntigravityReviewerLaunchTests`
55/55, plus the Kiro/Gemini/OpenCode launch suites, `ServiceEnvironmentTests`,
`DaemonReviewerCommandTests` and `UnattendedLaunchPolicyTests`. AOT publish clean
for both projects. No Linear IDs in `.cs`.

Each "disabled daemon" test now sets the flag false deliberately, and new tests
pin the case this change exists for: a **default** daemon launches a reviewer
with no opt-in, and the advertisement seam does not withhold one for lack of
consent.

## Note for reviewers

This widens a security default, so the reasoning above is the thing to attack,
not the diff. In particular: if you think the gate *was* buying something the
version floor does not, that is the finding worth raising.

Fixes AI-1806

### DIFF
--- a/README.md
+++ b/README.md
@@ -737,28 +737,51 @@ kcap daemon consent log -n 50                                    # tail consent-
 
 `show`/`set-default`/`allow`/`deny`/`remove` mutate the policy over the daemon's local socket and require the target daemon to be running; `log` reads `consent-decisions.jsonl` straight off disk, so it works even with the daemon stopped. All six take `--name <n>` like the rest of `kcap daemon`.
 
-> **Two different gates, deliberately.** `daemon consent` authorises an individual launch and defaults to
-> *allow*; the Gemini reviewer flag below is a **capability** gate that defaults to *off* and asks a
-> different question — whether this daemon may run that vendor unattended **at all**. A per-launch deny is
-> not a substitute: the reviewer's containment rests on a behaviour of the installed vendor build, so the
-> safe default has to be off rather than on-until-denied.
+> **`daemon consent` is the gate that authorises an individual launch**, and it defaults to *allow*. The
+> per-vendor reviewer switches below are a different thing — whether this daemon may run a given vendor
+> unattended **at all** — and they now default to *enabled* too. See the note directly below for why they
+> used to be opt-in and no longer are.
 
-#### Gemini as an unattended review-flow reviewer — off by default, and why
+#### Unattended reviewers are enabled by default (they used to be opt-in)
 
-A **hosted** Gemini agent needs nothing beyond the project setting above. Using Gemini as an **unattended
-review-flow reviewer** is a separate, deliberately opt-in decision, because of what an unattended review
-grants:
+Every reviewer vendor your daemon can host is available to a review flow out of the box. Four of them
+— Gemini, Kiro, OpenCode and Antigravity — used to require an explicit opt-in. That was removed, because
+the opt-in did not do the job it appeared to do:
+
+- **The reviewer vendor is chosen by the caller.** Claude, Codex, Cursor and Copilot have never been gated,
+  and each runs with a *full* tool surface — shell and file writes included. So anyone the gate was meant to
+  stop simply asked for one of those instead, with more capability, not less.
+- **It was attached to the wrong end of the risk scale.** Two of the four gated vendors (Kiro, OpenCode) run
+  *read-only* reviewers. The strictest policy was on the most contained configuration.
+- **It taxed the honest path.** A supervised daemon inherits nothing from your shell, so turning a reviewer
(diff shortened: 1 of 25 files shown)


realtonyyoung:
Thanks — the **Service install message wrong** bug was a real find, and in the one surface two review-flow passes (claude + copilot) didn't reach: the `kcap daemon service install` output. Fixed in 4ac9527.

It was the same root cause as a defect I'd already hit twice in this PR — a second reader of the consent value with its own idea of what it means. Since the switches became opt-outs, a captured `KCAP_*_UNATTENDED_REVIEWER=0` **disables** the reviewer, but the install line said "the unattended reviewer it enables stays on" — backwards for the disable case, stale opt-in framing otherwise.

Fixed the way the earlier instances were: one parser. `ReviewerConsent` moved to Core (the CLI install path can't see the daemon project), the daemon's methods became thin forwarders so no call site or test changed, and the install line now classifies through `ReviewerConsent.IsEnabled` and states what the captured **value** does — "DISABLES that reviewer" vs "keeps that reviewer ENABLED (already the default)". Relabelled `Consent:` → `Reviewer:`. Added `ReviewerConsentTests` pinning the classification, including the invariant that the parse and the warning can't disagree.

On the two **verbose docs** rule violations (`OpenCodeReviewerCapability` / `KiroReviewerCapability`): keeping them, as a deliberate call. This is a security-default change and the narrative in those blocks is the load-bearing rationale for it — both the claude and copilot review-flow reviewers relied on that reasoning being present in-place to assess the flip. Relocating it to a design note would satisfy the rule at the cost of the thing that made the review possible. The `ReviewerConsentKeys`/`CarriedConsentFlags` opt-in language you flagged alongside it is corrected.

PR #495, opened 2026-08-08
Authenticate headless runners with a machine credential

realtonyyoung:
Supplies the half of machine credentials that was never built: **a runner obtaining a token from one.**

## The gap

`kcap machine create` could provision a credential and register it, and the server would accept a machine's token. But nothing could get one. Verified by grep before writing a line:

- No `client_credentials` grant anywhere in the CLI — zero hits.
- No call to the token endpoint — zero hits.
- No token-exchange endpoint on the tenant server or the auth proxy.
- **Nothing read `KCAP_CLIENT_ID` / `KCAP_CLIENT_SECRET`.** Every occurrence was `README.md`, `help-machine.txt`, two `Console.Error.WriteLineAsync` lines, and a test asserting the help *mentions* them.

`help-machine.txt` step 3 says *"On the runner, both variables must be in the environment"* — promising behaviour that did not exist. The tests asserted the help **mentions** the variables, which is the same defect class as the dead-on-arrival `kcap machine` found the same day: testing the artifact I was asked to get right instead of the path that has to execute.

## What this adds

**`MachineAuth`** — reads the two variables, resolves the token endpoint.

**`MachineTokenProvider`** — exchanges them via `client_credentials`, caches the bearer **in memory only**, single-flighted behind a semaphore so a burst of callers (hooks, watcher, MCP servers) mints once.

Never the token store: that would cost the property the design rests on — a machine's bearer existing only for the life of the process that needs it — and buy nothing, since `client_credentials` returns no refresh token. "Refresh" here means "mint another", which needs only the credential the runner already has.

**One wiring point** — `CreateClientCoreAsync`, the single place every authenticated CLI call resolves a token:

- *After* the `None` check (a server needing no auth needs no credential).
- *Before* the token-store paths, because on a runner those find nothing and advise `kcap login`, which a runner cannot do.
- Gated on **either** variable being present, not both, so a half-configured runner is told which one is missing rather than getting that same wrong advice.
- **No `UnauthorizedRetryHandler`** — it refreshes through `TokenStore`, and there is no refresh token. A 401 arrives as `rejectedAccessToken`; re-minting is the repair.

A runner needs **no profile**: that path already honours `KCAP_URL`, and `/auth/config` discovery is unauthenticated.

## Two decisions worth scrutinising

**The token endpoint is hardcoded** (`https://signin.kcap.ai/oauth2/token`) with a `KCAP_WORKOS_TOKEN_URL` override — exactly as `AuthProxyEndpoint` hardcodes the proxy with `KCAP_AUTH_PROXY_URL`. It is one value for the whole fleet (a single WorkOS environment and application), and it **cannot** be derived from the tenant's `/auth/config`, because the field that would carry it — `authkit_domain` — is blank on every tenant, so deriving it would yield a broken URL everywhere. Verified live: that host answers this grant with an OAuth2 credential rejection, while `api.workos.com/oauth2/token` 404s.

**A rejection never echoes the response body.** A token endpoint's error body is attacker-influenced and can reflect the request — which contains the secret. The status is the diagnostic; the body isn't worth the risk. There's a test for that, using a stub that reflects the credential back.

## Tests

11 tests, going through the **real HTTP exchange** against WireMock rather than asserting shapes around it:

- Minting posts the correct form and returns the token — **the wire format is asserted**, since that's what would be silently wrong.
- A cached token is reused with no second request.
- A rejected token forces a fresh mint.
- A rejected credential reports a problem that **does not contain the secret**.
- Success with no `access_token` is a failure, not an empty bearer.
- Credential reading: both present; each half missing, naming the missing one; neither.
- **End-to-end: a constructed client actually carries the minted bearer with no profile present.** This is the test whose absence let the previous slice ship non-functional.
- A half-configured runner reports `NotAuthenticated` rather than a silently unauthenticated client.

**Mutation-verified**: disabling the branch (`Intended => false`) fails 4 tests including the end-to-end one, proving the branch is reached rather than merely present.

`MachineTokenResponse` is registered in `CapacitorJsonContext` and read through it explicitly, so the AOT build has no reflection path.

## Not done yet, deliberately

The **live end-to-end** on a real tenant — create a machine, record a session as it, confirm ownership/attribution/visibility, confirm no seat consumed, revoke, confirm the coded refusal. That runs from a dev build before this is tagged. I am not calling this feature done until it has, because a green build and a clean review have already hidden a dead version of it once.

Part of AI-990.


realtonyyoung:
Both open Qodo findings addressed in `196b01f`, with tests.

**1. No automatic 401 re-mint** — real, and the important one. The machine client installed no retry handler (unlike the token-store path), and the common `CreateAuthenticatedClientAsync` doesn't thread `rejectedAccessToken`, so a token revoked mid-life 401'd repeatedly until the cache aged out. New `MachineUnauthorizedRetryHandler` mirrors `UnauthorizedRetryHandler`: on a 401 it re-mints from the credential (no refresh token needed for `client_credentials`) and resends once, installed on the same `autoRetryUnauthorized` terms so the MCP servers' own loops aren't double-retried. Test drives a real request through the constructed client — 401 → re-mint → 200 — asserting exactly one extra mint.

**2. Unsanitized token URL in error strings** — the URL and caught exception message reach stderr, so both now go through the same helpers `RenderUnreachableError` uses: `UnusableUrlDiagnostic.Sanitize` (drops `user:pass@` — a `KCAP_WORKOS_TOKEN_URL` of `https://id:secret@host` would otherwise print the secret) and `StripControlCharacters` (made internal). Test proves userinfo doesn't survive into the `Problem`.

Findings 3 (cache scoping) and 4 (static race) were already resolved by the review-flow round that removed the static entirely and keyed the cache on `(ClientId, TokenUrl)`.

PR #497, opened 2026-08-08
Host Pi agents over pi --mode rpc (PR-1: interactive)

realtonyyoung:
Interactive hosting for the Pi coding agent (`pi --mode rpc`), the last supported harness with no daemon hosting. A bespoke JSONL-RPC runtime peer of the PTY / ACP / Antigravity runtimes — the shape the AI-894 spike recommended (upstream rejected an in-tree ACP mode; the third-party adapter is a capability/supply-chain dead end).

**Stacked on #494** (AI-1465 Pi memory injection) — review/merge that first; this PR's base is `claude-tyoung/ai-1465-pi-memory`, so its own diff is just the 10 hosting commits.

## What

- `PiRpc` — pure, AOT-safe JSONL-RPC protocol layer (parse frames, build `prompt`/`abort`/`get_state` commands, translate Pi events → vendor-neutral `AcpEventEnvelope`s). Wire shapes cross-checked against upstream `rpc.md`.
- `PiRpcProcess` / `IPiRpcProcess` — long-lived stdio child (unlike Antigravity's exec-per-turn): open stdin for the session's life, serialized line writes, bounded stderr diagnostics, safe dispose/terminate.
- `PiRpcHostedAgentRuntime` — `IHostedAgentRuntime` + `IAcpTranscriptSource`. Read pump → bounded transcript channel; a `get_state` ready barrier that resolves `AcpSessionId`/`ResolvedModel` and **always** faults (30s mandatory deadline) so a wedged child can't hang the orchestrator; queued input, escape→abort, turn-idle tracking, graceful stop, user-echo dedup. Never emits `session_ended` (the server owns it).
- `PiRpcHostedAgentRuntimeFactory` + `DaemonConfig.PiPath`/`PiModel` + `KCAP_PI_PATH`/`KCAP_PI_MODEL` + DaemonRunner registration. `SupportsUnattended=false` (reviewer lane is PR-2); refuses review / review-flow / borrowed-worktree launches before spawn.
- `KCAP_PI_PURE=1` dual-capture gate: the hosted spawn sets it and `kcap.ts` stands down, so the daemon-hosted session isn't also live-ingested by the global extension (the `OPENCODE_PURE` pattern).
- Gated hosted live cert (`KCAP_PI_HOSTED_LIVE=1`) — real spawn → ready barrier → live envelope translation → graceful stop — and README hosted-Pi docs.

## Reviews

TDD throughout; per-task reviews + a whole-branch opus review, all clean after fix rounds. Mutation-verified the load-bearing guards (dispose-vs-write race, ready-barrier fault, echo-dedup). AOT publish clean (both projects); no Linear ids in source. Copilot code-review flow: (running — will note result in a comment).

## Deliberate deferrals

- **Reviewer lane → PR-2** (`PiReviewerCapability`, `-e kcap-flow-result.ts` result channel, `--no-extensions --tools read,submit_review_result` clamp, seeded-defect cert). This is why the PR references AI-894 without a closing keyword.
- **kcap-server companion (before rollout):** add `pi` to `Vendors.NoTerminalSurface` or hosted Pi shows an empty Terminal tab (the Antigravity/AI-1412 precedent); likely also `Vendors.TurnScopedInput`.
- **Memory-injection parity:** `KCAP_PI_PURE` stands down the whole extension, so a hosted Pi session does not get AI-1465's SessionStart team-memory block (accepted; a hosted Claude still does via its hook).

## Before merge (human)

Run the live cert once (`KCAP_PI_HOSTED_LIVE=1` + an authenticated `pi`), ideally after extending its prompt to trigger one tool call (proves the tool-result text extraction end-to-end); record the observed `pi` version.

AI-894

### DIFF
--- a/README.md
+++ b/README.md
@@ -1262,7 +1262,7 @@ KCAP_CURSOR_PATH=/opt/cursor/bin/cursor-agent kcap daemon
 KCAP_CURSOR_MODEL=claude-opus-4-8 kcap daemon
 ```
 
-`KCAP_COPILOT_PATH` overrides the `copilot` binary the daemon spawns for **GitHub Copilot hosted agents** (`copilot --acp --stdio`), mirroring `KCAP_CURSOR_PATH` — the daemon hosts Claude, Codex, Cursor, Copilot, Kiro, Gemini, OpenCode and Antigravity. `KCAP_GEMINI_PATH` overrides the `gemini` binary the same way (`gemini --experimental-acp`), and applies to both hosted Gemini agents and the [unattended Gemini reviewer](#unattended-reviewers-are-enabled-by-default-they-used-to-be-opt-in) (enabled by default) — whose build-affirmation check reads whichever binary it names. `KCAP_OPENCODE_PATH` overrides the `opencode` binary the daemon spawns for **OpenCode hosted agents** (`opencode acp`) — no longer reserved; see [Hosted OpenCode agents](#hosted-opencode-agents) below.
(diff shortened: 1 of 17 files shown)


realtonyyoung:
All three Qodo findings addressed in 25987eed (scoped-re-reviewed, 60/60 Pi tests + AOT clean):

1. **stderr leaked via exception → server** (the important one): `DescribeLaunchFailure` no longer embeds `process.Diagnostics` in the propagated exception — since `AgentOrchestrator` forwards `ex.Message` to `LaunchFailedAsync`, that would have sent stderr (prompt fragments, paths, auth detail) off-host, defeating the process layer's own never-log-stderr boundary. Stderr is now logged **daemon-local** (Warning) and the propagated exception carries only `{reason} (stderr captured in daemon log)`. This now matches Antigravity's `DescribeLaunchFailure`, which classifies stderr into fixed non-sensitive templates and never appends the raw text. Good catch — this was introduced by the very fix that surfaced stderr for diagnosability.

2. **Unconfirmed model reported as resolved**: `ResolvedModel` is now null when `get_state` omits a model (the `IAcpTranscriptSource` "null ⇒ vendor default applies" contract) rather than echoing the requested model as if confirmed. The pump's transcript-envelope display fallback (`_resolvedModel ?? _requestedModel`) is unchanged. PR-2 (the model-selection lane) will revisit this when `set_model` lands.

3. **Direct JsonValueKind checks**: routed through `JsonElementExtensions` — added `IsString`/`IsArray`/`IsNull`/`Bool` helpers (behavior-identical to the inline checks they replace; `Bool` resolves only True/False JSON kinds, never a `"true"` string) and used them across PiRpc + the runtime.

PR #501, opened 2026-08-09
CLI telemetry and the signup-funnel measurement gap

### LINKED ISSUE #500: CLI telemetry to close the signup-funnel measurement gap
## Problem

Two PostHog vantage points exist today and neither sees the middle of the signup funnel.

**Web** fires `cli_snippet_copied` when a visitor copies the install snippet. The snippet is the static string `npm install -g @kurrent/kcap && kcap setup`, so the web person's trail ends at the clipboard.

**Server** emits `cli_setup_completed`, `user_registered` and friends — but only once a tenant exists.

Between "copied the command" and "a tenant exists" there is nothing. Someone who runs `kcap setup`, signs in, discovers they have no workspace, is offered one, and quits — the population that matters most for the signup funnel — is invisible from both ends. The CLI has no telemetry at all.

## What's needed

CLI-side telemetry whose primary job is making that middle segment measurable, and whose secondary job is reporting which commands and MCP tools people actually use.

The drop-off measure should be computable as `cli_setup_tenant_none` minus `cli_setup_workspace_provisioned`, split by the last step reached.

## Constraints

- Must work with **no tenant, no login, and no `server_url`** — that is precisely the segment being measured, so it cannot route through the user's own Capacitor server.
- Must honour `DO_NOT_TRACK`, plus a kcap-specific env var and a persisted opt-out.
- Must never record command arguments, file paths, repo names, session ids, or transcript content.
- Must never throw — an exception escaping to the NativeAOT runtime aborts the process with SIGABRT.
- Hooks are excluded: `kcap hook` runs on every tool use of every recorded session, inline in the agent's critical path.

alexeyzimarev:
Closes #500 · AI-1824

## Why

We can see a visitor copy `npm install -g @kurrent/kcap && kcap setup`, and we can see a workspace appear on the server. We cannot see anything in between — including the person who runs `kcap setup`, signs in, finds they have no workspace, is offered one, and quits. That population is the whole point of the signup funnel and it was invisible from both ends.

This adds CLI-side telemetry. The drop-off measure is `cli_setup_tenant_none` minus `cli_setup_workspace_provisioned`, split by last step reached.

## Design decisions worth knowing

**Direct to `phog.kurrent.io`, not via the user's server.** During the segment this exists to measure there *is* no server — no `server_url`, no tenant, no token. Anything routed through the server cannot observe the pre-server funnel.

**The CLI person stays anonymous.** An anonymous device id in `telemetry.json`, deliberately separate from `machine.json` (which is an auth identifier sent to the server). Where a workspace is known, events join the server's existing `organization` group — but **SaaS only**, because the Helm chart guarantees `Tenant__Name == slug` for `{slug}.kcap.ai` and nothing guarantees it self-hosted. Deriving a group there would look joined and not be.

**Funnel steps flush eagerly, mid-command.** The cohort being measured abandons setup and never runs `kcap` again, so anything deferred to a later invocation is lost. Everything else flushes once from a `ProcessExit` handler under a 1.5s budget; a failed flush spills to a bounded drop-oldest spool that the next successful flush replays.

**Hooks emit nothing.** `kcap hook` runs on every tool use of every recorded session, inline in the agent's critical path. MCP is instrumented per *tool call* instead, which is where recap and memory usage actually shows up.

## Opt-out

`KCAP_TELEMETRY` > `DO_NOT_TRACK` > `kcap config set telemetry off` > enabled.

`KCAP_TELEMETRY` outranks `DO_NOT_TRACK` **in both directions** — it is the kcap-specific deliberate statement and the only way someone with a blanket `DO_NOT_TRACK` can opt back in. That will surprise people, so it is documented explicitly. `kcap config show` reports the effective state and which setting decided it. A one-time notice prints to stderr on first run.

The `telemetry` key is machine-scoped, not profile-scoped — a profile switch silently re-enabling reporting would be a dark pattern.

## Where to look first

- `CommandEvents.cs` — the redaction boundary. Subcommands and verbs come from allowlists; flag *names* pass a shape rule whose 37-char bound is load-bearing (a GUID token is 38, the longest real flag is 31, and both edges are pinned by tests).
- `PostHogPayload.OrgGroup` — the SaaS-only group derivation.
- `WorkOSDiscovery.cs` — the signin events are anchored *inside* discovery, not on its return, because `RunWithLiveAuthAsync` does signin, enumeration and provisioning internally and its exit code is non-zero for declined offers and the retarget path.

## Notes for the reviewer

- **`McpReviewContextServer` is deliberately uninstrumented.** Its own code says "No backend URL or auth here — never any"; telemetry made it write config it has no authority over and POST to PostHog from a sidecar that runs under a `(deny default)` sandbox. An integration test enforces this.
- **The unit suite is red on this branch and on `main`.** 54 failures here, 52 at merge-base `5e84bed2d`, with the set churning in both directions. `AgentOrchestratorVendorTests` in isolation fails 3/207 here and 5/207 at merge-base — worse on `main`. All are timing-sensitive orchestrator/PTY/teardown tests in files this branch never touches. See AI-1815.
- **Companion change needed in `kcap-web`:** the privacy policy describes web and server collection only and needs a CLI paragraph.


alexeyzimarev:
Thanks — worked through all six findings. Five were real and are fixed in `804f777d`; one had a correct diagnosis but an incorrect remedy.

**Fixed**

- **P1, opt-out transmitted telemetry.** Confirmed: on a fresh machine `kcap config set telemetry off` initialised telemetry *before* persisting `false`, so it minted a device id, printed the notice, and flushed `cli_first_run` + `cli_command`. The documented opt-out phoned home. Now short-circuited before `Initialize`, with `DiscardAndDisable()` covering the case where a live facade must be torn down mid-command, and `SetEnabled(false)` deletes the device id — which also closes a known gap, since the spec justified a file separate from `machine.json` on the grounds that opt-out could delete the id outright.
- **P2, `KCAP_TELEMETRY=1` couldn't override a persisted `off`.** `GetOrCreateDeviceId` was re-deciding precedence that `TelemetrySettings.Resolve` already owns. Guard removed; `Initialize`'s gate remains the single enforcement point.
- **P2, MCP `ok` always true.** Correct — dispatch catches its own exceptions and returns `isError: true`, so the `false` branch was unreachable and success rates were systematically wrong. Now derived from the result.
- **Torn read could re-enable telemetry.** Fixed by making the write atomic (temp + `File.Move(overwrite: true)`) rather than locking every read, which would put a cross-process mutex on every command startup.
- **Flush budget bounded only the HTTP phase.** Now timed from entry, with the remaining budget passed to the POST and an exhausted budget spilling to the spool instead of starting a doomed request.

**Diagnosis right, remedy wrong**

- **GitHub `tenant_none` counting discovery failures** is real, but the suggested fix — gating on `ErrorMessage` — would have suppressed the event entirely: `TenantDiscovery.RunAsync:29-31` sets `ErrorMessage` for the *genuine* zero-tenant case too. Fixed instead by adding a discriminator to `DiscoveryOutcome` so only the real zero-tenant outcome emits.

For the record, the equivalent WorkOS path was checked and already correct — `WorkOSDiscovery.cs:81-90` returns early on `DiscoveryError`, before the zero-tenant check — so the primary funnel denominator was never affected.

PR #505, opened 2026-08-09
Render budget enforcement disclosure on flow start/status/round surfaces

realtonyyoung:
CLI half of the AI-1827 dynamic-budget work (server: kurrent-io/kcap-server#1381). The server now emits `budget_enforcement` (`full`/`partial`) + `unmetered_roles` on dynamic-run flow responses. `kcap mcp flows` reconstructs its textual envelopes from an allowlist of parsed fields, so a server-only change would reach raw REST callers and nothing else — this is a **required dependent change**.

A shared `AppendBudgetDisclosure` helper renders the pair on every formatting path (start/roundless, round, status, polled round, close), mirroring `AppendWorkspaceDiagnostics`. An absent field (catalog run, budget-irrelevant run, or an older server) renders **nothing** — byte-identical to before, so skew is safe both directions.

Tests (`McpFlowsBudgetDisclosureRenderTests`) go through the real formatters on every surface: partial names the roles, full omits them, and absent/null renders nothing.

Fixes AI-1827.

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,6 +445,16 @@ workspace: unknown
 
 Status and polled round-result output also cross-check the reviewer's **vendor**: if the `reviewer` participant's vendor disagrees with the run-level `applied_reviewer_vendor` (or the server flags `reviewer_vendor_mismatch`), the CLI renders a `⚠ reviewer vendor mismatch` warning telling the driver to treat the run's results as suspect, close the flow, and report it. Agreement renders nothing.
 
+A dynamic (`definition_yaml`) flow whose participants can be dollar-metered also reports its **budget enforcement** level, so you know whether every participant's spend counts against `budget_usd`:
+
+```
+budget enforcement: full
+
+budget enforcement: partial (unmetered roles: probe, helper)
+```
+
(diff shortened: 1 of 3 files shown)


realtonyyoung:
Thanks — qodo findings addressed:

1. **README missing budget enforcement (Maintainability)** — documented the `budget enforcement:` line in the `kcap mcp flows` section (8b0551b), parallel to the existing workspace-disclosure docs: what `full`/`partial (unmetered roles: …)` mean, and that the line is absent for catalog runs / no-dynamic-budget / older servers.

2. **Verbose comment in role loop (Maintainability)** — tightened the inline comment in `AppendBudgetDisclosure` (b01f66b).

3. **Unsanitized role name output (Security)** — `IsRenderableRole` now rejects a blank or control-char-bearing role name (a newline/CR would otherwise forge a line in this line-oriented output), in addition to the non-string/null/empty guards. Pinned by `Role_names_with_control_chars_or_whitespace_are_skipped` (b01f66b).

4. **Role parsing may throw** — already marked ✅ Resolved by the earlier `TryGetValue`/skip hardening (ab37b5c).

PR #507, opened 2026-08-09
[AI-1283] Deterministic update notice + version/preference headers

realtonyyoung:
Part of AI-1283 — the kcap-cli half of "your kcap CLI is out of date" notifications. The kcap-server half (npm poller, per-user version registry, web banner, notification-centre kind, hook `version` field) already merged in kcap-server#1384. This PR makes the local update notice deterministic and starts transmitting the observation headers that server side consumes. Immediately useful on its own; the headers are inert until a server that reads them is deployed (and the server already ships that).

## What changed
- **Deterministic exit notice** — the existing npm update hint was fire-and-forget and only ever awaited inside Claude hooks (where nobody reads it). It now runs in a `try/finally` around the whole dispatch (covering `--help` and the no-server early exits) and prints after every *human-facing* command. Suppressed for agent-spawned/protocol/long-running commands (`hook`, `generate-whats-done`, `set-title`, `copilot-finalize`, `mcp *`, `watch`, the whole `daemon` family) and for `update`/`uninstall`/`--no-update-check`/`update_check=false`.
- **Cache + backoff** — `UpdateCacheRecord` adds failure/backoff fields (backward-compatible with the legacy two-field cache): a failed poll is retained with a 1h backoff so repeated quick commands don't re-hit the network, and the passive path runs under a bounded budget (300ms warm cache / 3s network with a 500ms cleanup grace) so a slow registry can't delay exit or livelock. `kcap update`/`--check` keep the 5s bound and bypass the backoff.
- **`kcap status`** now prints the installed CLI version, with an inline `(update available: X)` when behind — sharing one check with the exit notice (no double print, at most one network call).
- **Observation headers** — every authenticated request (via the `CreateClientCoreAsync` choke point, plus the `whoami` raw probe) carries `X-Kcap-Cli-Version`; `X-Kcap-Update-Check: off` is added only when the user has opted out. These let the server surface the banner / notification and honor the opt-out.
- **Opt-out** — `kcap config set update_check false` is now the full, persisted opt-out: it silences the local notice, the `kcap status` annotation, the transmitted headers (so the server-side banner/notification honor it), and the in-agent nudge. `--no-update-check` stays a one-shot local suppression.
- README + `help-update.txt` + `config` help updated.

## Notes
- Built TDD, task-by-task, each reviewed (incl. mutation-testing the wire-contract headers and the notice-suppression gate).
- No Linear IDs in any `.cs` file (repo lint).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1811,18 +1811,21 @@ kcap logout         # delete stored tokens
 > releases correspond to server versions rolled out to internal tenants first,
 > so most users should stay on stable.
 >
-> **Staying current, and turning it all off.** Beyond the stderr hint above,
-> every human-facing `kcap` command prints an "Update available" notice after
-> it finishes if a newer version exists, and `kcap status`'s **Version** line
-> shows the same thing inline. Every request `kcap` sends also carries its
-> version (and, if update checks are off, an explicit opt-out marker) to your
-> server, which is how the web dashboard's own out-of-date banner and
-> notification-centre entry — and the in-agent nudge Claude Code sessions can
-> see (above) — know to show up. `kcap config set update_check false` (or
-> `--no-update-check` for a single invocation) turns off **all** of it at
-> once: the stderr hint, the exit-time notice, the `kcap status` annotation,
-> the headers that drive the server-side banner/notification, and the
-> in-agent nudge. It defaults to on.
+> **Staying current, and turning it all off.** Every human-facing `kcap`
+> command prints an "Update available" notice after it finishes if a newer
+> version exists, and `kcap status`'s **Version** line shows the same thing
+> inline. Every request `kcap` sends also carries its version (and, if update
+> checks are off, an explicit opt-out marker) to your server, which is how the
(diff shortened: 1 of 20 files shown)


realtonyyoung:
All three qodo findings addressed in a60c100 (docs/comments only, no logic change):

1. **Verbose `UpdateNotice` doc comment** — trimmed the class XML doc to a concise summary (kept only the load-bearing one-liners).
2. **Verbose `StatusCommand` comment** — trimmed the inline block to 1–2 lines; the helper/method names carry the intent.
3. **[Bug] help-text wire behavior** — corrected `help-update.txt` (and reconciled the README callout): with `update_check=false` the CLI still sends `X-Kcap-Cli-Version` and *adds* `X-Kcap-Update-Check: off` — the opt-out is a signal carried on the wire (the server honors it and suppresses its own banner/notification), not an omission. The prior 'omits the headers' wording was inaccurate.

PR #512, opened 2026-08-10
[AI-1283] Report the new CLI version to the server after kcap update

realtonyyoung:
Part of AI-1283 — companion to kcap-server#1389. Together they fix a reported bug: after `kcap update`, the "your kcap CLI is out of date" web banner/notification linger.

## Gap this closes
The server only learns your installed CLI version from an authenticated CLI→server request carrying `X-Kcap-Cli-Version`. **`kcap update` itself makes no such request** (it only talks to the npm registry), so the server didn't re-observe the new version until your next incidental hook/`status`/daemon call — leaving the out-of-date surfaces stale in the meantime.

## Fix
- New hidden `report-version` command: makes ONE quiet, fail-open **authenticated GET to a side-effect-free probe** (`/api/me/notification-prefs`, the same read `whoami` uses) via the header-carrying client, so the server's observer middleware records the new version from the `X-Kcap-Cli-Version` header. Never prints on the happy path, always returns 0, bounded ~5s. Skips silently when not authenticated; proceeds on both `Ok` and `NoAuthRequired` tenants; listed in `offlineCommands` so a no-server host still returns 0.
- The npm wrapper (`bin/kcap.js runUpdate`) invokes the **new** binary's `report-version` after a successful install/refresh and before exit, best-effort (`stdio: "ignore"`, timeout, try/catch) — it can never change the update's exit code.

Deliberately a side-effect-free GET, **not** the `/api/users/me/cli-setup` POST: that endpoint fires a one-time onboarding event, which would falsely mark a login-but-never-`setup` user as "Registered" (caught in review).

## Notes
- No Linear IDs in any `.cs` (repo lint).
- Tests: `ReportVersionCommandTests` 8/8 (header-on-GET, not-authed→no-request, no-auth-tenant, no-server, error/timeout all return 0), observation-header + update-notice suites unaffected.

### DIFF
--- a/npm/kcap/bin/kcap.js
+++ b/npm/kcap/bin/kcap.js
@@ -372,6 +372,20 @@ function runUpdate(binaryPath, updArgs) {
   require("./refresh").runRefreshes(fs.realpathSync(__filename));
   console.log("kcap updated.");
 
+  // Tell the server about the new version right away, so its "CLI out of
+  // date" banner/notification clear immediately instead of lingering until
+  // whatever the user runs next happens to hit the server. `binaryPath` now
+  // points at the freshly-installed NEW binary (overwritten in place above,
+  // or laid down at the same path after Windows's rename-aside). Best-effort
+  // and silent: this must never fail or slow down `kcap update` — the server
+  // observes the new version from the next incidental request anyway if this
+  // doesn't run or doesn't complete.
+  try {
+    execFileSync(binaryPath, ["report-version", "--no-update-check"], { stdio: "ignore", timeout: 8000 });
+  } catch {
+    // fire-and-forget: the server will observe on the next request anyway
+  }
+
   // A running daemon keeps executing the old image until restarted. On
   // macOS/Linux it self-detects the new binary and restarts when idle; on
   // Windows self-detection is off (the running image was moved, not replaced),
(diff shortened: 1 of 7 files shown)


realtonyyoung:
All qodo findings addressed in b4acedb:
1. **[Sec] Wrong-host probe URL** — now computes one `effectiveBaseUrl` with the same fallback chain as the client builder (incl. `KCAP_URL`), passed to both `CreateClientWithAuthStatusAsync` and the probe URL, so the authenticated request and the GET target can't diverge. Test added (null baseUrl + `KCAP_URL` ⇒ GET lands on the KCAP_URL host).
2. **[Rule] Verbose comments** — trimmed the `ReportVersionCommand` doc + the `kcap.js`/`kcap.test.js` blocks to concise form (code unchanged).
3. **[Reliability] Discovery time budget** — a single ~5s CancellationTokenSource now bounds the WHOLE command (provider discovery + the GET), threaded through both; still returns 0 on timeout. Test added (10s-delayed `/auth/config` ⇒ returns 0 within budget).
4. **[Maint] Fragile source test** — slimmed `kcap.test.js` to tolerant invariants (report-version present, ordered after refresh / before exit, inside try/catch) instead of exact string literals.

Issue #511, opened 2026-08-10
Daemon does not reliably pick up a new credential after `kcap login` (live hub pins handshake identity; profile/tenant change never heals)


alexeyzimarev:
Sibling of #509 (which makes a hook 401 tell the user to run `kcap login`). Once they do, a running daemon does not reliably pick the new credential up.

### What is already fine

Worth stating so a fix does not touch it: every daemon HTTP path resolves a token *per operation* through `TokenStore`, which has no in-process cache — `ServerConnection.ProcessEventQueueAsync` (`:1247`), `EvalRunner` (`:50`, `:99`, `:144`), `SpoolDrainLoop` (`:52`), `AgentOrchestrator` (`:262`, `:2824`). Those see a new token on the next call. `TokenRefreshLoop` likewise drives the shared, cross-process-locked `TokenStore.RefreshIfExpiringAsync`, so proactive refresh is not the problem either.

A hub connection that *fails* to authenticate also self-heals: `ConnectWithRetryAsync` retries forever (backoff capped at 30s) and each `StartHubAsync` re-invokes `AccessTokenProvider` (`ServerConnection.cs:192`), which re-reads the store.

### Problem 1 — a live hub connection pins its handshake identity

SignalR invokes `AccessTokenProvider` only at negotiate. kcap-server sets no `CloseOnAuthenticationExpiration` on the hub, so the `ClaimsPrincipal` captured at handshake is pinned for the life of the socket. A daemon whose token expires — or is replaced by `kcap login` — keeps operating as the old identity until something independently drops the transport.

This usually presents as silent staleness rather than an error, which makes it easy to miss: work attributed to the previous identity, or a tenant switch that appears not to take effect. `ForceReconnectAsync` (`:785`) already exists and is driven by `DaemonHeartbeatLoop`, so there is a natural hook — nothing currently triggers it on a credential change.

### Problem 2 — a profile or tenant change never heals, and never explains itself

`TokenStore.GetValidTokensForServerAsync` resolves the **active profile**, then withholds the token entirely unless it is bound to the requested server:

```csharp
if (!BoundToTarget(snapshot, targetBaseUrl)) {
    return new(null, AuthStatus.WrongServer, snapshot.ServerUrl, profile);
}
```

[...]

Issue #510, opened 2026-08-10
A 401'd lifecycle payload is dropped, not spooled, so `kcap login` cannot recover the session


alexeyzimarev:
Every hook path classes an HTTP 401 as a **permanent** failure (`< 500 and not 408 and not 429`) and therefore drops the payload instead of spooling it:

- `ClaudeHookCommand.cs` — the `session-start`, `session-end` and `subagent-stop` bounded-POST arms each compute `permanent` and skip `spool.Append` when it is true.
- `AgentHookPoster.PostOrSpoolAsync` — same rule for Codex, Cursor, Gemini, Copilot, Pi, Kiro and OpenCode.

So when a credential is rejected mid-session, the lifecycle event that hit the 401 is gone. Running `kcap login` afterwards resumes recording from the *next* event but never recovers the dropped one — a dropped `session-start` in particular means the session has no server-side record, while a dropped `session-end` leaves it stuck "Active".

This is the natural companion to the re-login nudge: the nudge tells the user what to do, but by then the event it was nudging about has already been discarded.

### Why it is not a one-line change

The permanent-drop rule is deliberate and load-bearing elsewhere. `AgentHookPoster.cs:208` documents the reasoning for skipping the spool drain entirely on an auth lapse:

> a POST with no bearer token would 401, and `LifecycleSpoolDrain`'s production poster treats a non-timeout/non-5xx status as a permanent drop — which would silently discard the very backlog this protects.

So spooling a 401'd payload without also changing how the *drain* classifies a 401 makes things worse, not better: the next drain pass would attempt it, get another 401, and drop it — converting a visible loss into a delayed silent one. Any fix has to treat 401 as retryable in `HookSpool` / `LifecycleSpoolDrain` as well, across every vendor, and needs a rule for how long a backlog that can only be delivered after a human logs in is retained (the existing 30-day reap is the current backstop).

Deliberately out of scope of the nudge change so that its blast radius stays contained.

PR #513, opened 2026-08-10
Tell the user to run kcap login when a hook is rejected with HTTP 401

### LINKED ISSUE #509: Hook HTTP 401 surfaces as an opaque hook error and never says `kcap login`
A credential that the local token store believes is usable, but that the **server** rejects with 401, produces nothing actionable in the agent. Observed in Claude Code:

```
⏺ Ran 2 stop hooks (ctrl+o to expand)
  ⎿  Stop hook error: Failed with non-blocking status code: HTTP 401
```

That banner does not say recording has stopped, and does not say `kcap login`. Recording stayed silently off until the user ran `kcap whoami` on a hunch and discovered they were logged out.

### Root cause

There are two distinct auth lapses and only one is handled.

| Lapse | How it is detected | Behaviour today |
|---|---|---|
| The local token store knows the credential is dead (`AuthStatus.Expired` / `NotAuthenticated` / `WrongServer`) | pre-flight, before any POST (`ClaudeHookCommand.cs:377`) | exit 0 plus a `systemMessage` nudge — but only on `session-start` |
| The store believes the credential is usable; the server rejects it (401) | only from the response itself | no nudge on any event |

`CreateClientWithAuthStatusAsync` reports `AuthStatus.Ok` whenever `TokenStore.GetValidTokensForServerAsync` returns a locally-valid token, and the hook path builds its client with `autoRetryUnauthorized: false`, so a 401 is final. It falls through to the shared failure arm (`ClaudeHookCommand.cs:786`), which writes the bare string `HTTP 401` to stderr and returns 1 — and Claude Code renders any non-zero, non-2 hook exit as the opaque "non-blocking status code" banner above.

The other events are worse than opaque, they are silent: `session-start`, `session-end` and `subagent-stop` each class a 401 as *permanent* (`< 500 and not 408 and not 429`), drop the payload, and return 0 without a word.

Other vendors (Codex, Cursor, Gemini, Copilot, Pi, Kiro, OpenCode) share `AgentHookPoster`, which prints an equally bare `[kcap] codex-hook stop: HTTP 401`.

### Fix

`systemMessage` on stdout with exit 0 — verified against the installed Claude Code bundle (2.1.226), whose embedded docs describe `systemMessage` as "Display a message to the user (all hooks)" and give the `Stop` event as a worked example.

- `stop` and `session-start` on a 401 → exit 0 and `{"systemMessage": "[kcap] The server rejected your credentials (HTTP 401) — session recording is paused. Run 'kcap login' to resume."}`. Exit 0 is what replaces the hook-error banner with a clean notice.
- Every other event on the shared path (`notification`, `subagent-start`, `pre-compact`) → exit 0 silently. They fire more than once per turn, so nudging from them would stack duplicate notices; `stop` and `session-start` are the once-per-turn user-facing events and need no throttle state on disk.
- Non-401 failures unchanged (bare status on stderr, exit 1). 403 is an authorization decision, not a dead credential — `kcap login` would not fix it, so it is excluded.
- Other vendors cannot carry a `systemMessage` (their stdout is a strict handshake contract the vendor parses), so only the stderr text becomes actionable.

Deliberately **not** `decision: "block"` with a `reason`: that is the only way to hand text to the *agent* from a `Stop` hook, but it makes Claude keep working instead of stopping — an extra model turn per lapsed turn, plus a loop hazard. The user is the one who runs `kcap login`, so the notice goes to the user.

Full design: `docs/superpowers/specs/2026-08-10-hook-401-login-nudge-design.md` (rides the implementation PR).

Follow-up split out separately: a 401'd lifecycle payload is dropped rather than spooled, so logging in does not recover that session.

alexeyzimarev:
Closes #509

AI-1835

## The bug

A credential lapsed mid-session and Claude Code showed only this:

```
⏺ Ran 2 stop hooks (ctrl+o to expand)
  ⎿  Stop hook error: Failed with non-blocking status code: HTTP 401
```

Nothing said recording had stopped, and nothing said `kcap login`. Finding out took a hunch and a manual `kcap whoami`.

There are two distinct auth lapses, and only one was handled:

| Lapse | How it's detected | Behaviour before |
|---|---|---|
| The local token store knows the credential is dead (`Expired` / `NotAuthenticated` / `WrongServer`) | pre-flight, before any POST | exit 0 + a `systemMessage` nudge — but only on `session-start` |
| The store thinks it's usable; the **server** rejects it (401) | only from the response | no nudge on any event |

This was the second. `CreateClientWithAuthStatusAsync` reports `AuthStatus.Ok` for any locally-valid token, and the hook path builds its client with `autoRetryUnauthorized: false`, so the 401 was final — it fell to the shared failure arm, which wrote a bare `HTTP 401` to stderr and returned 1. Claude Code renders any non-zero, non-2 hook exit as that opaque banner. The other events were worse than opaque: `session-start`, `session-end` and `subagent-stop` class a 401 as *permanent*, drop the payload, and return 0 without a word.

## The fix

A 401 becomes a recognized outcome instead of a generic failure.

- **`stop` and `session-start`** exit **0** and write `{"systemMessage": "[kcap] The server rejected your credentials (HTTP 401) — session recording is paused. Run 'kcap login' to resume."}`. Exit 0 is load-bearing: it's what replaces the hook-error banner with a clean notice.
- **Other events on the shared path** (`notification`, `subagent-start`) exit 0 silently. `notification` fires on every permission prompt, so nudging there would stack duplicate notices inside one turn. No throttle state on disk is needed as a result.
- **Non-401 failures are untouched** — a 500 keeps its bare-status stderr line and exit 1, with a characterization test guarding that.
- **The seven non-Claude vendors** (Codex, Cursor, Gemini, Copilot, Pi, Kiro, OpenCode — and Antigravity, which shares the seam) can't carry a `systemMessage`: their stdout is a strict handshake contract the vendor parses. Their stderr line now names the fix instead. Both `AgentHookPoster` entry points were edited, so it applies whichever one a vendor uses.
- All notice wording moved into one Core type, `AuthLapseNotice`, so the pre-flight nudge and the server-rejection nudge can't drift apart.

`decision: "block"` was deliberately **not** used. It's the only way to hand text to the *agent* from a `Stop` hook, but it stops Claude from stopping and costs an extra model turn per lapsed turn. The user is the one who runs `kcap login`, so the notice goes to the user.

401 only — a 403 is an authorization decision, not a dead credential, and `kcap login` wouldn't fix it.

## Deliberately out of scope

- **A 401'd lifecycle payload is dropped, not spooled** (#510). Because 401 is classed permanent, the `session-start`/`session-end` event that hit the 401 is already gone — logging in doesn't recover it. Making 401 retryable means changing the drop rule in `HookSpool`/`LifecycleSpoolDrain` for every vendor, and it contradicts the existing reasoning at `AgentHookPoster.cs:212`. Separate change, real blast radius.
- **The daemon doesn't reliably pick up a new credential after `kcap login`** (#511). A live SignalR hub pins its handshake identity, and a profile/tenant change yields `WrongServer` forever without explaining itself.
- **No token-recovery retry on 401.** `GetValidTokensForServerAsync` already refreshes proactively, so a 401 after it means the server rejected a *fresh* token — a revoked session or an org mismatch, neither of which a refresh heals — and WorkOS refresh tokens are single-use.

## Tests

New coverage: `stop`/401 → exit 0 + notice; `session-start`/401 → exit 0 + notice; `notification`/401 → exit 0, no notice; `stop`/500 → exit 1, no notice; the vendor 401 stderr line (and its non-401 bare-status counterpart); and the pre-flight lapse arm, which turned out to have had **no test at all** despite the design assuming one — including the `WrongServer` → "not authenticated" mapping.

`ClaudeHookCommandTests` 44/44, `AgentHookPosterTests` 9/9, `AuthLapseNoticeTests` 5/5, integration 217/217. `dotnet publish -c Release` shows no IL3050/IL2026.

The wider unit suite has ~63 failures on `main` itself in a local macOS environment (Codex `config.toml`, `kcap uninstall`, CLI-runner flood tests, daemon teardown/quarantine timing), churning in both directions run-to-run; this branch measured 62-66 with none in hook, 401, or auth-lapse tests. CI is the authority on those.

## One thing a reviewer should know

The design rests on Claude Code honouring `systemMessage` for the `Stop` event. That's documented in the installed CLI bundle ("systemMessage — Display a message to the user (all hooks)", with a worked `Stop` example), and kcap already ships a `systemMessage` write on the `session-start` path. It has **not** been confirmed end-to-end interactively, because headless `-p` doesn't fire `Stop` hooks at all. If the contract doesn't hold for `Stop`, the failure is silent — the user sees nothing where they previously saw an opaque banner. The `session-start` half is an improvement either way, since it was previously silent.


alexeyzimarev:
## Bot review responses

### ✅ Codex — "Emit the login advice from Cursor hooks" (P2): valid, fixed in 0e0d1507

This was a real gap and the reasoning behind it was mine, not the implementation's. `CursorHookCommand` uses `AgentHookPoster` only for the `IsAuthLapsed` predicate (`:128`) and POSTs directly via `TryPostHookAsync` (`:893`) plus its own spool-drain lambda (`:373`), both of which turned a non-success status into `false`/`DrainOutcome.Drop` with nothing on stderr. So Cursor would have been the single vendor left with no explanation and no recovery step — precisely the failure this PR exists to remove.

For the record on how it got through: an earlier task review raised exactly this as a "cannot verify from diff" item and asked for a spot-check of the vendor call sites. I dismissed it on the grounds that *both* poster entry points were edited so the wiring didn't matter — which only holds if the vendor calls one of them. Cursor doesn't.

Fix: `TryPostHookAsync` now emits the same `AuthLapseNotice.VendorStderrLine(...)`. The drain lambda deliberately stays silent — it replays many entries per pass and would repeat the line for each one, the same duplicate-stacking reason `notification` doesn't nudge on the Claude path. Two tests added (`server_rejected_credential_names_kcap_login_on_stderr`, plus a 500 control asserting the bare status line is unchanged, so the 401 assertion is non-vacuous). `CursorHookCommandTests` 42/42.

The design doc asserted Cursor shared the poster; corrected there with a dated note rather than a silent rewrite.

### ❌ Qodo — "Linear issue identifiers in docs": not applicable in this repo

The rule this cites is `CLAUDE.md`'s "DO NOT use Linear issue numbers **in comments**", and it is enforced by the `No Linear issue IDs in C# source` CI check — which passes on this PR. It is scoped to code, not to design docs, and deliberately so: 43 existing committed specs under `docs/superpowers/specs/` carry `AI-` identifiers, several in the identical `**Issue:** AI-nnnn` header form (e.g. `2026-06-30-token-store-corruption-resilience-design.md:4`, `2026-07-09-reviewer-auto-approve-design.md:3`).

There is a reason for the asymmetry: the team convention is that the reviewed spec lives on the Linear issue as the durable record, with the repo copy riding the implementation PR. A spec that cannot name its own Linear issue cannot be traced back to that record. The GitHub number `#509` is present too, so nothing depends on a reader having Linear access.

No change.

### ◐ Qodo — "Overly verbose new comments": partly fair, trimmed in 0e0d1507

Trimmed both flagged blocks in `ClaudeHookCommand.cs`, keeping only what isn't recoverable from the code: *why* exit 0 rather than non-zero (a future editor "correcting" that exit code silently reverts the whole feature), why `stop` only, and that the failure arm is the only stdout write because the envelope below is built from a 2xx body.

Left `AgentHookPoster.cs:50-56` as-is. Its previous, shorter form actively contradicted the code — it said "no re-login nudge is surfaced here", which stopped being true in this PR — and the current length is what it takes to distinguish a pre-flight lapse (still silent) from a server-returned 401 (now nudges). It also matches the register of the surrounding docs in that file.

---

Verification after the fix: `CursorHookCommandTests` 42/42, `ClaudeHookCommandTests` 44/44, `AgentHookPosterTests` 9/9, `AuthLapseNoticeTests` 5/5, AOT publish clean (no IL3050/IL2026).

The caveat in the PR description still stands and is the one thing worth a human's attention: the `Stop`-renders-`systemMessage` contract is documented in the installed CLI bundle but has not been confirmed interactively, because headless `-p` does not fire `Stop` hooks.

alexeyzimarev:
## Folded into `AuthRejectionNotice` (755f15ef), and a note on the CI failure

### The fold

#516 landed `AuthRejectionNotice` on main while this branch was in review. Its `StoredCredentialState` vocabulary already has a `LooksValid` case meaning "the server rejected a credential that looks perfectly valid locally" — which is exactly what this branch's `Rejected` string said in prose. Shipping a second, near-identically-named auth-notice type would have been precisely the drift the type was introduced to prevent, so there is now one type.

The two surfaces keep **different renderings of the same states**, because their length budgets genuinely differ:

| Member | Surface | Form |
|---|---|---|
| `Render(state, stored, target)` | MCP tool results | several sentences (unchanged) |
| `RecordingNotice(state)` | Claude `systemMessage`, vendor stderr | one line |
| `VendorStderrLine(tag, endpoint, code)` | vendor stderr | status line, enriched only on 401 |
| `FromAuthStatus(status)` | hook pre-flight arm | maps the `AuthStatus` the hook already holds |

Sharing the enum rather than the prose is deliberate: the states are what must not drift, and a five-sentence paragraph inside a `systemMessage` would be unreadable. `FromAuthStatus` matters for a second reason — the hook path is per-turn and budget-bounded, so it must not pay the two disk reads `ForPersistentUnauthorizedAsync` makes just to name a state it already knows.

**No wording changed.** Every rendered string is byte-identical, including `WrongServer` still rendering as the not-authenticated line in the short form. Naming both servers there (as `Render` does) would be more truthful and is now a one-line change, but it is a behaviour change and not this PR's to make.

`AuthLapseNoticeTests` folded into `AuthRejectionNoticeTests`, plus new coverage for the `AuthStatus` mapping and for the `WrongServer` short-form choice so it can't be changed silently.

Verification: `AuthRejectionNoticeTests` 12/12, `ClaudeHookCommandTests` 44/44, `AgentHookPosterTests` 9/9, `CursorHookCommandTests` 42/42, AOT publish clean.

### The CI failure is not from this branch

`NotAuthenticated_MakesNoRequest_AndReturnsZero` (added by #517 / AI-1283, not touched here) **fails on `main`'s own CI at tip commit `032f477d9`** with the identical assertion, so it is red independently of this PR.

It's order-dependent rather than broken logic: it asserts no HTTP request is made when unauthenticated, but `AppConfig.SetResolvedState` and the token store are process-global statics and the test carries no `NotInParallel`. Any test running concurrently that seeds auth makes it issue the request it asserts never happens. It passes 10/10 in isolation. Worth someone's attention as a separate issue — it will redden every PR until it's isolated.

The ubuntu leg also showed `McpTelemetryTests.No_argument_data_is_carried` (`Single()` on an empty sink) once — same species, a static `TelemetryState.PathOverride` guarded only by a *keyed* `NotInParallel`. It did not appear in any of four local full-suite runs, nor on main's ubuntu leg. Watching whether it recurs.

Current `main` is merged in as of 3d7caff8.

PR #517, opened 2026-08-10
Reap leaked Claude subagent watchers with an idle ceiling

### LINKED ISSUE #514: Claude subagent watchers never idle-exit: leaked kcap watch processes accumulate for the life of the parent session
## Summary

Claude **subagent** (child) watchers have no idle-timeout backstop. `ShouldEndOnIdle` is gated to `codex`/`antigravity`/`cursor`, so a Claude child watcher's only exit paths are the `StopWatcher` signal driven by `SubagentStop` and the parent-exit watchdog. If `SubagentStop` is missed for any reason, the watcher survives until the entire parent Claude process quits — which for a long-lived session means days.

https://github.com/kurrent-io/kcap-cli/blob/main/src/Capacitor.Cli/Commands/WatchCommand.cs#L1069

```csharp
if (vendor != "codex" && vendor != "antigravity" && vendor != "cursor") return false;
```

#140 fixed the *dominant* cause of missed `SubagentStop` (fd inheritance hanging the hook, documented at `ProcessHelpers.cs:279-290`). This issue is about the missing backstop: when a `SubagentStop` is missed for any *other* reason — interrupted/aborted subagent, crashed hook, killed shell — nothing reaps the watcher.

## Evidence

Observed on kcap `0.11.16` (macOS ARM64), from a live process table. One Claude session (`b1e7a6ed…`, open 12d 23h) owned **9** `kcap watch` processes: 1 legitimate session watcher plus 8 subagent watchers whose subagents had finished days earlier. Parent PID was alive in every case, so the parent-exit watchdog correctly never fired.

Watcher age vs. last write to the subagent transcript it was watching:

| watcher age | subagent transcript last written |
|---|---|
| 11d 21h | Jul 29 21:51 |
| 10d 01h | Jul 31 18:11 |
| 9d 15h  | Aug 1 04:34 |
| 7d 22h  | Aug 2 20:25 |
| 7d 00h  | Aug 3 17:16 |
| 6d 16h  | Aug 4 01:34 |
| 1d 13h  | Aug 9 04:28 |

Roughly 8 leaked out of ~170 subagent transcripts for that session (~5%), so the miss is intermittent rather than systematic — consistent with the "disrupted flow never fires the matching `SubagentStop`" path rather than a wholesale failure.

Each idle watcher holds ~40 MB RSS and ~0% CPU, so the symptom is memory and process-table pressure, not load. Across a machine running 11 concurrent agent sessions the leaked watchers were ~280 MB.

## Impact

- Unbounded process/RSS growth proportional to session lifetime × subagent count.
- Likely leaves the corresponding server-side sessions stuck `Active`, since the watcher never posts its end.
- Scales badly with the long-lived-session + heavy-subagent-fan-out workflow the product encourages.

## Suggested fix

Give Claude child watchers a bounded lifetime that does not depend on the parent process. Two options:

1. **Extend the idle ceiling to Claude child watchers.** The Cursor child-watcher exemption at `WatchCommand.cs:1080-1084` already handles the "child watchers never buffer, so `ThresholdReached` never flips" problem — the same carve-out would apply. Cheapest change, reuses tested policy.
2. **Hard ceiling on child-watcher lifetime independent of vendor**, on the grounds that a subagent is inherently bounded work and a child watcher outliving its transcript by hours is always wrong.

Option 1 is probably right — it keeps one idle policy rather than adding a second concept. Either way the tool-in-flight suppression already in `ShouldEndOnIdle` should be retained so a genuinely long-running subagent tool call isn't cut off.

Worth confirming whether the session watcher's end-synthesis semantics should apply to a child watcher exiting on idle, the way `CursorSuppressesEndPost` scopes it for Cursor.

## Repro

Not deterministic (depends on hitting a missed `SubagentStop`). Observable in aggregate: run a Claude session with heavy subagent fan-out for several days, then compare `kcap watch --agent-id` process ages against the mtimes of the `agent-*.jsonl` files they reference.

alexeyzimarev:
Closes #514 (AI-1844).

## The leak

A Claude **subagent** watcher had exactly two exit paths: the `SubagentStop`-driven `StopWatcher`
signal, and the parent-exit watchdog. `ShouldEndOnIdle` was gated to `codex`/`antigravity`/`cursor`,
so there was no third. Miss the `SubagentStop` and the watcher survives until the entire parent
`claude` process quits.

Found while diagnosing ~110 live `kcap` processes on one machine. A single 13-day-old session owned
9 watchers: 1 legitimate session watcher plus 8 whose subagents had finished between 1 and 12 days
earlier (~40 MB RSS each). Parent PID alive in every case, so the watchdog correctly never fired —
the parent being alive is precisely why they were stuck.

#140 / AI-820 fixed the *dominant* cause of a missed `SubagentStop` (fd inheritance hanging the
hook). This adds the missing backstop for when one is missed for any other reason — roughly 5% of
that session's ~170 subagents.

## The fix

`ShouldEndOnIdle` admits Claude **child** watchers. The vendor gate became a switch that states each
vendor's eligible role rather than a flat vendor list:

| vendor | eligible role | why |
|---|---|---|
| codex / antigravity | session only | shared GUI process never exits per-conversation |
| cursor | both | no reliable per-conversation parent-exit signal |
| claude | **child only** | session watcher has a working watchdog *and* a `sessionEnd` hook |

Claude's session watcher is deliberately excluded — a ceiling there could end a live session out
from under a user who is just thinking.

**Transcript silence alone does not reap.** A subagent running a long build writes nothing between
its `tool_use` and the matching `tool_result`, and `toolInFlight` was Codex/Antigravity-only
(`UpdateCodexPendingToolCalls` parses Codex's `response_item` shape — a permanent no-op for Claude).
So this adds `UpdateClaudePendingToolCalls`, tracking `tool_use.id` → `tool_result.tool_use_id`
inside `message.content[]`, shapes taken verbatim from a real subagent transcript.

Two layers, because either alone is insufficient:

1. **Tool tracking** — an in-flight tool suppresses the ceiling regardless of how long it runs.
2. **A generous 6h window** (`KCAP_CLAUDE_SUBAGENT_IDLE_MINUTES`) — so a parser mis-read cannot reap
   a live subagent, and a subagent quiet for a reason the tracker cannot see is still safe.

6h rather than the 60m used elsewhere because this is a leak backstop, not an end-of-conversation
detector. Nothing depends on reaping promptly: a child watcher posts no `session-end` (the
`agentId is null` gate on `PostSessionEndOnParentExitAsync`), so reaping early costs only the tail
of that transcript, while the leak it fixes lasts days.

Two extractions so the wiring is covered rather than buried in `RunWatch`:

- `ResolveIdleWindow(vendor, isSessionWatcher, env)` — the per-vendor window switch, with the env
  lookup injected. Each vendor keeps reading its own knob; a Claude subagent is not retunable via
  the Codex knob.
- `TracksClaudeToolCalls(vendor, isSessionWatcher)` — the drain-loop gate. If this drifts out of
  step with the watchers the ceiling applies to, `PendingClaudeToolCalls` stays empty, `toolInFlight`
  is permanently false, and the suppression silently vanishes. That is the dangerous regression, so
  it is a predicate with its own test rather than an inline condition.

## Testing

TDD throughout — the eligibility test was watched failing with `Expected to be true but found False`
against the vendor gate before the gate changed.

- `WatchCommandTests` 77/77, `UpdateClaudePendingToolCallsTests` 7/7.
- Covered: child eligibility without the threshold gate, tool-in-flight suppression, the session
  watcher staying ineligible, the composed policy at its real 6h default, per-vendor knob isolation,
  and the tracker against multi-tool messages, string content, a Codex line, and malformed JSON.
- `CursorTailingWatcherTests` integration 5/5.
- AOT publish clean — no IL3050/IL2026.
- Full unit suite: 55 failures vs **56 on the pre-change commit**. The four that differ are all
  `AgentOrchestratorVendorTests` daemon-teardown tests; that class runs 228/228 in isolation on this
  branch, and fails an identical 5 under load at baseline. Pre-existing load flakiness, not
  regressions.

## Not addressed

This bounds the damage; it does not fix *why* `SubagentStop` gets missed. That root cause is still
open.


alexeyzimarev:
Both bugs were real and are fixed in 2ca06b570. Thanks — the first one defeated the entire feature.

**Tool ids tracked from redacted lines (codex P1) — confirmed, fixed.**
`RedactLine` returns `OversizeLinePlaceholder` for any line over `MaxRedactableLineChars` (64 KiB),
and that placeholder has no `message.content[]` at all. Oversized `tool_result`s are not an edge
case — a large file read or a build log clears 64 KiB routinely. The id was therefore never
removed, `toolInFlight` was pinned true forever, and the ceiling could never fire. The net effect
was that the leak stayed exactly as it was, and specifically on the busiest sessions, which are the
ones that leak most. Now reads `drainRead.Lines`. Only ids are parsed, so no raw content leaves the
process — redaction still governs everything sent to the server.

Regression test `RedactedLines_StrandThePendingId_ButRawLinesClearIt` pins the invariant: it asserts
the id *is* stranded when the redacted line is used, and cleared when the raw one is.

**Tool state lost on resume (qodo) — confirmed, fixed.**
Agreed on both the mechanism and the suggested approach. `BackfillClaudePendingToolCallsAsync`
rebuilds the set from `[0, LinesProcessed)` once, right after `WatcherConnect` resolves the resume
cursor — before the main loop, so there is no interleaving with the drain and no double-application
of lines at or past the cursor. It opens `FileShare.ReadWrite` (the agent is still writing that
file) and fails soft: losing the backstop is better than failing a watcher's startup over it.
Covered by four tests including the stop-at-cursor boundary and the missing-file no-op.

I took the full-scan option rather than a bounded window — a `tool_use`/`tool_result` pair has no
bounded line distance, so the window would have been a guess.

**Verbose comments (qodo) — fair, trimmed.**
Cut the flagged blocks at `WatchCommand.cs[1029-1041]` and `[1103-1113]` and the test comment by
roughly half, keeping only the rationale that isn't recoverable from the code: why the ceiling is
child-only, and why 6h rather than the 60m the other vendors use.

**One thing I did not change:** `UpdateCodexPendingToolCalls` reads the same redacted list and has
the identical exposure — an oversized `function_call_output` never clears its `call_id`, which
would suppress Codex's idle timeout indefinitely. That is pre-existing rather than introduced here,
so I have left it alone rather than widen this PR. Worth its own issue.

alexeyzimarev:
All three addressed in 50e5fb081, plus the merge with #526.

**1. Unbounded backfill scan — fixed, and it was hiding a correctness bug.**
Bounded to the last `ClaudeToolBackfillWindowLines` (512) before the cursor. I'd previously argued
against a bounded window on the grounds that `tool_use`/`tool_result` distance is unbounded. That was
wrong in the way that matters here: an *unfinished* tool sits at the tail by construction, because
nothing is appended between its `tool_use` and the matching `tool_result`. Anything older is settled.

So the window isn't just cheaper, it's more correct. A full scan would keep any stale unmatched
`tool_use` — an interrupted turn — pending forever, pinning `toolInFlight` true and disabling the
ceiling permanently. That is the same failure shape as the redaction bug from the last round, and I'd
reintroduced it. Lines before the window are still read (no cheap seek to line N) but never parsed;
the JSON parse was the cost.

**2. Backfill logs on cancel — fixed.** Early return when the token is already cancelled, and
`OperationCanceledException` caught separately from real failures. The seed-quarantine path cancels
and then calls straight into the backfill, so this was firing on every quarantine exit. I put the
guard in the method rather than at both sites, since the method is the thing that must be safe.

**3. Hard-coded oversize payload — fixed.** Now `SecretRedactor.MaxRedactableLineChars + 1024`.

**One thing I checked and deliberately did not build.** The above made me suspect a broader hole: an
orphaned `tool_use` (subagent killed mid-tool) would suppress the ceiling forever through the live
drain path too, not just the backfill — and that's exactly the disrupted-flow scenario that leaks
watchers. I checked the four transcripts whose watchers actually leaked on the diagnosed session:

    agent-ab955e3bf976781ed  tool_use=110  tool_result=110
    agent-a3eff18723ea815a1  tool_use=209  tool_result=209
    agent-a489aaf9809ffa2c7  tool_use=210  tool_result=210
    agent-a3d9875ac9a1bd2c2  tool_use=75   tool_result=75

Perfectly balanced in every case — the subagents finished their tools and it was the *stop signal*
that went missing, so the pending set drains to empty and the ceiling fires. A hard cap on
in-flight suppression would be speculative complexity against a case the data says doesn't occur,
and it would risk reaping a genuinely long tool. Noting it as a known limitation instead: if a
subagent is killed mid-tool, its watcher is not reaped by this ceiling. That's the status quo, not
a regression.

**Merge with #526.** Both conflicts were additive collisions, kept both sides: `PendingClaudeToolCalls`
alongside `CodexSubagentTurn` in `WatchState`, and the two per-vendor drain blocks. #526 doesn't touch
`ShouldEndOnIdle`, so the eligibility switch is unchanged. Worth noting the two changes are
complementary — #526 gives Codex children a live `subagent-stop` on `task_complete`, which is exactly
the finalizer Claude children lack and that this ceiling backstops.

PR #529, opened 2026-08-11
Feed the Codex watcher state trackers raw lines, not redacted ones

### LINKED ISSUE #528: Codex watcher state trackers read redacted lines: oversized tool output strands call_id, wedging idle-end and live subagent-stop
## Summary

Two watcher state trackers read the **redacted** line list instead of the raw one. `SecretRedactor.RedactLine`
replaces any line over `MaxRedactableLineChars` (64 KiB) with:

```json
{"type":"redacted_oversize_line","reason":"line exceeded SecretRedactor size limit"}
```

That placeholder carries no `call_id` and no meaningful `type`, so the trackers silently mis-fold their
state whenever a line is oversized — which for a `function_call_output` (a build log, a large file read,
a verbose test run) is routine rather than exceptional.

This is the Codex counterpart of the bug fixed for Claude in #517, found while fixing that one.

https://github.com/kurrent-io/kcap-cli/blob/main/src/Capacitor.Cli/Commands/WatchCommand.cs#L1887

```csharp
foreach (var line in newLines) {                      // newLines = drainRead.Lines.Select(RedactLine)
    UpdateCodexPendingToolCalls(state.PendingCodexToolCalls, line);
}
...
if (vendor == "codex" && agentId is not null) {
    foreach (var line in newLines) {                  // same list
        state.CodexSubagentTurn.Observe(line);
    }
}
```

Note `SeedCodexSubagentTurnState` (the resume seed, `WatchCommand.cs:1170`) already reads the rollout
**raw from disk** and folds both trackers correctly. Only the live drain path is inconsistent.

## Impact

**1. Codex sessions stuck Active forever.** An oversized `function_call_output` never removes its
`call_id`, so `PendingCodexToolCalls` stays non-empty for the life of the watcher. `toolInFlight` is
then permanently true and `ShouldEndOnIdle` can never fire. For Codex that is not a degraded backstop
— it is the *only* per-conversation session-end path (the desktop app's shared `codex app-server`
never exits per conversation, so the parent-exit watchdog cannot fire). The session stays `Active` in
the read model indefinitely and the watcher never exits.

**2. #526's live subagent-stop never posts.** `ShouldPostSubagentStop` requires no tool call in
flight, so the same stranded `call_id` permanently blocks it — reinstating exactly the symptom #526
fixed (a finished child's chat card spinning for the parent's whole lifetime).

**3. Premature subagent-stop.** `Observe` treats any `response_item` as "turn re-opened"
(`TurnCompleted = false`). An oversized `response_item` is not recognised, so a child that re-engages
after a `task_complete` can be reported stopped while it is still working.

(1) and (2) are silent and permanent; (3) is a race.

## Fix

Feed both trackers `drainRead.Lines` (raw) rather than `newLines`. Neither tracker emits anything —
they read `call_id`/`type` only — so no unredacted content leaves the process, and redaction continues
to govern everything sent to the server. This is the same fix applied to the Claude tracker in #517
(`WatchCommand.cs:1896-1899`).

Worth a regression test in the shape of `ClaudeToolTrackingSourceTests.RedactedLines_StrandThePendingId_ButRawLinesClearIt`:
assert the `call_id` *is* stranded when the redacted line is used and cleared when the raw one is, with
the payload sized off `SecretRedactor.MaxRedactableLineChars` rather than hard-coded.

## Not covered here

`UpdateCodexPendingToolCalls` has no equivalent of #517's resume backfill for the *session* watcher, so
a watcher reconnecting mid-tool starts with an empty pending set and could idle-end a live Codex
session after `KCAP_CODEX_IDLE_MINUTES`. Pre-existing and independent of the redaction bug; noted
rather than fixed. (`SeedCodexSubagentTurnState` covers only the collab-child path.)

alexeyzimarev:
Closes #528 (AI-1844 follow-up — the Codex counterpart of #517).

## The bug

Two watcher state trackers read the **redacted** line list. `SecretRedactor.RedactLine` replaces any
line over 64 KiB with a placeholder carrying no `call_id` and a `type` neither tracker recognises,
and an oversized `function_call_output` — a build log, a large file read, a verbose test run — is
routine rather than exceptional.

```csharp
var newLines = drainRead.Lines.Select(SecretRedactor.RedactLine).ToList();   // for SENDING
...
foreach (var line in newLines) UpdateCodexPendingToolCalls(...);             // for DECIDING
```

`SeedCodexSubagentTurnState` (`WatchCommand.cs:1170`) already folds both trackers from the rollout
**raw on disk**. Only the live drain path was inconsistent — the seed had it right.

## Why it matters more for Codex than it did for Claude

**1. Sessions stuck `Active` forever.** The stranded `call_id` keeps `PendingCodexToolCalls`
non-empty for the watcher's life, pinning `toolInFlight` true so `ShouldEndOnIdle` can never fire.
For Codex that isn't a degraded backstop — the desktop app's shared `codex app-server` never exits
per conversation, so the idle timeout is the *only* per-conversation session-end path. The session
stays `Active` in the read model indefinitely and the watcher never exits.

**2. #526's live subagent-stop never posts.** `ShouldPostSubagentStop` requires no tool call in
flight, so the same stranded `call_id` blocks it permanently — reinstating the exact symptom #526
fixed (a finished child's chat card spinning for the parent's whole lifetime).

**3. Premature subagent-stop.** `Observe` treats any `response_item` as the turn re-opening. An
oversized one isn't recognised, so a child that re-engaged after `task_complete` can be reported
stopped while still working.

(1) and (2) are silent and permanent; (3) is a race.

## The fix

Both trackers read `drainRead.Lines`. Neither emits anything — they read `call_id` and `type` only —
so no unredacted content leaves the process and redaction still governs everything sent to the
server. Identical to the Claude fix at `WatchCommand.cs:1896-1899`.

## Testing

`CodexToolTrackingSourceTests` mirrors `ClaudeToolTrackingSourceTests`, covering both directions:
the `call_id` **is** stranded when the redacted line is used and cleared when the raw one is, and a
redacted `response_item` fails to re-open the turn while the raw one re-opens it. Payloads are sized
off `SecretRedactor.MaxRedactableLineChars` rather than hard-coded, per the review of #517.

Being straight about coverage: these tests pin the *invariant*, and they passed the moment they were
written. The defect is at a call site inside `RunWatch`'s local `DrainNewLines`, which closes over
loop state and isn't reachable from a unit test — the same limitation as #517. What the tests
guarantee is that if `RedactLine`'s placeholder or either tracker changes shape, the reason these
call sites must use raw lines is asserted rather than assumed.

- `CodexToolTrackingSourceTests` 2/2, `CodexSubagentTurnTrackerTests` 18/18,
  `UpdateCodexPendingToolCallsTests` 6/6, `ClaudeToolTrackingSourceTests` 7/7, `WatchCommandTests` 77/77.
- AOT publish clean, no IL3050/IL2026.
- Full unit suite 72 failures vs the 56-77 band this repo's daemon cluster produces on a loaded dev
  machine; integration 3 vs 2 on main, overlapping names (`SchemePresent_ProbeHitsAuthConfig` fails
  on both). No failures in the affected area on either.

## Not fixed here

`UpdateCodexPendingToolCalls` has no equivalent of #517's resume backfill for the **session** watcher,
so a watcher reconnecting mid-tool starts with an empty pending set and could idle-end a live Codex
session after `KCAP_CODEX_IDLE_MINUTES`. Pre-existing and independent of the redaction bug —
`SeedCodexSubagentTurnState` covers only the collab-child path. Left out to keep this reviewable;
noted in #528.

### DIFF
--- a/src/Capacitor.Cli/Commands/WatchCommand.cs
+++ b/src/Capacitor.Cli/Commands/WatchCommand.cs
@@ -1884,7 +1884,12 @@ static partial class WatchCommand {
             // but runs unconditionally so it is not gated on the title phase or
             // threshold). Non-Codex lines have a different top-level shape (no
             // response_item), so this is a cheap no-op for them.
-            foreach (var line in newLines) {
+            //
+            // Raw drainRead.Lines, not the redacted newLines: an oversized function_call_output
+            // (a build log) redacts to a placeholder with no call_id, which would strand the id
+            // and pin toolInFlight true forever — and for Codex the idle timeout is the ONLY
+            // per-conversation session-end path, so the session would stay Active for good.
+            foreach (var line in drainRead.Lines) {
                 UpdateCodexPendingToolCalls(state.PendingCodexToolCalls, line);
             }
 
@@ -1903,8 +1908,12 @@ static partial class WatchCommand {
             // (task_complete vs renewed activity) so the polling loop can post a live
             // subagent-stop once the child is done + idle. Gated, unlike the
             // pending-call tracking above, because only codex child watchers ever consult it.
+            // Raw lines for the same reason, and here in the other direction: a redacted
+            // response_item is not recognised as renewed activity, so a child that re-engaged
+            // would still look finished and be reported stopped while working. Matches what
+            // SeedCodexSubagentTurnState already does when it folds this state from disk.
             if (vendor == "codex" && agentId is not null) {
-                foreach (var line in newLines) {
+                foreach (var line in drainRead.Lines) {
                     state.CodexSubagentTurn.Observe(line);
                 }
             }
--- a/src/Capacitor.Cli/Commands/WatchCommand.cs
+++ b/src/Capacitor.Cli/Commands/WatchCommand.cs
@@ -1390,6 +1390,14 @@ static partial class WatchCommand {
     /// All other lines and malformed JSON are silently ignored so this is safe to
     /// call unconditionally for every line of any vendor transcript.
     /// </summary>
+    /// <summary>
+    /// Which watchers feed <see cref="UpdateCodexPendingToolCalls"/>. It reads RAW drain lines,
+    /// which are unbounded, so — unlike when it read the 64 KiB-bounded redacted list — running it
+    /// for every vendor would make an unrelated watcher parse a multi-megabyte line it can never
+    /// match. Not gated on watcher role: a collab child needs it for ShouldPostSubagentStop.
+    /// </summary>
+    internal static bool TracksCodexToolCalls(string vendor) => vendor == "codex";
+
     internal static void UpdateCodexPendingToolCalls(HashSet<string> pending, string line) {
         try {
             using var doc  = JsonDocument.Parse(line);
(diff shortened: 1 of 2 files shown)


alexeyzimarev:
Both fixed in 643204e85.

**Raw parsing for all vendors (codex P2 / qodo 2) — confirmed, and a regression I introduced.**
You both landed on the same thing and it's a genuine one. That loop was never gated on vendor, on the
original reasoning that a non-Codex line has no `response_item` and so is a cheap no-op. That held
*only* while it read the redacted list, where any oversized line arrived as a ~90-byte placeholder.
Reading raw lines removes the 64 KiB bound, so a Claude or Gemini watcher would `JsonDocument.Parse`
a multi-megabyte tool output on every drain to rediscover that it isn't a Codex record. The Claude
tracker added in #517 was gated (`TracksClaudeToolCalls`); this one simply never needed to be until
now.

Now gated by `TracksCodexToolCalls(vendor)`, mirroring that predicate so the decision is testable
rather than an inline condition — which is what caught the analogous bug last time. Deliberately not
gated on watcher *role*: a collab child needs the tracking for `ShouldPostSubagentStop`, so it's
`vendor == "codex"` for both roles, covered by the new parameterised test.

The `CodexSubagentTurn.Observe` loop was already gated (`vendor == "codex" && agentId is not null`),
so it never had the exposure.

**Verbose comments (qodo 1) — fair, trimmed.** Cut the production block at `DrainNewLines` and the
three XML doc blocks in the tests down to intent plus a `#528` reference, keeping only what isn't
recoverable from the code: why raw lines are required, and why the two trackers fail in opposite
directions.

Verified after the change: `CodexToolTrackingSourceTests` 6/6, `CodexSubagentTurnTrackerTests` 18/18,
`UpdateCodexPendingToolCallsTests` 6/6, `ClaudeToolTrackingSourceTests` 7/7, `WatchCommandTests` 77/77,
AOT publish clean, full unit suite 60 failures against the 56-77 band this repo's daemon cluster
produces on a loaded dev machine — none in the affected area.

PR #533, opened 2026-08-11
Back-fill in-flight Codex tool calls for session watchers too

### LINKED ISSUE #532: Codex session watcher has no resume backfill for in-flight tool calls: a live session can be idle-ended mid-tool
## Summary

A Codex **session** watcher that reconnects mid-tool starts with an empty `PendingCodexToolCalls`, so
`toolInFlight` reads false and `ShouldEndOnIdle` can end a session whose tool is still running.

`WatcherConnect` sets `state.LinesProcessed` to the server's acknowledged watermark
(`WatchCommand.cs:621`), and the drain only ever returns lines at or past it. A `function_call` that
opened *before* that cursor is therefore never re-observed, and nothing rebuilds it:

```csharp
if (vendor == "codex" && agentId is not null) {   // WatchCommand.cs:474
    SeedCodexSubagentTurnState(state, transcriptPath);
}
```

The seed exists and folds the pending set correctly — but it is gated to collab **child** watchers.
The session watcher has no equivalent, which is the same blind spot #517 fixed on the Claude side
with `BackfillClaudePendingToolCallsAsync` (`WatchCommand.cs:1469`).

## Impact

A live Codex session is ended after `KCAP_CODEX_IDLE_MINUTES` (default 60, `WatchCommand.cs:1211`)
while a tool is still running, because a long-running tool produces no rollout lines between its
`function_call` and the matching `function_call_output` — which is precisely the case
`PendingCodexToolCalls` exists to suppress.

Bounded severity, and worth stating plainly:

- It needs a watcher restart/reconnect **and** a tool spanning >60 min of rollout silence across it.
- It is self-correcting in the way `DefaultCodexIdleTimeout` already documents: Codex fires a Stop
  hook per turn, which re-spawns the watcher and reactivates the session in the read model.

So the damage is a spurious `session-end` (`reason: idle_timeout`) and a gap in the session's
lifecycle, not permanent loss. Lower severity than #528, which was silent and permanent.

## Suggested fix

Give the session watcher the same treatment as the Claude one — but **not** by simply reusing
`SeedCodexSubagentTurnState` for `agentId is null`. Two reasons:

1. **That seed folds the full on-disk prefix.** #517 deliberately bounded its Claude equivalent to
   the last `ClaudeToolBackfillWindowLines` (512) before the cursor (`WatchCommand.cs:1458`), because
   an unfinished tool is at the tail by construction — nothing is appended between its call and its
   output — while an unmatched `function_call` from an *interrupted* turn further back would be
   stranded in the pending set forever, pinning `toolInFlight` true and disabling idle-end
   permanently. That is exactly the failure #528 just fixed, reintroduced from a different direction.
2. **It also folds `CodexSubagentTurn`**, which is child-only state a session watcher has no use for.

A bounded, pending-set-only backfill mirroring `BackfillClaudePendingToolCallsAsync` is the right
shape — same window constant, same read-raw/`FileShare.ReadWrite`/fail-soft properties.

## Related: the existing child seed has the unbounded-fold exposure today

`SeedCodexSubagentTurnState` folds the rollout's entire prefix, so a stale unmatched `function_call`
in a collab child's history leaves its `call_id` pending permanently. `ShouldPostSubagentStop`
requires no tool call in flight, so the live subagent-stop added in #526 would never post and the
child's card would spin until the parent-end teardown — the very symptom #526 set out to fix.

Same root cause as this issue, different layer. Worth bounding that scan in the same change.

## Context

Found while fixing #528 (Codex trackers reading redacted lines) and deliberately left out of #529 to
keep that PR reviewable. #517 is the Claude-side precedent for both the fix shape and the bounding
rationale.

alexeyzimarev:
Closes #532 — AI-1872.

## The gap

Every Codex watcher resumes at the server's acknowledged watermark (`WatcherConnect`), so the drain
only ever returns lines at or past it. A `function_call` that opened *before* that cursor is never
re-observed — and only collab **child** watchers rebuilt it:

```csharp
if (vendor == "codex" && agentId is not null) {   // the old gate
    SeedCodexSubagentTurnState(state, transcriptPath);
}
```

A session watcher therefore came up with an empty `PendingCodexToolCalls`, read `toolInFlight`
false, and could POST `session-end` (`reason: idle_timeout`) on a session whose tool was still
running — precisely the case that set exists to suppress, since a long-running tool writes no
rollout line between its call and its output. The README already promised the correct behaviour
("no new rollout lines **and no Codex tool call in flight**"), so this makes the code match the
docs rather than changing them.

Same blind spot #517 fixed on the Claude side, arriving from the other role.

## The fix

`SeedCodexSubagentTurnState` → `BackfillCodexWatcherStateAsync`, mirroring
`BackfillClaudePendingToolCallsAsync`: same window, cursor stop, `FileShare.ReadWrite`, fail-soft.
Runs for both roles after `WatcherConnect`; turn state stays child-only, since a session watcher
never reads it. Gated through `TracksCodexToolCalls` rather than an inline role check — replacing
an inline condition with a tested predicate is what caught the analogous gap in #529.

Gating on `LinesProcessed > 0` is not a behaviour change for children: at line 0 the drain delivers
the whole file and folds it itself, which is what the seed was duplicating.

## On the bounding claim in the issue — I overstated it

The issue argued the unbounded seed could strand a `call_id` from an interrupted turn, pinning
`toolInFlight` true forever. I checked before building for it, and the data does not support that:

| | |
|---|---|
| rollouts scanned | 470 (`~/.codex/sessions`) |
| ever stranding a `call_id` — at EOF **or** at a `task_complete` boundary | **0** |
| containing `turn_aborted` | 50, none of them stranding |

Codex writes the `function_call_output` even for aborted turns. So the bound is cheap insurance on
a shape we have no evidence of, not a fix for an observed failure — worth stating plainly rather
than letting the issue's framing stand.

The same data says the bound costs nothing in the other direction, which is the reason to keep it
rather than drop it: at most **2** lines separate a rollout's last state-bearing line from EOF
(median rollout 265 lines, max 1767), against a 512-line window. So a child's `task_complete` can't
realistically fall outside it, and startup work stops scaling with rollout length.

`ClaudeToolBackfillWindowLines` → `ToolBackfillWindowLines`: one number, one rationale, now two
vendors.

## Verification

- `CodexSubagentTurnTrackerTests` 24/24, `WatchCommandTests` 77/77,
  `ClaudeToolTrackingSourceTests` 7/7, `CodexToolTrackingSourceTests` 6/6
- RED verified by reverting the method to pre-fix semantics (child-only, unbounded): the three
  tests that encode the bug fail —
  `Backfill_RecoversACallLeftOpenBeforeTheCursor_ForASessionWatcher`,
  `Backfill_KeepsASessionWithARunningTool_OffTheIdleEnd` (the decision itself, wired the way
  `RunWatch` wires it), `Backfill_IgnoresACallOlderThanTheScanWindow`
- AOT publish clean, no IL2026/IL3050
- No README change: no CLI surface, flag, or env var moves

### DIFF
--- a/src/Capacitor.Cli/Commands/WatchCommand.cs
+++ b/src/Capacitor.Cli/Commands/WatchCommand.cs
@@ -1158,16 +1158,11 @@ static partial class WatchCommand {
     }
 
     /// <summary>
-    /// Codex's counterpart to <see cref="BackfillClaudePendingToolCallsAsync"/>, rebuilding what a
-    /// watcher resuming at the server watermark can never re-read. Both roles need the pending
-    /// call set: a <c>function_call</c> that opened before the cursor produces no rollout line
-    /// until its output lands, so without this a session watcher reads <c>toolInFlight</c> false
-    /// and idle-ends a session whose tool is still running. A CHILD additionally folds turn state,
-    /// or a watcher dying between the <c>task_complete</c> ack and the grace-delayed stop POST
-    /// leaves the card spinning until the parent-end teardown; a session watcher never reads it.
-    /// Bounded and cursor-stopped like the Claude backfill — the drain folds the rest.
-    /// Best-effort: an unreadable rollout just leaves fresh-start state, with the idle timeout and
-    /// the parent-end teardown as backstops.
+    /// Codex's counterpart to <see cref="BackfillClaudePendingToolCallsAsync"/>, same window and
+    /// fail-soft properties. Both roles need the pending set — a <c>function_call</c> that opened
+    /// before the cursor writes nothing until its output lands, so a session watcher would read
+    /// <c>toolInFlight</c> false and idle-end a live session. Turn state is child-only: nothing
+    /// else re-delivers the acknowledged <c>task_complete</c> that arms the live stop.
     /// </summary>
     internal static async Task BackfillCodexWatcherStateAsync(
             WatchState state, string transcriptPath, bool isChildWatcher, int upToLine, CancellationToken ct) {
--- a/src/Capacitor.Cli/Commands/WatchCommand.cs
+++ b/src/Capacitor.Cli/Commands/WatchCommand.cs
@@ -464,16 +464,10 @@ static partial class WatchCommand {
 
         // Idle grace between a Codex collab child's task_complete and its live
         // subagent-stop POST (codex child watchers only — see the loop's stop check). The
-        // agent type is read lazily from the child rollout's own header, once. The seed
-        // rebuilds turn/pending state from the acknowledged prefix, which a restarted
-        // watcher's watermark-resumed drain would otherwise never re-deliver.
+        // agent type is read lazily from the child rollout's own header, once.
         var     codexSubagentStopGrace   = CodexSubagentTurnTracker.ResolveStopGrace(Environment.GetEnvironmentVariable("KCAP_CODEX_SUBAGENT_IDLE_MINUTES"));
         string? codexChildAgentType      = null;
         var     codexStopNextAttemptAt   = DateTimeOffset.MinValue;
-
-        if (vendor == "codex" && agentId is not null) {
-            SeedCodexSubagentTurnState(state, transcriptPath);
-        }
 
         if (skipTitle) {
             state.TitleGenerated = true;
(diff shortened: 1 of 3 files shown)


alexeyzimarev:
**1. Verbose comments — fair, trimmed in 3d7905f1a.** Third PR running this rule has caught me. Cut the
XML doc from ten lines to five and the test comments to one line each, keeping only what the code
can't say: why both roles need the pending set, and why turn state is child-only.

**2. Unbounded prefix reads — the observation is right, the restructure isn't worth it. Here are the
numbers.**

You're correct that the loop materializes a string per line below `firstParsedLine`, and correct that
Codex lines get large — I measured **2.1 MiB** for the biggest single line across 471 local rollouts,
so that part of the premise is real, not theoretical. But the cost this actually imposes:

| | |
|---|---|
| rollouts where the window binds at all (> 512 lines) | **56 / 471 (12%)** |
| read-and-discarded prefix — median | **0 KiB** |
| p95 | 1.0 MiB |
| max | 4.5 MiB |

For 88% of rollouts `firstParsedLine` is 0 and there is no discarded prefix; the worst case is a
one-time 4.5 MiB read at watcher startup. Worth noting this is also strictly cheaper than what it
replaced for child watchers: the old seed read to EOF *and* ran two `JsonDocument.Parse` calls per
line, where this reads no further than the cursor and parses at most 512 lines.

Against that, the suggested fix has a specific cost I don't want to pay here. A backwards tail-reader
or byte-chunk newline scan introduces a **second definition of "line N"** next to the cursor
frontier — and this file already carries a long comment about a "permanent, silent line/byte-frontier
misalignment" from exactly that class of mistake (`SeedCursorByteOffsetAsync`, `ByteOffsetForAckedLines`).
`ReadLine` counts `\r\n`, `\n` and a lone `\r`; a `\n`-counting byte scan doesn't. Trading a bounded
4.5 MiB worst-case read for a silent cursor-alignment risk is the wrong side of that deal.

It would also have to be duplicated into `BackfillClaudePendingToolCallsAsync`, which has the
identical loop and shipped in #517 with the same reasoning stated inline ("there is no cheap seek to
line N"). Two divergent implementations of one thing is worse than the read.

So: not a bug — the code is correct — and an efficiency item whose measured magnitude doesn't justify
the risk in this PR. If it's ever worth doing it should be one shared helper across both backfills,
built on the existing line/byte mapping rather than a new one, and that's its own change with its own
tests.

Issue #535, opened 2026-08-11
WorkOS has no headless auth path, which is what keeps legacy GitHub App auth alive


alexeyzimarev:
Follow-up to #534. That issue stops routing headless users onto legacy GitHub App auth; this one removes the reason the routing existed.

## Problem

WorkOS sign-in cannot complete in a non-interactive environment, so headless `kcap setup` has no working auth path at all once #534 lands.

The blocker is **not** the inability to launch a browser — `IBrowser` is already injectable (`OAuthLoginFlow.cs:527`). It is the loopback redirect:

```csharp
// OAuthLoginFlow.cs:525
var redirectUri = $"http://127.0.0.1:{GetAvailablePort()}/callback";
```

On a remote box (SSH, container) the user's browser runs on their laptop, so `http://127.0.0.1:{port}/callback` resolves to *their laptop's* localhost and never reaches the CLI's listener. GitHub App auth worked headless only because GitHub offers a device flow (`RunDeviceFlowAsync` → `github.com/login/device/code`) that needs no callback to the CLI's machine.

So: legacy auth is the only headless-capable provider today. That is the entire reason `ChooseDiscoveryProvider` falls back to it.

## Options

**A. OAuth 2.0 Device Authorization Grant via WorkOS.** Cleanest — no callback, and it matches the UX users already know from the GitHub flow. **Whether WorkOS supports the device grant is unverified.** Nothing in kcap-server or kcap-web implements it today. This needs confirming first; it decides the shape and cost of the whole change.

**B. Out-of-band paste-back.** If WorkOS has no device grant: authorize in a browser on any machine, redirect to a kcap-web page that displays a short-lived one-time code, user pastes it into the CLI, CLI exchanges it for tokens. Needs a small web endpoint plus a short-TTL code store. No WorkOS feature dependency; a standard fallback for exactly this constraint.

Either way `SpectreTenantProvisioner` also needs a non-Spectre counterpart, since the current provisioner is nulled when headless (`SetupCommand.cs:843`) and prompts interactively for the slug.

## Why it matters

Headless is not a rounding error: 1 of 9 real (non-CI, released-build) CLI setup devices was headless, and 1 of the 3 that reached the no-workspace fork. At the current ~3–5 signups/day that is roughly one affected person every 1–3 days. Small n — treat as an order of magnitude, not a rate.

Track it on the `is_headless` breakdown: https://eu.posthog.com/project/11377/insights/Cd3S4b0i

## Definition of done

[...]

PR #536, opened 2026-08-11
Stop routing headless setup onto legacy GitHub App auth

### LINKED ISSUE #534: Headless kcap setup routes to legacy GitHub App auth and dead-ends there
> **Rewritten.** This issue first proposed improving the copy on the headless dead-end. That was wrong: the dead-end is an artifact of routing headless users onto **legacy GitHub App auth**, and better copy there would advertise a path we are migrating off. The fix is to stop routing to it.

## What happens

A user with no workspace runs `kcap setup` in a non-interactive environment (container, SSH, no local browser). They complete a GitHub device-code sign-in, are told they have no workspace, and setup exits with legacy advice.

Observed 2026-08-11, Linux, v0.11.18, `is_headless: true` — they hit it, retried the whole flow, hit it again, gave up:

```
16:56:37  cli_setup_started          no_prompt: false
16:56:39  cli_setup_signin_opened    GitHubApp, device
16:57:21  cli_setup_signin_completed
16:57:22  cli_setup_tenant_none      ← setup exits 0.3s later
16:57:59  cli_setup_started          ← retries
17:02:09  cli_setup_tenant_none      ← same wall, gives up
```

`cli_setup_workspace_offered` never fired — they were never asked, they didn't decline.

## Root cause

```csharp
// OAuthLoginFlow.cs:70
return isInteractive ? AuthProvider.WorkOS : AuthProvider.GitHubApp;
```

They never passed `--github`. Being non-interactive selected legacy auth for them. The GitHubApp branch (`SetupCommand.cs:892-904`) has no provisioning path — by design, it discovers orgs you already belong to — so it emits `cli_setup_tenant_none` and returns null with:

> "No Capacitor tenants are linked to your GitHub orgs. **Ask your admin to install the Kurrent GitHub App on your org**, or re-run with `--server-url <url>` for a self-hosted server."

For someone evaluating kcap there is no admin, and the GitHub App is not a path we want new users on at all.

## Why this is the last legacy feeder

CLI telemetry, released builds, non-CI, since telemetry went live:

| Provider | Devices | Session type |
|---|---|---|
| workos | 6 | all interactive |
| GitHubApp | 1 | headless |

Every interactive user already lands on WorkOS. **Headless routing is the only remaining source of new GitHub App authentication.** Removing it takes new-signup traffic on legacy auth to zero.

## Fix (this issue)

Stop routing headless discovery to GitHubApp. Fail fast *before* authenticating, with a message that offers the two real options: create a workspace at the signup URL, or pass `--server-url <url>` for an existing one. No GitHub App mention.

This also turns a ~40s dead end (full device-code dance, then rejection) into an immediate, actionable failure.

**Blast radius** — deliberately narrow:
- `ChooseDiscoveryProvider` only affects *discovery* runs (no `--server-url`).
- Existing GitHub App workspaces are unaffected: with `--server-url` the provider comes from that server's `/auth/config` (`SetupCommand.cs:892`).
- `--github` remains as an explicit escape hatch.

## Not this issue

Making headless setup actually work end-to-end — tracked separately. Until that lands, headless users with no workspace are told to sign up in a browser rather than being routed onto legacy auth.

alexeyzimarev:
Closes #534. Follow-on work tracked in #535. Linear auto-imports both.

## Problem

A non-interactive `kcap setup` / `kcap login` with no `--server-url` fell back to the legacy GitHub App provider:

```csharp
// OAuthLoginFlow.cs:70 (before)
return isInteractive ? AuthProvider.WorkOS : AuthProvider.GitHubApp;
```

The fallback existed for a real reason — GitHub has a device flow, while WorkOS authenticates through a `http://127.0.0.1:{port}/callback` loopback that a browser on another machine can never reach. But it dead-ended: the GitHub App branch has no provisioning path by design (it discovers orgs you already belong to), so a headless user with **no** workspace completed a full device-code sign-in and was then told:

> "No Capacitor tenants are linked to your GitHub orgs. **Ask your admin to install the Kurrent GitHub App on your org** […]"

For someone evaluating kcap there is no admin. Telemetry caught one such user on 2026-08-11 who hit it, retried the entire flow, hit it again 4 minutes later, and gave up. `cli_setup_workspace_offered` never fired — they were never asked, they didn't decline.

## Why remove the routing rather than fix the copy

CLI telemetry since the funnel went live (released builds, non-CI):

| Provider | Devices | Session type |
|---|---|---|
| workos | 6 | all interactive |
| GitHubApp | 1 | headless |

Every interactive user already lands on WorkOS. **Headless routing was the only remaining source of new GitHub App sign-ins.** Better copy on that dead-end would have advertised a provider we're migrating off and polished the one path still feeding it.

## Change

`ChooseDiscoveryProvider` returns `null` for a non-interactive session. Both call sites report `HeadlessDiscoveryUnsupportedMessage()` and stop:

```
Setting up a new workspace needs an interactive terminal, and this session is non-interactive.
  • Create a workspace at https://capacitor.kurrent.io/signup, then run: kcap setup <slug>
  • Or point at an existing workspace: kcap setup --server-url <url>
```

No GitHub App mention; both routes that actually work are named. The signup URL derives from `ProvisioningEndpoint.Url`, so it honours the existing `KCAP_SIGNUP_URL` dev/preview override.

The check also moves **ahead of** the auth-service call in both call sites, so the run fails in ~2s instead of after a full device-code dance it can't use.

## Blast radius — deliberately narrow

- Discovery only (no `--server-url`).
- Existing GitHub App workspaces unaffected: with `--server-url` the provider comes from that server's `/auth/config` (`SetupCommand.cs:892`).
- `--github` remains an explicit escape hatch and **still works headless** — verified against the real binary, it proceeds to the device flow.

## Tests

TDD: tests written first, watched fail, then implemented.

- `Headless_discovery_does_not_fall_back_to_legacy_github_app`
- `Explicit_github_flag_still_selects_legacy_provider_when_headless`
- `Headless_discovery_message_offers_signup_and_server_url_but_never_the_github_app` — asserts the copy contains no `GitHub App`, and does contain `/signup` and `--server-url`
- Existing `ChooseDiscoveryProvider_honors_flags_and_default` updated (it pinned the old fallback)

Verified against the AOT binary with `SSH_CONNECTION` set:

| Command | Result |
|---|---|
| `kcap setup` headless | message shown, no network, exit 1 |
| `kcap login` headless | message shown, no network, exit 1 |
| `kcap setup --github` headless | reaches GitHub device flow (escape hatch intact) |

Full unit suite: 56 failures here vs **57 on a clean `origin/main` baseline**, all the known load-sensitive timing flakes. The two that differed run-to-run (`Agent_finalized_before_the_bind_completes…`, `Wedged_initialize_reaches_LaunchFailed…`) both pass in isolation and touch neither auth nor setup. AOT publish clean — no IL2026/IL3050.

## Docs

`README.md` and `help-login.txt` both asserted the old fallback ("discovery falls back to GitHub Device Flow"); both now describe the interactive requirement and the two routes.


alexeyzimarev:
Thanks both — one real regression caught, one doc gap, one claim that doesn't hold. Addressed in c870be55e.

## ✅ Codex: `--device` regression — valid, fixed

Correct and important. `ChooseDiscoveryProvider` returned `null` for *any* non-interactive session, which broke `kcap login --device` / `kcap setup --device` — documented as the SSH/container sign-in method in three places (`help-login.txt:31-33`, `help-setup.txt:13-14`, `README.md:1850`).

Reproduced against the AOT binary before the fix: `kcap login --device` with `SSH_CONNECTION` set printed the new headless message instead of starting the device flow.

`--device` is an explicit request in the same class as `--github`, not the implicit fallback this PR removes, so it now selects the legacy provider — **on the headless branch only**:

```csharp
if (args.Contains("--github")) return AuthProvider.GitHubApp;
if (isInteractive) return AuthProvider.WorkOS;
return args.Contains("--device") ? AuthProvider.GitHubApp : null;
```

Deliberately *not* folded into the `--github` check at the top. On a machine with a browser, `--device` has always meant "use the device flow if the login step needs GitHub", never "take discovery off org SSO" — hoisting it would regress the interactive path in the opposite direction. Both directions are now pinned:

- `Explicit_device_flag_still_selects_legacy_provider_when_headless`
- `Device_flag_does_not_divert_an_interactive_session_away_from_sso`

Verified end-to-end: `kcap login --device` headless now reaches `https://github.com/login/device` with a real user code; plain `kcap login` / `kcap setup` headless still print the message and exit 1.

## ✅ Qodo, first half: docs omitted the escape hatches — valid, fixed

Right that the docs implied discovery simply stops. Both now name what still works headless: `--server-url` to configure a known workspace, and `--device` / `--github` to discover via GitHub Device Flow — flagged as the legacy path being phased out (#535), so readers prefer the first.

## ❌ Qodo, second half: "`login` doesn't read `--server-url`" — doesn't hold

> it implies `--server-url` is a viable remedy from `kcap login` even though `login` doesn't read that flag

`kcap login` does read it. `Program.cs:79` resolves `baseUrl` via `AppConfig.ResolveServerUrl(args, …)`, which scans argv for `--server-url` (`AppConfig.cs:94`); a non-null `baseUrl` then makes `ShouldDiscoverLogin` return false, so login goes straight to that server instead of discovery.

Confirmed against the binary — it attempts the connection rather than falling into discovery:

```
$ kcap login --server-url https://nonexistent.invalid    # SSH_CONNECTION set
Kurrent Capacitor API cannot be reached, is it running? …
Error connecting to: https://nonexistent.invalid  nodename nor servname provided
```

Worth noting the message itself was never ambiguous — it says `kcap setup --server-url <url>`, naming the command that configures a workspace. The help-login prose was the loose part, and that's now explicit.

## Verification

Full unit suite: 59 failures vs **57 on a clean `origin/main` baseline**. The four that differ (`PostOnceAsync_times_out_after_specified_duration`, both `Teardown_*`, `Wedged_initialize_*`) are the known load-sensitive wall-clock-budget flakes — **all four pass in isolation**, and none touches auth or setup. AOT publish clean.

PR #538, opened 2026-08-12
[AI-1887] Emit judge-fact guidelines at session start for all harnesses

realtonyyoung:
## What

SessionStart guideline injection (`## Known patterns` / `## Guidance from past sessions`, evaluation-derived fact clusters) reached only Claude Code, which reads `top_clusters` off its hook POST response. The other eight harnesses got the team-memory index but no guidelines.

This adds a parallel **guidelines lane** to the existing per-vendor memory orchestration and composes **one combined fragment** (marker → `## Team memory` → guidelines) that rides each harness's existing delivery seam **unchanged** — no plugin/extension/envelope changes.

## How

- **`ISessionStartContextProvider`** seam on the orchestrator; **`SessionStartContextFetch`** shares the authenticated-GET / 401-refresh / 256 KiB bounded-read transport between the two lanes (byte-for-byte what the memory lane did before).
- **`SessionStartGuidelinesLane`** fetches `GET /api/repositories/{hash}/guidelines` and maps **404 → retryable** — the visibility race: a 404 means "not visible yet" (the endpoint gates on a projected caller-visible session), not "no facts". The memory lane keeps 404 → empty; this deliberate divergence is why it's a separate lane.
- **`SessionStartCompositeContextProvider`** resolves the repo/machine scope **once** (matters under Cursor's 2s budget) and runs both lanes in parallel, combining over **enabled lanes only**: any content ⇒ commit; all empty ⇒ complete-without-context; no content + ≥1 retryable ⇒ retry with the **max** of the lanes' `Retry-After` hints. A disabled lane contributes nothing and never blocks commit. Marker-first compose: the guidelines-only case prepends the shared `kcap-memory-index` marker so Pi/OpenCode capture (stdout must OPEN with it) is unchanged.
- The orchestrator's early-return becomes **both-disabled**; every non-Claude adapter routes through `SessionStartMemoryHookSupport.CompositeProvider` and passes the guidelines opt-out.
- **Claude is untouched** except reading `disable_session_guidelines` from the **effective** profile (the pre-existing `KCAP_URL`/`--server-url` defect where `ResolvedProfile.Profile` is null) — it keeps its memory-only provider and issues **no** guidelines GET (pinned by a test).
- **Cursor** gains effective-profile threading placed **inside** its `HandleCore` deadline race (so a slow config read can't blow the 2s dispatcher contract — `HandleCore` abandons the computation on the deadline), plus the same `DisableMemoryIndex` effective-profile fix.

No CLI-side size cap (matches Claude; the server clamps row count + text length).

## Server side

The endpoint `GET /api/repositories/{hash}/guidelines` already exists. Its response contract + a `GuidelineInjectionConfig` size-bound validator are hardened in the companion kcap-server PR (independent, non-blocking in either direction — this CLI change is well-defined against any server version/config: an over-256 KiB or missing response is a fail-open retryable lane failure, never a crash, and never blocks a delivered memory lane).

## Tests

New `GuidelinesLaneAndCompositeTests` (17): guidelines-lane status mapping incl. **404 → retryable**, composite disposition matrix + lease/compose rules, `Retry-After` max aggregation, marker-first compose, disabled-lane isolation, emitter row overload. Plus a `ClaudeHookCommandTests` isolation test proving Claude issues no guidelines GET. All existing per-vendor memory suites (138) + foundation (31) + Cursor (42) + Codex source (16) unchanged and green.


realtonyyoung:
Addressed both qodo findings. (1) Uncancellable Cursor profile load: `AppConfig.GetActiveProfileAsync` now takes an optional `CancellationToken` and threads it into `LoadProfileConfig(ct)`; the Cursor hook passes `dispatcherCt`, so when `HandleCore` abandons the inner computation on the 2s deadline the config read observes cancellation instead of lingering (existing parameterless callers unchanged). (2) Verbose comment trimmed to two intent-focused lines.

PR #545, opened 2026-08-12
[AI-1863] Cap the passive update notice + kcap status at the connected server's version

realtonyyoung:
## Context

The server half ([kcap-server #1404](https://github.com/kurrent-io/kcap-server/pull/1404), AI-1863) caps CLI-update recommendations at **`min(npm latest, server version)`** and emits an `X-Kcap-Server-Version` response header. This is the **client half**: the CLI now consumes that header so its own passive "update available" stderr notice and the `kcap status` version line never steer a user to a CLI newer than the server they talk to — manual tenant rollouts can trail npm by days, and a newer CLI risks protocol mismatch against the older server.

## What changed

- **`ServerVersionStore`** — a durable, per-normalized-server-URL cache of the server version (one flat file per server in the config dir; best-effort, with an in-process write-dedup so the per-response hot path touches disk at most once per distinct value).
- **`ServerVersionCaptureHandler`** — the outermost `DelegatingHandler` on every authenticated client (the `CreateClientCoreImplAsync` choke point, the same one that attaches the request-side observation headers), reading `X-Kcap-Server-Version` from each response into the store. No extra requests; best-effort, never alters the response.
- **`UpdateAdvisoryResolver`** — caps the npm-latest target at `min(npm, cached server version)`, **only** on the stable `latest` channel with a cached, stable server version present (beta deliberately rides ahead; an absent cache keeps today's behaviour — the cold-start doctrine). Recomputes "newer" against the capped target, so a user already at/ahead of their server is never nagged.
- **`UpdateNotice`** (passive exit-time stderr notice) and **`kcap status`** render the capped advisory: when capped, a pinned `npm install -g @kurrent/kcap@<target>` command and a `(server version)` marker (plain `kcap update` follows the dist-tag and would overshoot the cap). **`kcap update` and `--check` stay UNCAPPED** — they're explicit actions, not recommendations.

**Surface D (the in-agent `VersionNudgeEmitter`) needs no change**: the server *omits* the hook `version` field when the target is capped, so the emitter only ever receives an uncapped target (where plain `kcap update` is correct).

## Tests

29 unit tests, all green locally:
- `ServerVersionStoreTests` (round-trip, normalization, per-server independence, overwrite, no-op guards);
- `ServerVersionCaptureHandlerTests` (header present → captured; absent → nothing);
- `UpdateAdvisoryResolverTests` (the cap truth table — passthrough when uncapped/beta/prerelease/absent, cap at min, user-ahead-not-nagged, plus the stable-release gate);
- `StatusVersionLineFormattingTests` (bare / uncapped / capped-with-marker rendering).

README + `help-update.txt` updated in the same PR (house rule).

Completes the client half of [AI-1863](https://linear.app/kurrent/issue/AI-1863).

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,8 @@ error, so you no longer have to run `kcap whoami` to work out why recording stop
 hooks print the same advice to stderr instead of an in-session notice, since not every agent
 surfaces hook output in its UI. `kcap status` prints its own
 **Version** line — the installed CLI version, with an inline `(update available: …)` annotation
-when a newer one is out — see [`kcap update`](#other-commands) for the full opt-out story.
+when a newer one is out (capped at your connected server's version, marked `(…, server version)` when
+your tenant trails npm) — see [`kcap update`](#other-commands) for the full opt-out story.
 
 Setup closes with a **Next steps** box. Each item opens with a question, because neither step is for
 everyone:
--- a/src/Capacitor.Cli.Core/HttpClientExtensions.cs
+++ b/src/Capacitor.Cli.Core/HttpClientExtensions.cs
@@ -78,11 +78,18 @@ public static class HttpClientExtensions {
         HttpClient NewClient(DelegatingHandler? retry = null) {
             var primary = new HttpClientHandler { AllowAutoRedirect = allowAutoRedirect };
 
-            if (retry is null) return new(primary);
+            HttpMessageHandler inner = primary;
 
-            retry.InnerHandler = primary;
+            if (retry is not null) {
+                retry.InnerHandler = primary;
+                inner = retry;
+            }
 
-            return new(retry);
+            // Capture the server's own version (X-Kcap-Server-Version) from every response — outermost,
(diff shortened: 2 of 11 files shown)


realtonyyoung:
Both qodo findings addressed in 6373435:

1. **Wrong server cache key** — `ServerVersionStore.Normalize` now reuses the repo's own `ServerIdentity.Canonicalize` (scheme+host lower-cased, default ports converged, path preserved case-sensitively), with a conservative non-lowercasing fallback for inadmissible URLs. Removes the divergent second URL-identity function; path-routed tenants no longer collide and `:443` converges with implicit.
2. **Build metadata in target** — the capped target is now run through `CapacitorVersion.Display` to strip `+buildmetadata`, so a MinVer commit-SHA server version can't leak into the `(server version)` copy or the pinned `npm install -g @kurrent/kcap@<target>` command.

Added tests: default-port convergence, path-case significance, and build-metadata stripping on a capped target.

PR #555, opened 2026-08-13
[AI-1941] Fix daemon doctor --clean never removing lock-only stale entries

realtonyyoung:
## Problem

`kcap daemon doctor --clean` reports N stale entries, claims to clean them, but the next `kcap daemon doctor` shows the exact same entries — forever. A user hit this with 9 stale daemon entries that `--clean` could never remove.

## Root cause

`doctor`'s listing is driven by `DaemonLockPaths.EnumerateNames()`, which unioned `*.lock` into the name set. For a confirmed-stale entry whose only leftover was the `.lock` file, `--clean` deletes the state markers but **deliberately leaves the `.lock`** — and it must: the lock file is a per-inode `flock` mutex, not a liveness marker. Unlinking it is genuinely unsafe (a later `daemon start` recreates a fresh inode at the same path and holds a *second* independent flock; supervised launchd/systemd daemons take no serialising start-lock, so there's no safe window to delete it). Because the `.lock` stayed and enumeration keyed on it, the name was re-listed on every run.

## Fix

Derive names only from the **state markers** — `*.pid` / `*.restart-pending` / `*.version` — never from a lone `*.lock`. A live daemon always writes `.pid` inside the flock during startup, so nothing real is hidden; a bare lock is an inert leftover and simply stops being listed. Once a stale entry's markers are cleaned it disappears from the listing even though the inert lock lingers (reused by the next start of that name).

This is a **single conceptual change** with no concurrency surface: `DaemonCommands` / the lock-deletion behaviour is untouched. It also makes the user's existing bare-lock leftovers vanish on upgrade with zero file deletion, on every platform.

### Why not just delete the lock in `--clean`?

That was the first attempt; it's unsound. `.NET` maps only `FileShare.None` to `flock(LOCK_EX)` (any shared mode is `LOCK_SH`), and unlinking a flock file breaks the per-inode mutex — reintroducing exactly the double-daemon race the original code avoided by never deleting it. A codex review flagged this; the enumeration change sidesteps it entirely.

## Tests

- `DaemonLockEnumerationTests` — a bare `.lock` is **not** a name source; markers are; dedup across markers holds.
- `DaemonDoctorCleanTests` — removing a stale entry's markers drops it from `EnumerateNames` while the lock file lingers; a bare-lock leftover (the exact user shape) is never listed.

All green locally (31 daemon lock/enumeration/marker tests, 0 failures).

Fixes AI-1941.

### DIFF
--- a/README.md
+++ b/README.md
@@ -692,7 +692,7 @@ kcap daemon restart --name laptop              # restart now if idle; refuses wh
 kcap daemon restart --name laptop --when-idle  # queue the restart for the next idle moment
 kcap daemon restart --name laptop --force      # restart now even if busy (tears down running agents)
 kcap daemon doctor                  # diagnose lock-file state for every daemon name
-kcap daemon doctor --clean          # also remove stale lock/pid files (held entries are never touched)
+kcap daemon doctor --clean          # also remove a stale entry's pid/marker files, dropping it from the list (held entries are never touched; the inert lock file is left in place)
 ```
 
 `agent` was this group's name before it was renamed to `daemon`, and it now belongs to a different group — [`kcap agent`](#local-agents-kcap-agent) runs coding agents. The two still share `start` and `stop`, so check which one you mean: `kcap daemon start` starts the daemon, `kcap agent start <vendor>` starts a coding agent. Daemon-only verbs typed against the wrong group (`kcap agent status`, `restart`, `logs`, `doctor`, `service`) answer with a pointer back here.
--- a/src/Capacitor.Cli.Core/DaemonLockPaths.cs
+++ b/src/Capacitor.Cli.Core/DaemonLockPaths.cs
@@ -117,20 +117,27 @@ public static partial class DaemonLockPaths {
     public static void EnsureDirectory() => System.IO.Directory.CreateDirectory(Directory);
 
     /// <summary>
-    /// Returns the daemon names visible on disk — the union of names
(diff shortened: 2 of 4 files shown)


realtonyyoung:
Thanks — addressing the qodo findings against `ac50751`:

1. **Coverage chain broken** (🐞, already ✓ Resolved) — correct and resolved. This was against the earlier approach that unlinked `<name>.lock` in `CleanStaleEntryUnderProbe`, which (as flagged) would also break `CoverageJournal.RecordBoot` chain validation. That approach was abandoned; the final fix does **not** delete any lock file — it changes `EnumerateNames()` to stop deriving names from a lone `.lock`. No lock deletion, no `DaemonCommands` change, so the coverage chain is untouched.

2. **README `doctor --clean` outdated** — fixed in `ac50751`. The line now reads that `--clean` removes a stale entry's pid/marker files (dropping it from the list) and leaves the inert lock file in place.

3. **Verbose test header comments** — the cited `TryProbeStaleLock`/`CleanStaleEntryUnderProbe` docs no longer exist (that code was reverted). Tightened the remaining `DaemonDoctorCleanTests` / `DaemonLockEnumerationTests` headers to be concise in `ac50751`, keeping only the essential rationale.

The one substantive design comment (`.NET` maps only `FileShare.None` to `flock(LOCK_EX)`; unlinking a flock file breaks the per-inode mutex; supervised daemons take no serialising start-lock) is exactly why the final fix is enumeration-only.

Issue #565, opened 2026-08-15
Empty the TreatWarningsAsErrors triage list so new warnings actually fail the build


Inok:
`TreatWarningsAsErrors` is on, but `Directory.Build.props` carries a ~50-code `WarningsNotAsErrors` triage list plus a `NoWarn` block — every code that was already warning when the flag went on. While a code sits on that list it is a warning, so a *new* occurrence of it lands silently.

The list should be empty: fix the sites, or move the decision into `.editorconfig` as a deliberate severity choice, and delete the entry either way.

Also in scope: the NuGet audit codes (NU1901–NU1904) are all exempted repo-wide, so a high or critical advisory in a shipped dependency is only a warning.

AI-1932

Issue #543, opened 2026-08-12
macOS unit-suite failures are mostly one deterministic bug, not flakes: CodexConfigToml rejects /var as a symlink


alexeyzimarev:
## Summary

The unit suite's "macOS-local flake cluster" — the reason `ci.yml` deliberately keeps macOS out of
the test matrix — is mostly **not flaky**. 49 of the 64 failures are a single deterministic bug that
reproduces on every Mac, and it's in production code, not the tests.

`CodexConfigToml` refuses to write through a symlink by walking *every ancestor component* of the
config path and rejecting any reparse point:

```csharp
static void RejectSymlink(string path) {                       // CodexConfigToml.cs:600
    if (!File.Exists(path) && !Directory.Exists(path)) return;
    if ((File.GetAttributes(path) & FileAttributes.ReparsePoint) != 0)
        throw new IOException($"Refusing to update symlinked Codex configuration target: {path}");
}
```

On macOS `TMPDIR` is `/var/folders/…` and **`/var` is a symlink to `private/var`**. So every test
that writes a Codex config under `Path.GetTempPath()` fails at the first path component, before
anything is written — surfacing either as `Change.Failed` where `Updated` was expected, or as
`FileNotFoundException` when the test then stats the config that was never created.

## Evidence

One variable changed, nothing else:

| run | `CodexConfigTomlTests` | full unit suite |
|---|---|---|
| default `TMPDIR=/var/folders/…` | **20 / 41 fail** | **64 fail** |
| `TMPDIR=/private/tmp` | **0 / 41 fail** | **15 fail** |

The 49 that disappear span `CodexConfigTomlTests` (20), `UninstallCommandTests` (11),
`CodexConfigWriterTests` (6), part of `AgentOrchestratorVendorTests` (5), `CodexLauncherTests` (3),
`CodexProjectKeyTests` (2), `PluginCommandCodexInstallIntegrationTests` (2). They fail in isolation
too, so this was never a parallelism problem.

## The 15 that remain really are load flakes

`AgentOrchestratorVendorTests` (12 — mostly `TimeoutException`, plus a process-occupancy assertion)
and `BorrowedReviewAuthBrokerTests` (3 — `elapsed 15.3s` against a 10s
`BorrowedReviewTokenCommand.Timeout` assertion). Wall-clock assertions on a loaded dev machine.
Worth a separate look, but they're a much smaller and more honest target than 64.

## Careful with the obvious workaround

[...]

PR #567, opened 2026-08-15
Clear the TUnit analyzer warnings and bump WireMock to 2.15.0

Inok:
Part of #565 (AI-1976). **Stacked on #566** — base is `pavel/warnings-src`, so review that one first; this diff shows only the two commits on top.

Completes the triage-list removal that #566 starts. With these two commits the repo builds with **0 warnings and 0 errors**.

## Changes

- **Cleared the TUnit analyzer backlog** (`TUnit0055`, `TUnit0301`, `TUnitAssertions0005`, `TUnitAssertions0015`) and deleted the exemption list from `test/Directory.Build.props`, which is now down to the NuGet-audit line alone. `TUnit0055` sites moved to a shared `ConsoleOutput` helper in a new `Capacitor.Tests.Helpers` project rather than each test overwriting the console writer.
- **TUnit 1.18.37 → 1.65.0.**
- **WireMock.Net 1.7.0 → 2.15.0**, which routes via `WireMock.Net.Minimal` to `Scriban.Signed` 7.2.5 and clears all 13 NuGet advisories (26 warnings) that #566 leaves standing under `test/`. This is what makes that PR's `NU1903`/`NU1904` escalation green rather than merely aspirational.

## Notes for review

- The `NU1903`/`NU1904` exemption in `test/Directory.Build.props` now exempts nothing — with Scriban at 7.2.5 there is no advisory left for it to cover. It is standing policy for the next advisory, not dead config.
- WireMock 2.x annotates `ILogEntry.RequestMessage` nullable (undocumented in any changelog), which turned every log-entry assertion into `CS8602` — 235 sites across ~76 files, one single pattern. Rather than 235 `!` edits, `CS8602` is a suggestion under `[test/**.cs]`. **This is wider than the sites that forced it**: null-deref checking is now off for all test code. It hides nothing today (tests had zero `CS8602` before the bump), but a genuine null bug in a new test will not fail the build. Narrowing this is a fair ask.
- The documented 2.x breaks do not reach us: .NET Framework <4.8 (we are net10.0), the Handlebars `File`/`Environment` helpers, and `LinqMatcher`/DynamicLinq removal. The suite uses no response templating at all and no matcher beyond `MatchBehaviour`.

## Verification

Build: **0 warnings, 0 errors**. Restore: **0 audit warnings**. AOT publish clean. App suite 528/528.

Test-suite counts are noisy on macOS and I could not separate signal from churn locally — flagging rather than glossing:

- Integration 224/225, with a **different** test failing on each run (passes in isolation).
- Unit 54 failed vs **50** on an `origin/main` baseline measured on the same machine: 47 failures shared, 7 branch-only, 3 main-only. Both sides have unique failures, which reads as order-dependent shared state rather than a regression — the bulk is the known macOS cluster (#543).
- One thing worth a second look in CI: `AgentOrchestratorVendorTests` run in isolation fails 5/244 here vs 1/244 on the baseline.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -358,6 +358,7 @@ dotnet_diagnostic.ca1001.severity = suggestion ; Types that own disposable field
 dotnet_diagnostic.ca1707.severity = none       ; Identifiers should not contain underscores
 dotnet_diagnostic.ca1816.severity = none       ; Dispose methods should call SuppressFinalize
 dotnet_diagnostic.ca1861.severity = none       ; Avoid constant arrays as arguments
+dotnet_diagnostic.cs8602.severity = suggestion ; Dereference of a possibly null reference
 
 ; =============================================================================
 ; Native PTY shim
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -137,6 +137,7 @@ This is a public repository — we develop in the open.
 ## Dos and donts
 
 - DO use `JsonElementExtensions` instead of checking JSON value kind.
+- DO capture console output in tests with `ConsoleOutput.StartCapture()` / `StartErrorCapture()` (`test/Capacitor.Tests.Helpers`), never a hand-rolled `Console.SetOut`/`SetError` save-restore — TUnit0055 is an error. Console is process-global, so every caller needs bare `[NotInParallel]`; a group key is not enough.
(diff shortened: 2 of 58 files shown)


Inok:
Duplicate of #559, which already tracks this branch. Retargeting #559 onto `pavel/warnings-src` instead.

PR #582, opened 2026-08-18
[AI-1761] Codex app-server: hosted-agent runtime, transport & transport wiring

### LINKED ISSUE #581: Codex app-server: hosted-agent runtime + transport + factory
Daemon-side runtime for hosting unattended Codex reviewers over `codex app-server` (JSON-RPC over stdio) instead of the interactive PTY, behind a `codex.transport` config default of `pty`. Builds on the launch-input layer (#577). Adds the JSON-RPC transport, the control-plane runtime (handshake → hook-trust preflight → thread/start → per-round turn/start, turn-completion waits, always-decline approval bridge, backpressure retry, usage capture, reaper feed), and the composite codex factory that routes review-flow launches to app-server when active and delegates everything else to the PTY factory.

Linear: AI-1761 (epic AI-1759).

realtonyyoung:
Hosts unattended **Codex reviewers over `codex app-server`** (JSON-RPC 2.0 over stdio) instead of the interactive PTY, behind a `codex.transport` config that defaults to `pty` (so this is fully inert until an operator opts a daemon in). Builds directly on the launch-input layer (#578).

Closes #581. Linear **AI-1761** (epic AI-1759).

### What's here (4 commits)

1. **Transport** — `CodexAppServerConnection`: newline-delimited JSON-RPC 2.0 over stdio, a lean mirror of the proven `AcpConnection` (id correlation, single write-gate, one resilient read loop, always-answered server requests), reusing the shared envelope records and dropping ACP's reconnect-latch machinery a one-shot unattended reviewer has no design for.
2. **Runtime** — `CodexAppServerHostedAgentRuntime`: control-plane only (transcript stays on the hooks + `kcap watch` rollout path, so `EmitsTerminalOutput=false`). Lifecycle: spawn → `initialize` (delta opt-out) → `hooks/list` trust preflight (proceed / seed-and-restart / fail-closed) → `thread/start` (resolved model + deterministic thread-id identity) → per-round `turn/start`. `WaitForTurnIdleAsync` resolves from `turn/completed` (and unblocks on transport death). Approvals pinned to `never`, so any server-initiated approval request is answered with a valid on-the-wire `decline` (never a JSON-RPC error) and logged at Error. Client→server sends are bounded-retried on the `-32001` backpressure rejection. Usage captured from `thread/tokenUsage/updated`; reaper fed via `AgentActivityClock`.
3. **Sandbox fix + live smoke** — driving the **real codex 0.146.0** binary caught a wire bug a fake couldn't: `thread/start.sandbox` is the coarse `SandboxMode` string, a *different* shape from `turn/start.sandboxPolicy`'s object (the object form is rejected there). Fixed and re-verified. Adds a gated (`KCAP_CODEX_APPSERVER_SMOKE=1`) live smoke that drives the no-model handshake surface through the production connection against the installed binary in an isolated `CODEX_HOME` — the schema-drift tripwire for a Codex bump (skips loudly in CI, which has no codex).
4. **Composite factory + wiring** — `CodexHostedAgentRuntimeFactory` is the single "codex" factory: routes review-flow launches to app-server when `DaemonConfig.CodexAppServerActive`, delegates everything else (all interactive launches, all launches under the PTY default) to the wrapped PTY factory byte-identically. `CodexTransportDecision` is the ONE rule (operator selection AND the pinned 0.146.0 floor), resolved once into that field and read by **both** the router and the `codex-appserver-unattended-v1` advertisement, so the advertised policy and the transport used cannot diverge.

### Testing
- **26 new unit tests + 22 factory/decision tests**, all green: transport (15), runtime lifecycle vs an in-process fake app-server (10), transport-decision truth table + factory routing (delegation, app-server route via seam, fail-closed on missing hooks).
- **Live-validated against real codex 0.146.0** (the gated smoke).
- Existing capability-advertisement suites stay green (default config advertises `codex-unattended-v1` unchanged). **AOT clean** (no IL2026/IL3050).
- Grounded on the authoritative schema generated from the pinned binary.

### Scope / follow-ups
- Wires app-server for **review-flow (unattended) launches only**; interactive stays PTY (that's AI-1762).
- The vendored-schema (Q7) conformance test and the server-side `codex-appserver-unattended-v1` classifier (AI-1761 server half / PR 3) are follow-ups; the server already accepts the new policy string via its existing unknown-string handling, so this daemon is safe against an old server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1302,6 +1302,18 @@ KCAP_CLAUDE_PATH=/opt/claude/bin/claude kcap daemon
 KCAP_CODEX_PATH=/opt/codex/bin/codex  kcap daemon
 ```
 
+##### Codex transport (`KCAP_CODEX_TRANSPORT`)
+
+Hosted Codex **reviewers** (review flows) can run over the `codex app-server` JSON-RPC protocol instead of the interactive PTY. It is opt-in and env-only:
+
+```bash
+KCAP_CODEX_TRANSPORT=app-server kcap daemon   # opt in; default is `pty`
+```
+
+- **Default `pty`** — the interactive-terminal path, unchanged.
+- **`app-server`** takes effect only for review-flow (unattended) launches and only when the installed Codex meets the minimum version (**0.146.0**); a lower build falls back to `pty` automatically. Interactive launches always use `pty` in this release.
+- Rollback is a restart with the value flipped back to `pty`; it governs new launches only.
+
 The Cursor CLI path (`cursor-agent` by default, used to spawn the `cursor` hosted-agent vendor) is env-only for now — there is no `daemon.cursor_path` profile key yet, so set it per-launch:
 
 ```bash
--- a/src/Capacitor.Cli.Daemon/DaemonConfig.cs
+++ b/src/Capacitor.Cli.Daemon/DaemonConfig.cs
@@ -122,6 +122,18 @@ public class DaemonConfig {
     public string ClaudePath { get; set; } = "claude";
     public string CodexPath  { get; set; } = "codex";
 
+    /// <summary>Transport for hosted Codex reviewers: <c>pty</c> (default) or <c>app-server</c>.
+    /// Set via <c>KCAP_CODEX_TRANSPORT</c>. Governs NEW review-flow launches only; interactive
+    /// launches always take the PTY path in this phase. The effective decision (this selection AND
+    /// the version floor) is resolved once into <see cref="CodexAppServerActive"/>.</summary>
+    public string CodexTransport { get; set; } = "pty";
+
+    /// <summary>Resolved once at startup: true when <see cref="CodexTransport"/> is <c>app-server</c>
+    /// AND the installed Codex meets the app-server version floor. Read by BOTH the launch router and
+    /// the certification advertisement so the advertised policy and the transport used are one fact.
+    /// Never set from config directly.</summary>
+    public bool CodexAppServerActive { get; set; }
+
     /// <summary>
     /// Path or bare command for the Cursor CLI's ACP entry point, spawned as
     /// <c>{CursorPath} acp</c> by <c>AcpHostedAgentRuntimeFactory</c>. Overridable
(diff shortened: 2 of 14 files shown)


realtonyyoung:
Thanks — addressed the Qodo review (0 bugs, 6 rule violations) in the latest commit:

**Fixed:**
1. **Codex logic escapes vendor folder** — moved the transport/version-floor resolution into `CodexTransportDecision.ResolveActive`; `DaemonRunner` now only reads `KCAP_CODEX_TRANSPORT` and registers, and the startup comment is trimmed to the *why*.
2. **ValueKind vs JsonElementExtensions (runtime)** — the app-server runtime now uses the shared `Obj`/`Arr`/`Str`/`Num` helpers, removing its duplicate local `Str`/`Long`.
3. **TempDir paths built manually** — the factory tests now use `TempDir.CreateFile`/`PathTo`.
5. **KCAP_CODEX_TRANSPORT undocumented** — added to the README daemon-config section (default `pty`, `app-server` opt-in, 0.146.0 floor, review-flow-only, rollback).
6. **Comment restates code** — folded into #1; the narration is gone.

**Kept, with rationale:**
- **(2) ValueKind in `CodexAppServerConnection`** — the connection is a deliberately close mirror of its sibling transport `Acp/AcpConnection.cs`, which uses direct `ValueKind` checks throughout; keeping them identical keeps the two transports diffable. The extensions are adopted in the runtime, where they removed real duplication.
- **(4) `[NotInParallel("HomeEnvVarMutation")]` keyed** — this matches the established repo pattern: the sibling `CodexConfigWriterTests` uses the same `"HomeEnvVarMutation"` key, so all HOME/CODEX_HOME-mutating tests serialize against *each other* under one group (a bare `[NotInParallel]` would coordinate differently). The CLAUDE.md bare-`[NotInParallel]` rule is specifically about process-global **console** capture (TUnit0055).

56 Codex unit tests + the gated live smoke against real codex 0.146 stay green; AOT clean.

PR #592, opened 2026-08-19
Correlate CLI, web and server analytics: the CLI half

### LINKED ISSUE #574: Correlate CLI, server and web PostHog events with one join key
## Problem

kcap-web (browser), kcap-cli, kcap-web's Worker and kcap-server all publish to the **same** PostHog project, but each uses a different `distinct_id`:

| Producer | `distinct_id` |
|---|---|
| Browser | posthog-js anonymous id, then `identify(email)` on the org form |
| CLI | random GUID in `telemetry-device.json` |
| kcap-web Worker | the tenant slug |
| kcap-server | `HMAC-SHA256(PostHog:IdSalt, userId)` |

So one person moving from website → `kcap setup` → workspace → recorded sessions is stored as four unrelated people, and "which acquisition channel produces activated workspaces" is unanswerable. `www.kurrent.io` and `capacitor.kurrent.io` additionally mint separate web persons, because posthog-js is initialised with no `cookie_domain`.

## Approach

One opaque per-auth-run key (`join_id`, 32 hex chars, grants nothing) published as a small number of **bridge facts**, each pairing the key with one identity. No storage and no read-model change: the join is resolved at query time, because every producer already stamps its own `distinct_id`.

The browser is reached by replacing the CLI's dead-end loopback closing page with a redirect through two `/api/cli/return` hops and back to the same local page. `/api/*` is excluded from A/B arm assignment, so a CLI user observes the experiment without joining its sample.

Full design: `docs/superpowers/specs/2026-08-16-cli-web-event-correlation-design.md` (ships with the kcap-cli PR).

## Scope

**kcap-web** (ships first — the CLI redirects to this route, so it must exist before the CLI ships)
- [ ] New `src/pages/api/cli/return.ts` + `src/server/cli/return.ts`: validate `j`/`p`, consent gate, capture `cli_auth_return`, 302 onward
- [ ] `referrer-policy: no-referrer`, `cache-control: no-store`, `x-robots-tag: noindex, nofollow` on both hops
- [ ] `src/server/signup/provision.ts`: accept optional `joinId`, stamp `join_id` on `signup_reserved` / `signup_blocked` / `signup_flagged`
- [ ] Test that `isApiPath("/api/cli/return")` is true, so the path can never enter arm assignment
- [ ] Privacy-policy update

**kcap-server**
- [ ] `CliSetupPayload` / `RecordCliSetup` / `UserCliSetupCompletedEvent` gain optional `JoinId`
- [ ] `PosthogEventMapper.CliSetup` stamps `join_id`
- [ ] README analytics section

**kcap-cli** (ships last)
- [ ] New `SetupJoin` in `Capacitor.Cli.Core/Telemetry/`: mint at `setup`/`login` entry, single-use `Accept` with per-field validation
- [ ] `LoopbackBrowser`: redirect script on the closing page, listener-ownership contract, background drain, response headers
- [ ] `joinId` on the provisioning request and the cli-setup ping (camelCase on the wire)
- [ ] README Telemetry section

## Requirements worth calling out

- **Consent-gated.** No web-side bridge is created for a visitor who did not accept cookies: nothing read, nothing captured, nothing passed on. The redirect still happens either way — it is how the person gets back to their terminal. Gate the capture, not the hop.
- **`$device_id`, never `distinct_id`,** is what travels back to the CLI. After `identify(email)` the `distinct_id` *is* an email address; the `^[A-Za-z0-9_-]{1,64}$` check is a fail-closed guarantee against importing one.
- **`cli_auth_return` is not a sign-in event.** The loopback closing page is written before state validation and before the token exchange, so it means only "a callback arrived with no `error=`". Anything measuring "signed in" must pair the key with `cli_setup_signin_completed` or `signup_reserved`.
- **Opt-out suppresses everything.** `KCAP_TELEMETRY=0` / `DO_NOT_TRACK` / persisted opt-out → no key, no redirect, closing page identical to today.
- **Query contract.** The server-side bridge only covers users whose *first* `kcap setup` happens after rollout (`RecordCliSetup` is once-only). The spec lists every population that must be excluded from an activation metric.
- **Not solved:** cross-device. Someone browsing on a phone and installing on a laptop stays two people.

## Blocking prerequisite

A ~15-minute browser check before any implementation: posthog-js runs `persistence: "localStorage+cookie"`, and whether the **cookie** carries `$device_id` cannot be determined from source. Read it in three states — anonymous, after `identify(email)`, and after withdrawing consent (the gate uses cookie presence as its consent signal). If `$device_id` is absent from the cookie, the browser half of the design must change shape.

## Docs

Per the standing same-PR rule, and because the current wording stops being accurate: the CLI first-run notice and `README.md` Telemetry section change from *anonymous* to *pseudonymous* and state that CLI usage can be associated with a workspace and its creator. Retention for CLI telemetry needs deciding in the same pass.

stktung:
The CLI half of the CLI↔web↔server event correlation work. Design: `docs/superpowers/specs/2026-08-16-cli-web-event-correlation-design.md`, which ships with this PR per the repo convention.

Refs `kurrent-io/kcap-cli#574` · `AI-1999`

> **Not `Closes #574` on purpose.** That issue is the umbrella for three repos and the kcap-server slice (three lines: an optional `JoinId` on the cli-setup payload, and stamping `join_id` in the event mapper) has not been built. Closing here would orphan it. Worth a follow-up issue before this merges.

## Why

Web, CLI, kcap-web's Worker and kcap-server all publish to the **same** PostHog project but each stamps a different `distinct_id`, so one human is recorded as four unrelated people — and "which acquisition channel produces activated workspaces" is unanswerable. This mints one opaque per-auth-run key and publishes it as a small set of bridge facts, resolved at query time. Nothing is stored and there is no mapping table.

## Ship order — this must merge last

`kurrent-io/kcap-web#171` deploys the page this redirects to. **If a released CLI points at a route that does not exist, people hit a browser error in the seconds after signing in.** kcap-server can land any time; unknown JSON members are ignored by default.

Separately, the disclosure wording and CLI-telemetry retention decision are tracked in `kurrent-io/kcap-web#168` and gate the **release**, not this merge — nothing collects until a released CLI redirects.

## What changes

| | |
|---|---|
| `Telemetry/SetupJoin.cs` *(new)* | Mints the key, publishes it as a shared property, builds the URL the closing page navigates to, and validates what comes back |
| `Auth/LoopbackBrowser.cs` | The listener now outlives `InvokeAsync`, behind an explicit ownership contract |
| `Auth/OAuthLoginFlow.cs`, `Auth/WorkOSDiscovery.cs` | Disposal ownership at the three construction sites |
| `Auth/ProvisioningModels.cs`, `Auth/TenantProvisioningClient.cs`, `Commands/SetupCommand.cs` | `joinId` on both request bodies |
| `Telemetry/CliTelemetry.cs`, `Telemetry/TelemetrySpool.cs` | The two guards that keep the key out of logs and off disk |
| `README.md`, `Resources/help-config.txt` | anonymous → pseudonymous |

## The parts worth reviewing carefully

**The listener now outlives the call that created it.** Previously its lifetime was exactly one function call, guaranteed by `using var` — the compiler tore it down on every path including exceptions. The closing page now navigates the browser out to kcap-web and back to `/joined` on that same listener, so that guarantee is gone and replaced by a hand-written protocol. Four parties can decide it is finished; exactly one must act. Twice throws, never holds the port for the process lifetime. Handoff happens at one point only: after a successful callback whose redirect was written cleanly. Both deadlines enforce themselves *by* tearing down rather than abandoning the wait, because an abandoned accept keeps the port alive with nobody watching.

**`cli_auth_return` is deliberately a weak event.** The closing page is written before the CSRF state check and before the token exchange, so it means exactly "a callback arrived carrying no `error=`" — never "signed in". A run that goes on to fail authentication entirely still emits the redirect. Anything measuring "signed in" must pair the key with `cli_setup_signin_completed` or `signup_reserved`.

**Only `$device_id` ever crosses back, never `distinct_id`.** After the web signup form calls `identify(email)` the site's own id **is** an email address. The `[A-Za-z0-9_-]{1,64}` character class is the fail-closed enforcement, asserted directly: an email-shaped value reaches nothing, and the merged property bag contains no `@` at all.

**Two wire names that look like a typo are deliberate.** `joinId` camelCase on both request bodies; `join_id` snake_case as a PostHog property. `CapacitorJsonContext` is globally `SnakeCaseLower`, so the provision DTO needs an explicit attribute or it silently ships under the wrong name and kcap-web never sees it — nothing errors. The cli-setup ping stays a hand-built literal for the mirror-image reason.

**Where the key is minted decides most of the coverage.** At command entry, before any funnel event and before a lane is chosen. Minting inside an auth lane would leave `cli_setup_started` — the funnel's own denominator — without the key; minting inside the loopback lane would mean device-code, headless and no-browser runs never get one. Neither failure throws or turns a test red, so a guard reads the source and checks the position.

## Testing

`SetupJoinTests` 24, `LoopbackBrowserTests` 15, `LoopbackOwnershipTests` 5, `TelemetrySpoolTests` 11, `CliSetupPingBodyTests` 5, plus additions to the wire, provisioning, funnel, disclosure and placement suites. NativeAOT publish clean.

**A live end-to-end run passes** (`JoinChainLiveTests`, gated behind `KCAP_JOIN_E2E=<base url>`): the real listener, the real closing page, the real `/api/cli/return` route over HTTP, both hops, the real return, the real merge, and the debug placeholder. It is the only case that can catch a mismatch *between* the two halves — a renamed parameter, a wrong host rule, a redirect that never comes back — which no unit test on either side can see.

Two source-reading guards exist where the behavioural form would bind a port or launch a browser: the mint position, and that no construction site leaves a browser inline or omits the collaborator.

## Review history

A codex round on this branch found a real defect: the key is attached as a shared property, so a failed telemetry flush parked it in `telemetry-spool.jsonl` on disk — contradicting a guarantee stated in the spec, the comments and several commit messages. Fixed in `0ca416b`; both escape routes now key off one constant. The pre-existing test asserting the key never reaches a file had passed throughout, because it substitutes a fake collector and the spool is unreachable on that path.

## Known-flaky, not caused by this branch

The full unit suite reports ~11 failures in ACP orchestration, import chains, session-start memory injection and spool sweep. Each passes in isolation, none is in a subsystem this branch touches, and an earlier run of identical code failed a different 9. Not verified against the base commit.


realtonyyoung (changes requested):
**[P1] Mint before telemetry initialization emits the one-time first-run event**

`Program.cs` calls `CliTelemetry.Initialize(...)` before dispatching to either auth handler. On a fresh installation, `Initialize` immediately calls `NoticeAndFirstRun()`, which captures `cli_first_run`; only afterwards does `SetupCommand.HandleAsync` or `LoginCommand.HandleAsync` call `SetupJoin.Mint()`. Consequently `cli_first_run` permanently lacks `join_id`, contradicting the PR’s event contract that every `cli_*` event for the auth run carries the key. Because this event is emitted only once, a later run cannot repair the missing correlation.

Please ensure the key is minted for `setup`/`login` after telemetry is enabled and the shared bag is initialized, but before `NoticeAndFirstRun()` can capture anything—for example, conditionally inside `Initialize` for those two commands—while keeping non-auth commands unkeyed.

PR #603, opened 2026-08-19
[AI-2033] Install the agent skills from every vendor's own plugin install

George-Payne:
`kcap plugin install --cursor` wrote hooks and registered MCP servers but installed no skills — same for `--copilot`, `--gemini`, `--pi` and `--opencode`. Only `--codex`, `--skills` and a full `kcap setup` wrote the shared `~/.agents/skills` tree, so five of nine vendors had no `guided-tour` and no `recap`, and the tour handoff named a slash command their agent had never heard of. The target flags are mutually exclusive, so `--cursor --skills` cannot be one invocation — nothing but the install itself can close the gap.

`kcap setup` already treats this as policy and says so in `CodingAgentsStep.HandleAgentSkills`: install the shared tree when any non-Claude agent is detected, *independent of whose hooks were written*. This makes `plugin install` obey the rule setup already documents.

- **`PluginCommand`** — one `InstallVendorSkillsAsync` replaces the two near-identical private helpers Kiro and Antigravity had, rather than adding a third and fourth copy. Their behaviour is unchanged (they still write their own trees, which is what they read), and both are now covered by a test — neither was before.
- **A refresh tops a tree up; it never creates one.** `plugin remove --skills` drops the marker precisely so an upgrade cannot undo it, and `refresh.js` runs the `--if-installed` form of every vendor on each `npm install -g` with output discarded. Without this gate a deliberate removal came back silently, on a command the user never ran. This was caught in review, not by me — the first version of these tests asserted the buggy behaviour.
- **The legacy `~/.codex/skills` sweep** now runs here too, as it does for the shared tree's other two writers. Skipping it let a Cursor-first install stamp the marker current, so the next `--skills --if-installed` fast-pathed past the sweep and the stale directory survived.
- **Skills failing is a warning, not a failed install** — hooks and MCP are what make capture work. Note this differs from `--codex`, which fails the install; see the open question below.
- **Docs** — `README.md` per-command sections and code fences, `help-plugin.txt` per-vendor entries, and the `refresh.js` contract comment, which no longer described what the code does.
- **Tests** — the fresh path (the behaviour being fixed), that a refresh creates nothing for a vendor never installed (the npm-postinstall safety property, previously unpinned and unpinnable — the sibling tests use `ResolvePluginPath: () => null`, so they would pass straight through the regression), and that a removal is not resurrected. The last fails on four of the five vendors without the gate above. `EnvScope` moves to `Helpers` rather than gaining a fifth copy.

- **An opt-out per vendor** — `--skip-cursor-skills`, `--skip-copilot-skills`, `--skip-gemini-skills`, `--skip-pi-skills`, `--skip-opencode-skills`. Kiro and Antigravity already had one and every other artifact these installs write has a skip flag (15 of them, all `--skip-<vendor>-<artifact>`), so the gap was one this branch created: before it the five wrote no skills, so there was nothing to decline. Vendor-scoped rather than a shared `--skip-skills`, which would break that naming and read as a sibling of `--skills` — a target flag, not a skip.

Not changed here, pre-existing, and now filed as AI-2086: `remove --codex` deletes the shared tree while the other seven removes leave it, so removing Codex on a Codex+Cursor machine takes Cursor's skills with it. Six commands write that tree, two remove it, nothing tracks ownership. This branch widens the blast radius rather than causing it.

AI-2033

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ In `--no-prompt` mode, the wizard installs hooks for every detected agent by def
 > **Behavior change: `--no-prompt` now also imports this repo's history.** The Step 6 import (above) defaults to yes like every other prompt in the wizard, so `kcap setup --no-prompt` now uploads this repository's past sessions too — when run inside a git repo with an origin remote and authentication requirements are satisfied (including no-auth/provider-`None` servers). Existing unattended/scripted `kcap setup --no-prompt` invocations will start uploading current-repo session history unless you add `--skip-import`.
 
 > **Need hooks for an agent installed after setup, or scoped to a single repo?**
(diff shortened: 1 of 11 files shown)


realtonyyoung:
no findings

realtonyyoung (approved):
Approved — no findings.

PR #605, opened 2026-08-19
[AI-2027] Open the browser from setup and poll a machine pairing

### LINKED ISSUE #604: kcap setup should open the browser and poll a machine pairing
**Canceled 2026-08-19 — duplicate of** [AI-2027](https://linear.app/kurrent/issue/AI-2027/kcap-setup-opens-the-browser-and-polls-the-pairing)**, which tracks this work.**

Not a real intake. I filed GitHub issue kurrent-io/kcap-cli#604 on the mistaken belief that a `kcap-cli` PR required one, and Linear imported it straight back as this issue. The rule is the opposite: we file Linear tickets in both repos, and GitHub issues are for external contributors. `CLAUDE.local.md` now says so.

The work shipped as [AI-2027](https://linear.app/kurrent/issue/AI-2027/kcap-setup-opens-the-browser-and-polls-the-pairing) (PR kurrent-io/kcap-cli#605).

---

`kcap setup <tenant>` configures a machine entirely in the terminal today. The server now offers a machine-pairing channel so the browser can carry the parts of first-run setup that are better done there, and the CLI needs the other half of it.

What `kcap setup` should do when it has been given a tenant and that server offers the channel:

* Mint a pairing, print the human code **and** the fallback URL, open the browser, then poll.
* Print the code every time. The browser showing a code is only half a comparison — the half that makes it a defence is the one printed by the machine being paired, and without it there is nothing to check against.
* Assert that the account that approved is the account the CLI then authenticates as, and abort otherwise. The channel carries approval and never a credential, so nothing else binds the two.

It has to stay invisible where it cannot help: a server that does not serve the routes, a headless environment with no browser to open, `--no-prompt`, and the `None` auth provider all keep today's behaviour exactly.

George-Payne:
The CLI half of the machine-pairing channel. `kcap setup <tenant>` (or `--server-url`) now mints a pairing on that server, prints the human code beside a fallback link, opens the browser, and waits for someone to approve before going on to sign-in.

- **`Auth/BrowserPairingFlow`** — mint, show, open, poll. The code is printed in the terminal **every time**: the browser showing a code is only half a comparison, and the half that makes it a defence is the one printed by the machine being paired. The fallback link is equally non-optional, because nothing can confirm a browser actually opened.
- **`Auth/PairingIdentity`** — asserts that the account which approved is the account the CLI then authenticates as, and stops setup if it is not. The channel carries approval and never a credential, so nothing else binds the two. Reads the claims each provider actually mints — `github_id` plus a `github|{n}` subject for GitHub, `sub` verbatim for WorkOS — because `kapacitor:user_id` never goes on the wire; a claims transformation adds it per request. Distinguishes "somebody else approved" from "this build cannot tell", which need different remedies.
- **`Auth/PairingPoll`** — a pure classifier in the shape of `ProvisioningPoll`, so every branch is covered without a server. `401` is terminal here rather than transient: there is no token source to refresh on the next tick, the secret was minted once.
- **`Auth/PairingClient`** — degrade-don't-throw on `TenantProvisioningClient`'s convention, behind an `IPairingChannel` seam so the loop is testable.
- **`Auth/SystemBrowser`** — hoisted out of `LoopbackBrowser`'s private opener so there is one, and one place saying why it is best-effort.
- **`Commands/SetupCommand`** — Step 1b between server resolution and login; the identity assertion immediately after login; `/complete` last, because completion invalidates the secret and so closes the channel this machine reports progress on.

**Availability needs no flag and no version check.** The routes exist only when the tenant has `Features:FirstRunSetup` on, so a `404` on mint *is* the oracle — as are `401`/`403`/`405`, which a gateway can answer on an unmapped anonymous route and which all mean the same thing and have the same remedy. Headless, `--no-prompt` and the `None` provider skip it for the same reason. In every one of those cases setup behaves exactly as it does today.

**Reviewed before opening.** Three passes (general, architecture, conventions) over the first commit; the second commit is what they found. Worth calling out two:

- The identity check read claims **no token has ever carried**, so it would have aborted setup for every user on every GitHub tenant, every time. The tests missed it because they synthesised the claim shape instead of copying it — the fixtures now come from the server's own token construction.
- Reading claims without an object guard crashed on a valid JWT whose payload is not a JSON object, because `TryGetProperty` throws there and only `Format`/`Json` exceptions were caught.

Also fixed: the poll compared the server's `expires_at` against the local clock (a fast clock gave up without polling once, on a live pairing); an approval naming no approver returned a result that degraded to "carry on without the check"; `/complete` discarded the `403` that *is* the server disagreeing about who approved; a mint missing `expires_at` aborted setup claiming expiry.

`kcap setup` gains no new flags, and `--no-prompt` behaviour is unchanged.

**Testing.** 30 new tests over the flow, the client wire shape and the identity comparison — the deadline, the back-off, the guards and the header/body casing, none of which had coverage. 1949 pass in `Capacitor.Cli.Core.Tests.Unit`; AOT publish is clean of IL2026/IL3050.

**Deliberately left.** The check lives in `SetupCommand` rather than `OnboardingFacade`'s commit boundary. Moving it means threading an expected identity through `LoginAsync`, which every caller shares — worth doing when the Avalonia wizard is a real surface, not on this ticket.

Closes #604

AI-2027


src/Capacitor.Cli.Core/Auth/BrowserPairingFlow.cs
```diff
@@ -0,0 +1,155 @@
...
+    ///
+    /// <para>It goes to a shell-executed open, so an unchecked one is not merely a bad link: a
+    /// <c>file://</c> path or a registered custom scheme would be launched, by a server the CLI has
+    /// no reason to trust that far. The origin test is the second half — a page on somebody else's
+    /// host is where a human would be asked to approve somebody else's pairing.</para>
+    /// </summary>
+    static bool IsOpenable(string setupUrl, string serverUrl) =>
+        Uri.TryCreate(setupUrl, UriKind.Absolute, out var parsed)
+     && (parsed.Scheme == Uri.UriSchemeHttp || parsed.Scheme == Uri.UriSchemeHttps)
+     && ServerIdentity.SameServer(parsed.GetLeftPart(UriPartial.Authority), serverUrl);
```
realtonyyoung: [P1] This reduces `setupUrl` to its authority before calling `SameServer`, so it discards the configured server path. `ServerIdentity` explicitly treats paths as significant for path-routed deployments: with `--server-url https://host/tenant-a`, a mint response such as `https://host/tenant-b/setup?p=...` passes this check and is opened, even though it belongs to a different server/tenant. Validate the setup URL against the configured server base (preserving its query string) and add a path-routed regression case.

George-Payne: Good catch, thank you — and it was wrong in both directions, which makes it worse than a missed rejection.

Reducing to the authority means `https://host/tenant-a`'s own `https://host/tenant-a/setup?p=…` canonicalizes to `https://host:443` and no longer matches its base `https://host:443/tenant-a`, so a path-routed deployment had **every** legitimate setup URL refused. And as you say, nothing in that comparison separates `tenant-b` from `tenant-a`. Authority-only matching wasn't a check worth having either way.

Fixed in 5c9f2dc: both sides go through `ServerIdentity.Canonicalize` whole, and the target must equal the base or sit under `base + "/"` — the trailing separator so `/tenant-a` can't match `/tenant-abc`. Query and fragment come off before canonicalizing (`Canonicalize` refuses a base carrying a query, and `setup_url` always has one); the URL actually opened keeps them.

Also refused userinfo while I was in there — the fallback link is printed for a human to read, and `https://acme.kcap.ai@evil.example.com/` reads as one host while addressing another.

Regression cases added: path-routed base accepts its own page, rejects a neighbour's, rejects one above the base, rejects the `/tenant-abc` prefix collision, rejects userinfo.

realtonyyoung (approved):
Approved — verified the addressed setup URL validation finding; no additional actionable issues.

PR #607, opened 2026-08-19
[AI-2034] Add --skip-title to import, matching watch

George-Payne:
`kcap import` titles every Claude and Codex session it loads by shelling out to the user's own `claude` (15s) / `codex` (30s), three abreast, with no way to turn it off. At Everything scope that is hundreds of headless calls on their subscription. `kcap watch` has had `--skip-title` for the identical generator all along — `TitleGenerator` is deliberately shared between the two paths — so this is the missing half of a pair rather than a new idea, and it takes the same name.

- **`ImportCommand`** — the flag gates the title task before it is scheduled, next to the `SupportsTitleGeneration` check that answers the same question for a different reason. Counters, the Spectre progress task and `RanBackground` all fall out correctly, since nothing is ever queued.
- **The Done grid's Titles row** now keys off whether titling was *requested* rather than whether any background work ran. With `--skip-title --generate-summaries` the old guard printed `Titles  0 generated, 0 skipped, 0 failed`, which reads as titling that found nothing rather than titling that never happened. That case only exists because of this flag, so the fix belongs here.
- **Summaries are untouched.** They are opt-in through their own flag, and someone who typed `--generate-summaries` has asked for that spend. Worth knowing both lanes share one `SemaphoreSlim(3)`, so the concurrency limit is across all background work, not per kind.
- **Docs name the vendors this actually affects**, because it is easy to assume all nine: `SupportsTitleGeneration` is true for **Claude and Codex only**. The other seven never title locally — Cursor and Copilot forward the name their transcript carries, OpenCode its native title, and Antigravity, Gemini, Kiro and Pi leave the server to derive one — so the flag changes nothing for them. `README.md` and `help-import.txt` both say so.
- **`VendorSelectionTests.global_flags_pass_through`** gains `--skip-title` (and `--reimport`, which was already missing). That test exists so a new global flag cannot be mistaken for a vendor typo.

On the test: it shadows `claude` on PATH with a one-line script rather than letting the real one answer. The positive control has to prove a title really is posted *without* the flag, or the skip assertion proves nothing — and reaching a live model would be slow, unrunnable on CI, and billed to whoever ran it. That control is Unix-only: the prompt arrives as a single ~1.4KB argv element full of newlines, which `cmd.exe` cannot hand to a `.cmd` shim intact, so on Windows it would fail for a reason unrelated to the flag. The skip assertion itself runs everywhere.

Not changed, deliberately: `kcap setup`'s embedded import (`SetupCommand.DefaultImportRunner`) does not pass the flag, so setup still titles the current repo's history. That is one repo rather than a full backfill, and changing it is a product decision rather than part of adding the flag — but it is the one remaining unattended path that spends a first-time user's quota, so it is worth a decision rather than an omission.

AI-2034

### DIFF
--- a/README.md
+++ b/README.md
@@ -655,10 +655,13 @@ kcap import --org EventStore --private        # mark every imported session as O
 kcap import --org EventStore --since 2026-01-01  # only sessions on or after this date
 kcap import --org EventStore --cwd /path/to/project  # filter by working directory
 kcap import --org EventStore --session abc123    # single session
+kcap import --org EventStore --skip-title    # don't spend your own agent quota on titles
 kcap import --opencode --session ses_x --reimport  # force one OpenCode session past its ledger entry
 ```
 
 `--reimport` forces OpenCode sessions to re-import even when the local completeness ledger (described above) records them as already loaded — the escape hatch for a session that was deleted server-side (e.g. via `kcap disable`) but is still marked complete locally, which a plain re-run would otherwise skip. Scope it with the usual vendor/`--repo`/`--cwd`/`--session` filters to force just the affected sessions; the re-send is idempotent, and a successful forced import refreshes the ledger entry. It has no effect on other vendors, which already re-classify every run.
+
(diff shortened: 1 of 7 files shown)


realtonyyoung:
Reviewed the final PR head. No additional actionable findings.

realtonyyoung (approved):
Approved — reviewed the final PR head; no additional actionable findings.

PR #608, opened 2026-08-19
[AI-1992] Report a provisioning timeout as pending, not as a failed sign-in

### LINKED ISSUE #573: Surface workspace provisioning-in-progress as a distinct sign-in outcome
Deferred from the kurrent-io/kcap-cli#571 review rounds.

When workspace provisioning times out (the 4s × 150 poll expires with the tenant still provisioning), Core collapses the `ProvisionOffer.InProgress` outcome into `AuthResult.Failed`, so the wizard's Sign-in step shows the generic "Sign-in failed." headline. The accurate copy — "still provisioning — finish later by joining `<slug>` from the Connect step" — currently only reaches the detail line and log because the provisioner emits it through the progress sink.

Fix shape: carry the provision outcome through the result algebra (e.g. `AuthFailureReason.ProvisioningInProgress`, or `WorkOSDiscoveryFlow` carrying the `ProvisionOfferStatus`) so the wizard and `kcap setup` can headline it as a pending state rather than a failure. Touches `src/Capacitor.Cli.Core/Auth/OnboardingFacade.cs` / `WorkOSDiscovery.cs` and the Sign-in step's status mapping.

---

**Updated 2026-08-19: this is free-standing, and it matters more than it did.**

It was filed as a blocker for [AI-2031](https://linear.app/kurrent/issue/AI-2031/workspace-picker-restyle-the-proxy-picker-link-creation-out-to-signup) (screen 3), on the assumption that a browser "creating" state would surface the outcome. That assumption is gone — workspace creation stays with kcap-web's existing signup, so no browser screen of ours renders this. The blocking relation is removed.

What it fixes is unchanged and entirely CLI-side, which is precisely why it survives: **the terminal creation path is now the one that carries this case.** Someone who runs `kcap setup`, creates a workspace, and waits past the 4s × 150 poll is told "Sign-in failed." for a workspace that is provisioning correctly. The accurate copy already exists and only reaches the detail line.

One amendment to the fix shape: the description names "the wizard's Sign-in step" as a consumer, meaning the Avalonia wizard that [AI-2053](https://linear.app/kurrent/issue/AI-2053/retire-the-avalonia-onboarding-wizard) retires. The consumer that outlives this work is `kcap setup`'s own terminal rendering, so map the outcome there rather than only in the ViewModel.

Background: `~/notes/capacitor-onboarding/signup-handoff.md`.

George-Payne:
When WorkOS workspace provisioning outran its 4s × 150 poll, Core collapsed the `ProvisionOffer.InProgress` outcome into an undistinguished `AuthResult.Failed`, so the wizard's Sign-in step headlined **"Sign-in failed."** over a workspace that was on its way — sign-in had already succeeded to get that far. The accurate copy survived only as a detail line and a log entry.

* `AuthFailureReason` **gains** `ProvisioningInProgress`**.** It rides `Failed` rather than becoming its own `AuthResult` case deliberately: every consumer discriminates on `Committed` and funnels the rest into a catch-all (`SignInStepViewModel:422`, `LoginCommand:66`, `SetupCommand:934`, `App.axaml.cs:429`), so a new case would have landed silently in the very `default:` arm that prints the wrong headline. Riding the reason forces the one switch that had to change to change, and `SetupCommand:50` already sets the `{ Reason: … }` precedent.
* `ProvisionOffer` **carries the pending slug**, so a caller can name the workspace it is telling the user to come back to. `InProgress` becomes a factory, matching the file's existing split between outcomes that carry data (`Created`, `ExistingWorkspace`) and ones that do not.
* **The headline is the fact; the sink's line is what to do about it.** Both land in front of the same reader in a GUI, so the headline is deliberately not a paraphrase of the guidance. `WizardAuthBridges` reports that line as a `Notice` rather than an `Error`, since nothing went wrong.
* `App.axaml.cs` printed the same untruth to stderr when the window was closed mid-poll — narrow, but it is literally the claim this ticket is about.

`kcap setup` **needed no change, which is worth stating rather than implying otherwise.** Its failure path prints nothing for this reason (`SetupAuthProgress.ReportFailure` only speaks for `Unreachable`, and `OnboardingFacade.Stop` never prints), so the only line a terminal user sees is the provisioner's own yellow "Still provisioning" — already correct. The slug is passed there too for symmetry, but nothing reads it on that path yet.

`Declined` **is left generic, and deliberately not pinned by a test.** It is arguably the same defect — the user chose to decline and the step still says sign-in failed — but it is a separate decision from this one, so asserting the current behaviour here would lock the wrong answer in with a passing test. Happy to file it.

Tests: the discovery-level reason and slug, the reason surviving the flow→`AuthResult` mapping, and the wizard headline itself (`SignInStepViewModelTests`) — which is the user-visible half and had no coverage. Verified by reverting each production change in turn and confirming the matching test reddens. `WizardAuthBridgesTests` also now pins `PendingSlug`, which nothing did.

Closes kurrent-io/kcap-cli#573

kurrent-io/kcap-cli#573

### DIFF
--- a/src/Capacitor.App/App.axaml.cs
+++ b/src/Capacitor.App/App.axaml.cs
@@ -427,9 +427,13 @@ public partial class App : Application {
 
         switch (result) {
             // Sign-in succeeded and the workspace is on its way; closing the window mid-poll is not a
-            // failure, and stderr saying so would be untrue.
+            // failure, and stderr saying so would be untrue. The guidance is spelled out here rather
+            // than carried on the message: in the wizard it arrives through the progress sink, and
+            // repeating it there would say the same thing twice — but by now that view is gone, so
+            // this line is the only thing left telling the user how to resume.
             case AuthResult.Failed { Reason: AuthFailureReason.ProvisioningInProgress } pending:
-                Console.Error.WriteLine($"kcap: {pending.Message}");
+                Console.Error.WriteLine(
+                    $"kcap: {pending.Message} Join it from the Connect step once it is ready.");
                 break;
             case AuthResult.Failed failed:
                 Console.Error.WriteLine($"kcap: onboarding sign-in ended with a failure: {failed.Message}");
--- a/src/Capacitor.App/App.axaml.cs
+++ b/src/Capacitor.App/App.axaml.cs
@@ -426,6 +426,11 @@ public partial class App : Application {
         var result = await attempt.Result.ConfigureAwait(false);
 
         switch (result) {
+            // Sign-in succeeded and the workspace is on its way; closing the window mid-poll is not a
+            // failure, and stderr saying so would be untrue.
+            case AuthResult.Failed { Reason: AuthFailureReason.ProvisioningInProgress } pending:
+                Console.Error.WriteLine($"kcap: {pending.Message}");
+                break;
             case AuthResult.Failed failed:
                 Console.Error.WriteLine($"kcap: onboarding sign-in ended with a failure: {failed.Message}");
                 break;
--- a/src/Capacitor.App/Services/Onboarding/WizardAuthBridges.cs
+++ b/src/Capacitor.App/Services/Onboarding/WizardAuthBridges.cs
@@ -243,10 +243,12 @@ public sealed class WizardTenantProvisioner(
             }
         }
 
-        progress.Error($"Still provisioning — finish later by joining '{slug}' from the Connect step.");
+        // Notice, not Error: the workspace is being created, nothing has gone wrong. The reason on the
+        // result is what lets the step headline it as pending rather than failed.
+        progress.Notice($"Still provisioning — finish later by joining '{slug}' from the Connect step.");
         SetupFunnel.WorkspaceFailed("poll_timeout");
 
-        return ProvisionOffer.InProgress;
+        return ProvisionOffer.InProgress(slug);
     }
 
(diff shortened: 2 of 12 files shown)


src/Capacitor.App/App.axaml.cs
```diff
@@ -426,6 +426,11 @@ internal static async Task<bool> HandoffAfterWizardAsync(
         var result = await attempt.Result.ConfigureAwait(false);
 
         switch (result) {
+            // Sign-in succeeded and the workspace is on its way; closing the window mid-poll is not a
+            // failure, and stderr saying so would be untrue.
+            case AuthResult.Failed { Reason: AuthFailureReason.ProvisioningInProgress } pending:
+                Console.Error.WriteLine($"kcap: {pending.Message}");
```
realtonyyoung: **P2 — preserve the recovery instruction in the post-close handoff.** This stderr line is the only feedback after the wizard has closed: `UiAuthProgress.Notice` was routed to the now-closed view model, and `pending.Message` is only `'<slug>' is still being created.` It drops the provisioner's only actionable guidance — finish later by joining the workspace from the Connect step — that the active wizard displays as its detail. A user who closes at this point knows the workspace is pending but not how to resume setup. Include that recovery instruction in the handoff output (or carry it in the pending result).

alexeyzimarev (approved):

PR #621, opened 2026-08-19
[AI-2041] Name an agent session that was already running at first install

George-Payne:
A Kiro session running when kcap's integration first arrives loaded nothing, so it isn't captured live. A freshly started one loads it by itself — which is why the only case worth mentioning is that one, and why almost every install says nothing at all.

- **The trigger is a *first* install, not a timestamp.** Dating staleness from the installed file's mtime reads plausibly and is wrong: an install rewrites its own agent, so a months-old session that *is* being captured would be told it isn't — every time the installer re-ran, including on every npm upgrade, where the postinstall re-runs it with output discarded. On a first install the question needs no clock: nothing running can have loaded something that did not exist. The call site already computed that fact.
- **The line names the remedy that exists.** The transcript is on disk either way, so `kcap import --kiro` backfills the session once it ends. Saying only "it won't be recorded" would name a problem and conceal its fix — and the ticket's "no button" constraint is about the *restart*, which is genuinely undoable, not about withholding a non-destructive remedy the product already ships.
- **No restart offered.** That decomposes into kill and relaunch and only the destructive half is ours, on a terminal we don't own.
- **Nothing is claimed after an install that failed.** The processes are sampled before anything is written and reported only once the install actually landed — otherwise "anything you start from now on is captured" would be untrue in the direction that matters.
- **Kiro only, and that is the whole shippable surface.** A target has to be recognisable from a process name alone, which means a real binary with an exact name. `kiro-cli` qualifies. Pi does not, and the reason is structural rather than a missing refinement — see below.
- **The probe is injected through `PluginEnvironment`** rather than called statically, so the existing plugin tests don't sweep whatever the CI box happens to be running, and the wiring is testable at all.

**Pi was in this PR and is not any more.** It runs through a node shim, and node's `process.title` setter rewrites the argv region, so a live Pi process is either named `node` with `…node_modules/pi/bin/pi` in its command line, or named `pi` with a command line of just `pi` — never both. `pi` is a name anything can have, so it needs corroboration; the corroborating evidence only exists in the case where the name doesn't match. Name-matching and command-line corroboration are therefore mutually exclusive, and the target I had written could not fire on Linux. Verified in a container rather than reasoned about. The command-line reader added for it (`/proc/<pid>/cmdline`, `KERN_PROCARGS2`) came out with it — recognising Pi needs a signal neither of those provides, which is its own ticket.

Tests cover the rule, and separately the wiring — including that a re-install stays silent, which is the bug above, and that a failed install claims nothing. I verified the re-install case bites: removing the first-install gate reddens it and nothing else.

Also not done, deliberately: `kcap setup` doesn't route through these methods (it drives `CodingAgentsStep` directly), so its prompt-driven install doesn't report. The browser flow drives `plugin install --<vendor>` per selection, so the channel the ticket names *is* covered; setup is the terminal path and can follow separately.

AI-2041


realtonyyoung (approved):
Approved - no findings

PR #627, opened 2026-08-20
[AI-2123] Let the daemon suite finish a run outside CI

George-Payne:
Running `Capacitor.Cli.Daemon.Tests.Unit` the documented way — directly, as an executable — aborted with SIGABRT partway through, so most of its 2764 tests never reported. The one suite you can't iterate on locally was the one full of PTYs, sockets and launch lanes. Two causes, neither in the product.

- **`PiHostedLaunchTests`** asserted `KCAP_URL` is *absent* from the built `ProcessStartInfo`. That environment is seeded from the current process and the repo's own `.envrc` exports `KCAP_URL`, so the test was reading the developer's shell — it fails with the variable set and passes under `env -u`, same commit, same machine. `BuildPsi` only ever adds the key from the context and never removes an inherited one, which looks deliberate (a hosted agent falling back to the daemon's own server is reasonable), so the assertion moves to the invariant that actually holds: BuildPsi contributes nothing, whatever it inherited. Comparing against the inherited value needs no `EnvScope` — which would itself have required `[NotInParallel]` and raced the nine other daemon tests that touch `KCAP_URL`.
- **`PtySpawnTests`** caused the abort. It waits on a pid `pty_spawn` has already reaped, so the number is free for the OS to reassign, and a concurrent spawn landing on it turns that into a wait on someone else's child. Winning is worse than losing: `System.Diagnostics.Process` owns most of the others, and stealing one leaves .NET's SIGCHLD handler with `ECHILD`, which it answers by `FailFast`-ing the host. The `UnixSpawnerThread test-forced crash` trace in the log is collateral from that abort, not the cause. Bare `[NotInParallel]`, for the reason Console already has one.
- **`CLAUDE.md`** — both rules generalised, plus the invocation below.

Neither reproduces in isolation, which is why CI has never seen them: it passes `--maximum-parallel-tests 1`.

**That flag is doing more work than it looks.** Underneath the abort the suite has ~39 more tests that collide concurrently — measured, by running the unmodified suite serially (1 failure, the `KCAP_URL` one) against the same suite in parallel (39). It isn't a speed setting, it's the only thing holding the suite together, and it lived in `ci.yml` where nobody running the local loop would find it. Documented now, next to the command.

An assembly-level `[ParallelLimiter<T>]` would have moved that into the project instead of the command line, and I tried it. **It is not equivalent** — it serialises by wall-clock but still leaves `CodexLauncherTests` failing intermittently (4 one run, 2 the next; 59/59 in isolation), where the flag does not. Recorded in `CLAUDE.md` rather than shipped, so the next person doesn't spend the same afternoon on it.

Fixing those ~39 is a separate piece of work and deliberately not attempted here.

`Capacitor.Cli.Daemon.Tests.Unit` 2764 · 0 failed · 27 skipped · 1m 23s.

AI-2123

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -150,6 +150,13 @@ dotnet build src/Capacitor.Cli/Capacitor.Cli.csproj
 
 - Throwaway directories come from Helpers' `TempDir` — `using var tmp = new TempDir();` — never a per-class copy. Build paths under it with its own members — `tmp.PathTo(…)` for a path that must not exist yet, `tmp.CreateDir(…)`, `tmp.CreateFile(…)` — not `Path.Combine(tmp.Path, …)` + `Directory.CreateDirectory`/`File.WriteAllText`. A class that keeps one as a field must implement `IDisposable` and dispose it: CA1001 is an error, so an `[After(Test)]` hook will not build.
 - Capture console output with `ConsoleOutput.StartCapture()` / `StartErrorCapture()`, never a hand-rolled `Console.SetOut`/`SetError` save-restore — TUnit0055 is an error. Console is process-global, so every caller needs bare `[NotInParallel]`; a group key is not enough.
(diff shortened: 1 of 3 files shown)


realtonyyoung:
NO FINDINGS

realtonyyoung (approved):

PR #629, opened 2026-08-20
[AI-2043] Lead the import summary with what landed and what didn't

George-Payne:
Import distinguishes three outcomes, and the Done block listed seven reasons without ever answering the question the user has. The reasons are worth keeping — "too short" and "excluded repo" are choices someone may want to revisit — but they are the detail, not the answer.

- **The block now opens with the three buckets** it can honestly stand behind: `148 imported · 4 skipped · 2 failed`. *Imported* is loaded plus resumed, because a resume is an import that finished rather than a third outcome. *Skipped* is already-there, too-short and excluded-repo — choices, not failures. *Failed* is probe errors and upload errors alike, since both mean the session is not on the server. The per-reason grid is unchanged underneath.
- **A failure gets a sentence, not a bare red count.** Import exits 0 on partial failure, deliberately, because the rest landed — a lone red number under a "Done" rule says the opposite. The note names the remedy: re-running is safe, because classification asks the server what it already has, so a second run retries the failures and skips everything else.
- **Nothing is said about titles.** `--skip-title` exists (AI-2034), but AI-2047 — the server-side enqueue that would make titles actually arrive — is still Todo. A line promising them would be false the moment anyone read it.

Rolling the seven rows into a single number was the thing to avoid, so this doesn't: three buckets is what import actually knows, and the detail stays.

No new flag and no changed default, and the README doesn't reproduce this output, so it owes no README change.

`Cli.Tests.Unit` 3335 · 0 failed · AOT publish clean, zero IL warnings.

AI-2043


realtonyyoung (changes requested):
**[P2] Do not claim deliberately skipped sessions are in**

`failureNote` is emitted whenever `Failed > 0`, regardless of `Skipped`. But `Skipped` includes `TooShort` and `Excluded`, which explicitly were not sent to the server. A mixed run with one upload error and one excluded session therefore says “1 didn’t land. Everything else is in,” incorrectly claiming the excluded session landed. Please distinguish the successfully imported sessions from the deliberately skipped ones in this note, and cover the mixed skipped-plus-failed case.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):

PR #630, opened 2026-08-20
[AI-2131] Retire the machine pairing channel

George-Payne:
Deletes the CLI half of the machine pairing channel (#605). The server half never reached `main`, and this half is untagged — `v0.11.28` predates it — so no released artefact contains any of it.

A pairing that carries no token cannot authenticate a headless CLI, and the interactive path never needed one: it signs in moments later, so there is nothing to consent to. What remained was RFC 8628's device authorization grant with the payoff removed. WorkOS has shipped the real device grant since June 2025, which is what replaces this.

Nothing here changes behaviour for a user. The step already skipped on headless, `--no-prompt` and the `None` provider, `BrowserPairingFlow` treated 404/401/403/405 as "no channel here" and continued, and no deployed server serves the routes.

- **`Capacitor.Cli.Core/Auth/`** — 8 of the 9 files #605 added. **`SystemBrowser.cs` stays**: `LoopbackBrowser.cs:16` is its caller on the live OAuth path, so deleting it would be a CS0103 on a shipped flow.
- **`SetupCommand.cs`** — loses step 1b, the two identity checks around it (`AssertPairingIdentityAsync`, `CompletePairingAsync`) and the `SpectrePairingProgress` renderer. Step labels are untouched: the pairing was "1b", so 1..6 stand.
- **`Models.cs` / `HttpClientExtensions.cs`** — the three `JsonSerializable` registrations and the `PairingSecretHeader` constant.
- **`help-setup.txt` / `README.md`** — the "approving this machine" copy in both, and its quick-reference restatement.
- **`Capacitor.Cli.Core.Tests.Unit.csproj`** — drops `Microsoft.Extensions.TimeProvider.Testing`, which #605 added and only `BrowserPairingFlowTests` used. `TelemetryClientTests` declares its own local `FakeTimeProvider` and is unaffected.

1,490 deletions, no insertions, against #605's +1512/−10.

Verified before deleting: every type declared in the removed files (`IPairingChannel`, `MintOutcome`, `PollOutcome`, `PairingContinuity`, `MintPairing*`, `PairingStatusResponse`, `PairingVerdict`, `IPairingProgress`) is referenced only from within the delete set, `SetupCommand.cs`, or `Models.cs`. `JwtPayload` had exactly one caller, `PairingIdentity`.

`Capacitor.Cli.Core.Tests.Unit` is green (1932 passed). `Capacitor.Cli.Tests.Unit` has one failure, `LaunchdStartStopTests.WriteAndBootstrap_writes_the_unit_and_bootstraps_without_a_leading_bootout` — confirmed pre-existing by stashing this change and re-running on a clean tree. It is environmental: the devcontainer's umask is `0002`, so the test's temp dir is created group-writable and `ServiceFiles.RequireNotWorldWritable` refuses it by design.

Design: `docs/superpowers/specs/2026-08-19-ai2025-retire-pairing-design.md` §7, in `kcap-server`. The server half, the `machine_pairings` DROP and the feature gate follow there.


realtonyyoung:
NO FINDINGS

realtonyyoung (approved):

PR #635, opened 2026-08-20
Exclude where state is global, use identity where a pid is involved

### LINKED ISSUE #634: Daemon suite is documented as serial, but the flag that enforces it does not serialise anything
`CLAUDE.md` tells everyone to run `Capacitor.Cli.Daemon.Tests.Unit` with `--maximum-parallel-tests 1`, calling the suite "not parallel-safe". Both halves of that are wrong, and the guidance hides a real defect.

**The flag does not serialise a suite.** `GetMaxParallelism` feeds only the unconstrained bucket's `Parallel.ForEachAsync`; `ConstraintKeyScheduler` has no global-limit semaphore and its phase runs concurrently with the parallel phase. Measured on the daemon suite under that exact flag: peak simultaneous tests 4, not 1 — `CodexLauncherTests`' `HOME` window still overlapped 17 other tests, and one `CodexHostedAgentRuntimeFactoryTests` window overlapped 48. It narrows race windows; it does not close them.

**The suite is green at full parallelism.** Four consecutive runs, 2763 tests, 0 failed, 67–90s — against 3m13s under the documented invocation. So the flag costs every local run ~2.5× wall clock and buys no guarantee.

**The actual defect it hides.** 16 tests mutate a process-global environment variable (`HOME`, `CODEX_HOME`, `GEMINI_CLI_HOME`, `KCAP_DAEMON_SUPERVISED`) behind a *keyed* `[NotInParallel("HomeEnvVarMutation")]`. A keyed constraint excludes only tests carrying the same key, but the readers are not in the cohort: 18 sites in `src/` read `HOME` or the user profile, and 15 files in the suite spawn real `git`, which inherits it. Those tests were racing under the flag the whole time. A further 8 in `CodexConfigWriterTests` mutate `HOME` through a static helper, so a body-level scan does not even see them.

Two adjacent problems in the same area:

- A method-level `[NotInParallel(…)]` silently shadows a class-level one (the engine takes the first constraint, and the method's comes first). `Services/LocalPermissionBridgeTests.cs` has a dead class-level bare attribute because of this.
- Three tests poll `kill(pid, 0)` on a pid they do not own (a grandchild reparented to init). That is wrong in both directions: a recycled pid reads as "still alive", which is a false *pass* in the one positive assertion. The Windows `ConPtyJobObjectTests` has the same shape via `Process.GetProcessById`.

Inok:
Closes kurrent-io/kcap-cli#634 — kurrent-io/kcap-cli#634 (under [AI-1977](https://linear.app/kurrent/issue/AI-1977/kcap-cli-make-the-test-suite-state-independent-instead-of-serialising)).

`CLAUDE.md` told everyone to run `Capacitor.Cli.Daemon.Tests.Unit` with `--maximum-parallel-tests 1`, calling the suite "not parallel-safe". The flag does not serialise a suite — `GetMaxParallelism` feeds only the unconstrained bucket, and `ConstraintKeyScheduler` has no global-limit semaphore — so under that exact invocation peak concurrency was 4, `CodexLauncherTests`' `HOME` window still overlapped 17 other tests, and one `CodexHostedAgentRuntimeFactoryTests` window overlapped 48. The suite meanwhile runs green at full parallelism. So the guidance cost every local run \~2.5× wall clock, guaranteed nothing, and hid the defect it was blamed on.

## What this changes

**Exclusion where the state is genuinely process-global.** Five classes promoted from keyed to bare `[NotInParallel]`. A keyed constraint excludes only tests carrying the same key, which is sound only when every *reader* is in the cohort — and for `HOME`/`CODEX_HOME`/`GEMINI_CLI_HOME`/`KCAP_DAEMON_SUPERVISED` the readers are 18 sites in `src/` plus 15 files in the suite that spawn real `git` and inherit it. Three keys are deleted rather than promoted: two on reader-only tests, and one dead class-level bare in `LocalPermissionBridgeTests` that every method's own attribute had been shadowing.

`EnvScope` **enforces the constraint instead of asking for it.** The constructor requires some `NotInParallel`; `EnvScope.Exclusive` requires an unkeyed one. It resolves first-wins exactly as `GroupTestsByConstraintsCore` does, so a method-level key shadowing a class-level bare is caught rather than blessed, and it rejects a co-present `[ParallelGroup]` (that combination lands in the constrained-group scheduler and never gets `RequiresGlobalNotInParallelLock`). All 24 promoted tests plus the 6 already-bare ones go through it and lose their hand-rolled `try/finally`.

`PidIdentity` **replaces pid-number liveness with incarnation comparison** over the existing `ProcessStartToken`. `kill(pid, 0)` on a pid we no longer own is wrong in both directions: a recycled pid reads as our process still alive, which was a false *pass* in `Agent_survives_unrelated_pool_thread_churn_while_the_thread_lives`. `PtySpawnTests` loses its class-level bare — reaping a child you forked needs no exclusion, since the zombie holds the pid until you wait on it. The one `waitpid` on a shim-reaped pid becomes an identity assertion, which is strictly stronger than `ECHILD`: an unreaped zombie still carries its token, so "the shim reaped it" is now asserted directly rather than inferred from a wait that could equally have been stolen.

**Docs.** `CLAUDE.md`'s rule is rewritten (bare vs keyed vs shadowing, and what CI's flag actually buys); the whole-suite serial invocation is deleted. Separately, the 122-line "What this project does" section of accreted per-PR narrative moves verbatim to `docs/CHANGES.md`, leaving nine invariants behind — see the second commit.

## Verification

* 3× daemon suite at full parallelism: **0 failed** (2767/2767/2768 tests), 62s / 66s / 107s, against 3m13s under the documented invocation.
* Overlap analysis over a run's report JSON: all **44 bare test instances overlapped 0 peers**; sequential tail 1.78s.
* Full solution the way CI runs it: **9390 passed, 0 failed**.

## Also in here

`GitProviderRouterTests` fails at full parallelism for the same class of reason — the router memoizes into a production static and the per-test reset that clears it is itself a global mutation. Found while verifying this change; CI's serial flag is why it had not been seen. Fixed with a class-level bare `[NotInParallel]`. It is one test, not an audit of that suite.

## Reviewer notes

* **CI is unchanged.** Both flags stay in `ci.yml`; this makes `--maximum-parallel-tests 1` unnecessary for the daemon suite but does not audit the other three. Dropping it is a follow-up.
* **Two changes are not exercisable on macOS** and land verified by review plus the CI leg that runs them: `ConPtyJobObjectTests` (Windows) and `PtySpawnTests.Child_side_exec_failure_reports_failed_step_exec_and_reaps_cleanly` (`[RunOn(OS.Linux)]`). The latter is the most load-bearing pid rewrite here, so its Linux leg is the real gate.
* **Promoting to bare hides a race rather than removing it.** Passing `home` explicitly through `CodexLauncher.Prepare` is the actual fix; exclusion is the correct interim guard, not the destination. Follow-up in the design doc.
* A source-scan test that would fail the suite on a raw `Environment.SetEnvironmentVariable` was written and dropped as severable — that bypass stays open, and is listed under Follow-ups.

Design: `docs/superpowers/specs/2026-08-21-test-parallel-safety-exclusion-and-pid-identity-design.md`

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,6 +1,6 @@
 # Kurrent Capacitor CLI
 
-**File paths:** CLI source at `src/Capacitor.Cli/`, shared core at `src/Capacitor.Cli.Core/`, daemon at `src/Capacitor.Cli.Daemon/`, npm packages at `npm/`, Claude Code plugin at `kcap/`.
+**File paths:** CLI source at `src/Capacitor.Cli/`, shared core at `src/Capacitor.Cli.Core/`, daemon at `src/Capacitor.Cli.Daemon/`, desktop app at `src/Capacitor.App/`, npm packages at `npm/`, Claude Code plugin at `kcap/`.
 
 **Harness layout:** vendor-specific code lives under `Harness/`. Vendors: Antigravity, Claude, Codex, Copilot, Cursor, Gemini, Kiro, OpenCode, Pi.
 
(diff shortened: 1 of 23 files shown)


Inok:
Both Qodo findings were real; fixed in 86a82db2.

**1. Flaky guard test (High) — correct, and it was this PR's own defect in miniature.** `EnvScopeGuardTests` carries three different constraints deliberately, and shared one variable name across them. Keyed `[NotInParallel("EnvScopeGuardTests")]` does not exclude the unmarked test, so the unmarked test could read what the keyed one was holding — and would then fail on a value that actually proves `EnvScope` works. Each test now owns its variable.

The second half of the finding is also right and is already the repo's rule elsewhere (*"Never assert that an environment variable is absent"* — the host environment is free to define anything). Every assertion now compares against a snapshot taken at test entry rather than `null`, which is a better assertion anyway: for the two refusal tests the claim is "it threw before writing", and unchanged-vs-before says exactly that.

**2. Doubled apostrophe (Low) — fixed.** `'{key}''s` became `the parallel constraints for '{key}'`.

Daemon suite still green at full parallelism after both: 2770 tests, 0 failed.

alexeyzimarev (approved):

PR #638, opened 2026-08-21
[AI-2052] Headless WorkOS: the device authorization grant

George-Payne:
Gives WorkOS a headless sign-in path, so a CLI that cannot reach a loopback callback can still authenticate. This is the last thing keeping legacy GitHub App auth alive. Answers the auth half of #535, which stays open until AI-2163 covers non-interactive workspace creation.

- **The grant** — RFC 8628 against AuthKit's `user_management` endpoints, public client, proxy not involved. Confirmed live against our AuthKit application before any of it was written, which is also where `verification_uri_complete` came from.
- **A shared poll loop** — the RFC 8628 half of `RunDeviceFlowAsync` is extracted and parameterised, fixing three defects once for both providers: the `while (true)` that never honoured `expires_in`; the unchecked deserialize that made a 429 or a 5xx HTML page an NRE rather than a backoff; and an unguarded send that let one slow POST end a sign-in the user was still completing.
- **Selection is a ladder, not a heuristic** — loopback opens for everyone and only an explicit `--device` skips it. There is no environment input at all: `IsHeadless()` is true for every SSH session including Remote-SSH and `ssh -L`, where loopback works perfectly, so selecting on it would demote the largest group of users to typing a code. The device code stays reachable mid-wait by pressing `d`.
- **`SystemBrowser`** — reports whether it launched anything. No browser here means the user's browser is on another machine, so a `127.0.0.1` callback cannot reach them; both rungs answer that with a device code rather than a URL ending in a refused connection.
- **The org gate corrects rather than refuses** — the authorize leg takes no `organization_id`, so the human picks at the AuthKit screen and the CLI cannot constrain it. A mismatch switches the token and re-verifies the result before anything is stored.
- **`ChooseDiscoveryProvider`** — loses its headless branch and its `--device` arm; `--github` keeps its discovery meaning. Arming `LoginCommand.NewFacade`'s provisioner belongs here rather than in a follow-up: without it a zero-workspace headless user now completes a sign-in and holds a live credential with nowhere to spend it.
- **Output** — the white-label provider goes unnamed (GitHub keeps its name, since there the user is looking at a github.com page), a pre-filled code is checked rather than entered, and the escape hatch sits under the URL it is an alternative to.
- **Docs** — `help-setup.txt`, `help-login.txt`, `README.md`.

Four of these were found by driving the CLI by hand rather than by the suite, which stayed green throughout and grew from 1950 to 1967 tests: a crash after a successful auth, a wrapped command that could not be copied, an escape-hatch keystroke that answered the next prompt unattended, and the unguarded send above.


src/Capacitor.Cli.Core/Auth/OAuthLoginFlow.cs
```diff
@@ -570,31 +656,196 @@ internal static Parameters WorkOSFrontChannel(string? organizationId) {
...
+        progress.Notice("");
+        progress.Notice(
+            browserOpened
+                ? $"  1. Your browser should have opened {device.VerificationUri}"
+                : $"  1. Open {device.VerificationUri} in a browser");
+
+        // No clipboard copy, unlike the GitHub flow: the code has to be READ ALOUD OR RETYPED on
+        // another device for this flow to mean anything, and a silent copy invites pasting it into
+        // whatever page is already open.
+        progress.DeviceCode(device.UserCode, device.VerificationUri, provider: null, prefilled: browserOpened && !string.IsNullOrEmpty(device.VerificationUriComplete));
```
alexeyzimarev: [P2] `prefilled` is inferred from best-effort process launch, but the fallback link exposed to the user is the bare `verificationUri`. If the shell reports a successful launch but the user follows that printed link instead, the page is not prefilled while every progress sink says to check an already-entered code. Either expose the complete URI for that fallback or render instructions that remain correct for both routes.

test/Capacitor.Cli.Tests.Unit/Commands/TenantProvisionerHeadlessTests.cs
```diff
@@ -0,0 +1,42 @@
...
+/// </summary>
+public class TenantProvisionerHeadlessTests {
+    static WorkOSTokenSource Tokens() =>
+        new("access-token", refreshToken: null, (_, _) => Task.FromResult<WorkOSAuthResponse?>(null));
+
+    [Test]
+    public async Task Declines_instead_of_throwing_when_there_is_no_terminal_to_prompt_on() {
+        // Asserted, not assumed: this is the condition under test, and a test host that happened to
+        // own a TTY would otherwise pass it vacuously.
+        await Assert.That(AnsiConsole.Profile.Capabilities.Interactive).IsFalse();
```
alexeyzimarev: [P2] This test depends on ambient test-runner interactivity and deliberately fails under a TTY. The repository documents direct execution of test projects, so the same code can be green in CI and red in a developer terminal before behavior is exercised. Inject or override the interactivity predicate and assert the controlled non-interactive case instead.

src/Capacitor.Cli.Core/Auth/OAuthLoginFlow.cs
```diff
@@ -44,42 +45,44 @@ public static class OAuthLoginFlow {
...
 
+    /// <summary>
+    /// Only an explicit request skips the browser. There is deliberately no environment heuristic
+    /// here: HeadlessEnvironment.IsHeadless() is true for every SSH session, including Remote-SSH and
+    /// `ssh -L` where loopback works perfectly, so selecting on it would demote the largest developer
+    /// population from "the browser opens" to "read a URL, type a code". It briefly survived as a
+    /// wording variant that said "this looks like a remote session"; that went when the escape hatch
+    /// moved under the URL, where it is equally visible to everyone and the guess earns nothing.
+    /// </summary>
+    internal static WorkOSFlow ChooseWorkOSFlow(bool forceDevice)
```
alexeyzimarev, line 54: [P1] Bare non-interactive setup still selects loopback unless `--device` was explicit. Its only in-flight escape is a keyboard press, while `SystemBrowser.TryOpen` explicitly says success only proves that a handler launched, not that the remote user can reach this process at `127.0.0.1`. A headless host with an opener but no keyboard therefore waits for the callback until timeout, which leaves the core #535 scenario unresolved. Please provide an automatic device route for this case instead of relying only on launch failure or `d`.

src/Capacitor.Cli.Core/Auth/OAuthLoginFlow.cs
```diff
@@ -150,32 +149,109 @@ internal static bool ShouldDiscoverLogin(string? baseUrl, string[] args)
...
+            TResponse? parsed = null;
+
+            try {
+                parsed = await response.Content.ReadFromJsonAsync(typeInfo, attempt.Token);
+            } catch (Exception ex) when (ex is System.Text.Json.JsonException or HttpRequestException
+                                          || (ex is OperationCanceledException && !ct.IsCancellationRequested)) {
+                // fall through to the transient arm below
+            }
+
+            if (parsed is null) {
```
alexeyzimarev, line 238: [P2] This retries only when deserialization returns null. A transient non-2xx JSON response such as HTTP 500 `{}` or `{"message":"temporarily unavailable"}` deserializes into a non-null response object, then reaches the default arm with no recognized RFC error and aborts sign-in as `unknown error`. Treat non-success responses with an unrecognized or missing device error as transient too, and cover a parsed-JSON 5xx response.

src/Capacitor.Cli.Core/Auth/OAuthLoginFlow.cs
```diff
@@ -44,42 +45,44 @@ public static class OAuthLoginFlow {
     internal static GitHubFlow ChooseGitHubFlow(bool forceDevice, bool isHeadless, bool hasExchangeUrl)
         => forceDevice || isHeadless || !hasExchangeUrl ? GitHubFlow.Device : GitHubFlow.Browser;
 
+    /// <summary>
```
alexeyzimarev, line 48: [P3] `CLAUDE.md` requires comments to be short, explain only the enduring non-obvious constraint, and never narrate a change. This block records earlier wording and iterations; the same pattern recurs in the new polling, ladder, loopback, and prompt-hygiene comments. Please reduce those blocks to the durable reason and leave defect/change history in the commits or spec.

test/Capacitor.Cli.Core.Tests.Unit/Auth/WorkOSFlowLadderTests.cs
```diff
@@ -0,0 +1,207 @@
...
+using Capacitor.Cli.Core.Auth;
+using WireMock.RequestBuilders;
+using WireMock.ResponseBuilders;
+using WireMock.Server;
+
+namespace Capacitor.Cli.Core.Tests.Unit.Auth;
+
+/// <summary>The WorkOS sign-in ladder: loopback, escape hatch, automatic fallback.</summary>
+public class WorkOSFlowLadderTests {
+    const string Device = """{"device_code":"dc","user_code":"WXYZ-1234","verification_uri":"https://signin.example/device","interval":0,"expires_in":900}""";
```
alexeyzimarev, line 11: [P2] These tests feed a real HTTPS verification URI into `RunWorkOSDeviceFlowAsync`, which unconditionally calls the static `SystemBrowser.TryOpen`. On a developer machine the unit suite can therefore open real browser tabs. Inject the device-flow browser opener, as `LoopbackBrowser` already permits, and pass a fake here.

src/Capacitor.Cli/Commands/SpectreTenantProvisioner.cs
```diff
@@ -22,6 +22,22 @@ public async Task<ProvisionOffer> OfferCreateAsync(WorkOSTokenSource tokens, Can
         AnsiConsole.MarkupLine("  [yellow]Single sign-on found no Capacitor workspace for your account.[/]");
         AnsiConsole.MarkupLine("  [dim]A workspace that signs in with the GitHub App won't appear here.[/]");
 
+        // Every way out of this fork is a prompt, so with no terminal there is nothing to offer.
+        // Spectre throws NotSupportedException from inside the prompt rather than returning, which
+        // turned `kcap login --discover` on a piped stdin into a crash the moment we armed the
+        // provisioner for headless sessions. Deliberately fires no funnel event: nothing was offered,
+        // and recording a decline would attribute to the user a choice they were never shown.
+        if (!AnsiConsole.Profile.Capabilities.Interactive) {
```
alexeyzimarev: [P1] Issue #535 explicitly requires a non-Spectre counterpart and defines success as headless `kcap setup` being able to create a workspace. This branch instead returns `Declined`, so the zero-workspace non-interactive path can authenticate but still cannot complete setup. Please add the required non-interactive provisioning path, including organization and slug input, rather than ending at manual web guidance.

George-Payne: Split off into it's own ticket: [AI-2163](https://linear.app/kurrent/issue/AI-2163/non-interactive-workspace-creation-for-kcap-setup)
https://github.com/kurrent-io/kcap-cli/issues/535#issuecomment-5369106737

alexeyzimarev (approved):

PR #640, opened 2026-08-21
[AI-2156] kcap setup creates the flow and polls it to completion

George-Payne:
Gives the server's two rendezvous routes a caller. They shipped with none, so nothing generated a flow id and the browser's claim-on-arrival was what established ownership - which is where it sat under the retired pairing, and is the one property of the design the server half could not realise alone.

The leg runs after login, since both routes are authenticated. It generates a 128-bit base64url id, creates the flow, opens `{server}/setup?s=<id>`, and polls until every step it knows has settled.

- **`FirstRunFlowId`** - 16 CSPRNG bytes as base64url, 22 characters. The server's floor is what makes that the only shape that fits; it can check length and alphabet but never entropy, so the guarantee is the generator's alone.
- **`FirstRunFlowClient`** - the two routes, degrading rather than throwing. Refusals are handled apart: 404/401/403/405 on the create mean the tenant does not serve the flow and say nothing to the user; 429 reports the server's own `Retry-After` rather than sleeping through ten minutes of it; 409 retries with a fresh id, since it means the id is taken rather than the credentials wrong.
- **`FirstRunFlowPoll`** - the poll's decision, extracted so every branch is tested without a socket. 410 is a dead link, 404 a flow that will never be ours, 401 a re-login rather than a new link, and 5xx or a transport blip is another tick.
- **`FirstRunFlowOutcomes`** - outcomes, never instructions. Step and status strings map onto closed local sets and an unrecognised member is dropped, because `kcap setup` writes Claude Code hooks and a hook entry is a command string Claude Code runs. Which steps are gates stays the server's to say, through `can_finish`, rather than being restated here where an old CLI could get it wrong.
- **`BrowserFirstRunFlow`** - create, then open, then poll. The setup URL is composed locally, so unlike the pairing there is no server-supplied URL reaching a shell-executed open to validate.
- **`SetupCommand`** - an unnumbered leg after login. Skipped on `--no-prompt` and the `None` provider. Headless deliberately is **not** a skip: the link is printed as well as opened, which is what keeps the screens available to the device-path population rather than designing it out of them.
- **Any key ends the wait.** The 30-minute budget is the backstop for a terminal nobody is sitting at; a closed tab should not cost half an hour of dots.

Two things worth flagging for review. **401/403 on the create are read as "no flow here"** even though the route is authenticated - a gateway answering them on a path it does not know is indistinguishable from the feature being off, and a login succeeded seconds earlier, so guessing wrong here silently skips an additive leg while guessing the other way prints an alarming auth failure on every tenant that has the flow off. And **the leg reports, configuring nothing**: the screens that would push configuration are their own tickets, so the terminal steps remain what wires the machine up, and which of the two renders a given step is a decision that belongs to neither.

Unblocks two things the server half deferred: refusing a flow no CLI created (and with it metering the claim path, which is unlimited today), and `s` being single-consumer.

`Capacitor.Cli.Core.Tests.Unit` and `Capacitor.Cli.Tests.Unit` are green apart from `WriteAndBootstrap_writes_the_unit_and_bootstraps_without_a_leading_bootout`, which fails identically on an unmodified tree - it refuses a group-writable temp directory, which is a devcontainer artefact rather than anything here. AOT publish is clean of IL2026/IL3050.

AI-2156


src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -0,0 +1,164 @@
...
+        var interval = PollInterval;
+        var deadline = _clock.GetUtcNow() + PollBudget;
+        var first    = true;
+
+        FirstRunFlowResponse? last = null;
+
+        while (_clock.GetUtcNow() < deadline) {
+            // Polled before the first sleep: a flow the browser has already finished — a resumed link,
+            // or a tab that was quicker than this process — should not wait out an interval to be noticed.
+            if (!first) await Task.Delay(interval, _clock, ct);
```
alexeyzimarev: **[P2] Keep the key escape hatch responsive while waiting.** The key is checked only after the full delay, and a poll-time 429 grows that delay to 30 seconds; an in-flight HTTP call can then add the 15-second client timeout. A press can appear ignored for roughly 45 seconds even though the UI promises that any key ends the wait. Race the delay/request with a small key-watch loop or cancellation signal, and cover a key arriving while each await is in flight.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -0,0 +1,164 @@
...
+                _keys.Drain();
+
+                return new FirstRunFlowResult.Dismissed(last);
+            }
+
+            var poll = await channel.PollAsync(serverUrl, flowId, ct);
+
+            switch (FirstRunFlowPoll.Classify(poll.StatusCode, poll.Body is not null)) {
+                case FirstRunPollVerdict.State:
+                    last = poll.Body;
```
alexeyzimarev, line 151: **[P2] Reject a poll response for a different flow.** The create path checks `Body.FlowId`, but this state path accepts any readable body and can return `Finished` from it. A malformed response or misrouted cache/proxy answer with a missing or different `flow_id` can therefore mark this link finished based on another flow. Compare `poll.Body.FlowId` to `flowId` before storing/evaluating it, and add the symmetric poll-mismatch test.

README.md
```diff
@@ -98,6 +98,19 @@ The setup wizard walks you through:
...
+   ```
+   Opening your browser to finish setup.
+     If it didn't open:  https://acme.kcap.ai/setup?s=b7f3…
+
+     Press any key to carry on here instead.
+
+   Waiting…
+   ```
+
+   The link is printed as well as opened, so a machine with no browser of its own can be finished from a browser on another. Press any key to stop waiting. Setup carries on in the terminal whatever happens in the browser — the steps below run either way. Skipped on servers that do not offer it and under `--no-prompt`.
```
alexeyzimarev: **[P3] Document the no-auth skip.** The implementation returns early for `AuthProvider.None`, but this paragraph says the flow runs whenever the server offers it and lists only unavailable servers and `--no-prompt` as skips. Please mention the no-auth provider here and in `help-setup.txt` so the documented setup behavior matches the command.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -0,0 +1,164 @@
...
+                    break;
+
+                case FirstRunPollVerdict.Expired:
+                    return new FirstRunFlowResult.Expired();
+
+                case FirstRunPollVerdict.Gone:
+                    return new FirstRunFlowResult.Failed("The server no longer recognises this setup link.");
+
+                case FirstRunPollVerdict.Unauthenticated:
+                    return new FirstRunFlowResult.Failed("The server stopped accepting this sign-in mid-setup.");
```
alexeyzimarev: **[P3] Make the 401 outcome give the specified remedy.** The PR contract says a poll 401 means re-login, but this result only says that the sign-in stopped being accepted and `SetupCommand` immediately prints `Carrying on here.` No re-login is attempted and no copy tells the user how to recover. Return a distinct outcome so setup can re-authenticate, or at least tell the user to re-run/login rather than leaving the remedy implicit.

src/Capacitor.Cli.Core/FirstRun/FirstRunFlowPoll.cs
```diff
@@ -0,0 +1,57 @@
...
+        // Deliberately indistinguishable from an unknown id: a 403 would confirm the flow exists, and
+        // the id is the only thing an attacker would have had to guess.
+        404 => FirstRunPollVerdict.Gone,
+
+        410 => FirstRunPollVerdict.Expired,
+        429 => FirstRunPollVerdict.SlowDown,
+
+        // Terminal rather than retried. The bearer was minted before the loop started and nothing in
+        // it refreshes one, so every later tick answers the same 401 — seventeen minutes of polling
+        // followed by "expired", for a flow that was only ever a re-login away. The budget below is
```
alexeyzimarev: **[P3] Correct the stale duration/location in this rationale.** `PollBudget` is 30 minutes in `BrowserFirstRunFlow`, not seventeen minutes and not below this switch. The repo comment rule requires short, accurate non-obvious rationale; this can simply say that retrying the same bearer cannot recover a 401.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -876,6 +908,84 @@ internal static async Task<int> RunLoginStepAsync(
...
+        // Login ran immediately above, so no usable token here is a failure that already reported
+        // itself. Saying it twice would be the only thing this added.
+        if (tokens?.AccessToken is not { Length: > 0 } accessToken) return;
+
+        await Console.Out.WriteLineAsync();
+
+        FirstRunFlowResult result;
+
+        using (var http = new HttpClient { Timeout = BrowserFlowHttpTimeout }) {
+            http.DefaultRequestHeaders.Authorization = new("Bearer", accessToken);
```
realtonyyoung: **Finding 1 (most severe) — the 30-minute poll runs on a raw, never-refreshed bearer.**

Hand-rolled `HttpClient` + `TokenStore.LoadAsync` bypasses the `HttpClientExtensions.cs:59` choke point ("the ONE choke point every authenticated CLI request flows through"), so there is no `UnauthorizedRetryHandler`, no `IsExpired` check, and no `ServerIdentity.SameServer` binding check.

**Failure scenario:** WorkOS tenant. Login mints tokens whose `ExpiresAt` comes from the JWT's own `exp`, which for WorkOS user-management access tokens is minutes, not the hour the test fixtures' `expires_in: 3600` suggests. The user clicks through the agents + import screens for ten minutes. Every poll from expiry onward returns 401 -> `FirstRunPollVerdict.Unauthenticated` -> `Failed("The server stopped accepting this sign-in mid-setup.")`. The browser flow then *completes*, and the CLI never sees it.

Independently triggerable a second way: the machine's own daemon `TokenRefreshLoop` rotates the refresh token and publishes a new access token while this leg holds the old one in a `DefaultRequestHeaders` it never re-reads.

The comment at `FirstRunFlowPoll.cs:~29` — "The budget below is far shorter than any token's life, so meeting this at all means something else changed" — is the load-bearing assumption, and it is the one to challenge: 30 minutes is *longer* than a WorkOS access token's life, so this is the expected path, not the exceptional one.

This also makes the flagged 401/403 -> `Unavailable` trade on the create riskier than the write-up argues: because the token is read raw with no expiry check, the most likely cause of a 401 on `POST /api/first-run/flows` is a stale credential, not a gateway on an unknown path — and that gets silently swallowed as "this tenant doesn't offer the flow."

**Precedent with both guards present, same file:** `PingCliSetupAsync` (`SetupCommand.cs:1034-1048`) does the expiry *and* binding checks by hand, with the comment "would otherwise send the previous server's bearer to the new one". The new leg is the same shape with both guards dropped.

**Fix:** `CreateClientWithAuthStatusAsync(serverUrl, ct)` (or `CreateAuthenticatedClientAsync` with `autoRetryUnauthorized: true`), bail when status is not `Ok`/`NoAuthRequired`. That also restores `AttachObservationHeadersAsync` (`X-Kcap-Cli-Version` / `X-Kcap-Update-Check` are absent on both new routes today) and `ServerVersionCaptureHandler`. Keep the 15s per-request cap by setting `Timeout` on the returned client.

realtonyyoung (commented):
## Peer review via `kcap` code-review flow (reviewer vendor: `claude`)

Two rounds, reviewer signed off at round 2. The 8 findings below are those **still open at sign-off** — nothing fixed or superseded in round 1 is repeated here. Each is posted as an inline thread so it can be resolved individually.

**Verification scope — please read before acting.** The reviewer's worktree was on the merge-base (`a5d92c4d`), not the PR branch, so the inlined diff was authoritative for the new files. That means:

- **Verified** against the real checkout: every collaborator the leg leans on — `HttpClientExtensions`, `TokenStore`, `SystemBrowser`, `IKeyWatcher`, and `SetupCommand`'s existing login/ping legs. Findings 1, 2, 3, 5 and 8 are anchored in those, and I independently re-confirmed the citations behind 1, 2 and 3.
- **Not verified**: the six test files, the README hunk, and `help-setup.txt`'s rendered result. Test-coverage remarks are flagged as unverified rather than asserted.
- **Could not be settled**: the server contract for `/api/first-run/flows` (absent from both this repo and `kcap-server`'s main). Finding 4 is a contract risk, not a confirmed defect — it needs a check against a real payload.

**Ranking:** 1-3 are the ones worth acting on before merge. 4 is a should-confirm. 5-8 are cheap fixes; 6 and 8 are explicitly not blockers.

**Explicitly not findings**, for the record: the create-before-open ordering, composing the setup URL locally rather than taking it from the server, the closed-set outcome mapping, and treating the payload as outcomes-never-instructions all look right. `Base64Url.EncodeToString(RandomNumberGenerator.GetBytes(16))` is a correct 22-char/128-bit id. No AOT/trimming issue found: both new models are registered on `CapacitorJsonContext` (`Models.cs:962-963`), every serialize/deserialize goes through a `JsonTypeInfo`, and `Base64Url`/`Enum.ToString`/the LINQ `All` are all AOT-safe.

Of the author's two flagged-for-review decisions: the outcomes-never-instructions boundary holds up. The 401/403-on-create trade does not fully — see Finding 1 for why the premise weakens once the token is read raw.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -0,0 +1,164 @@
...
+            // Polled before the first sleep: a flow the browser has already finished — a resumed link,
+            // or a tab that was quicker than this process — should not wait out an interval to be noticed.
+            if (!first) await Task.Delay(interval, _clock, ct);
+
+            first = false;
+
+            // The way out of a wait whose browser is never coming back — a closed tab, a link opened on
+            // a machine that then went away. Drained rather than read, because the key is usually
+            // followed by a Return that the next prompt would otherwise take as its answer.
+            if (_keys.CanWatch && _keys.KeyAvailable) {
```
realtonyyoung: **Finding 2 — stale buffered input dismisses the wait before the first poll ever runs.**

Nothing drains the console buffer when the wait *starts*, and `first` suppresses only the `Task.Delay`, not this key check — so the check runs on iteration 1. Any byte left in stdin from the preceding steps counts as "the user pressed a key to carry on here."

**Failure scenario:** the user finishes browser sign-in, tabs back to the terminal, and taps Enter while Step 2 is printing "Logged in as ..." (an extremely common habit while waiting on an OAuth round-trip). `RunBrowserFlowStepAsync` then runs, `Opening` prints "Press any key to carry on here instead", the browser opens on `/setup?s=...`, and on the first loop iteration `KeyAvailable` is already true -> `Dismissed` -> "Left the browser to it." The whole leg is skipped in the same instant it advertises itself, with the browser sitting on a page nobody will now wait for.

**Precedent:** `OAuthLoginFlow.WatchForEscapeHatchAsync` (`OAuthLoginFlow.cs:817-838`) does not have this problem — it requires a *specific* key and ignores everything else, and re-reads `settled` before touching the keyboard with the comment "once the browser leg has finished, anything buffered belongs to the next prompt rather than to this one."

**Fix:** `_keys.Drain()` once after `progress.Opening(...)` and before the loop, so only keys pressed after the prompt is on screen count. (A deliberate key, as the login escape hatch uses, would be stronger still.)

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -876,6 +908,84 @@ internal static async Task<int> RunLoginStepAsync(
...
+    /// leaves setup running — sign-in has already happened, so nothing in this leg can strand a machine.</summary>
+    static async Task RunBrowserFlowStepAsync(string serverUrl, string provider, string profile, bool noPrompt) {
+        // --no-prompt is a scripted run and this waits on a human. None has no identity for a flow to
+        // be owned by, and its routes are authenticated. Headless is deliberately NOT a skip: a
+        // machine with no browser of its own is exactly the one whose user is sitting at another, and
+        // the URL is printed for them to carry across — the device path keeps the screens rather than
+        // designing that population out of them.
+        if (noPrompt || provider == AuthProvider.None) return;
+
+        var tokens = await TokenStore.LoadAsync(profile);
```
realtonyyoung: **Finding 3 — `TokenStore.LoadAsync` sits outside the try that the leg's "cannot crash setup" promise depends on.**

The method's doc says "Every outcome leaves setup running", and there is an explicit `catch (Exception ex) when (ex is not OperationCanceledException)` around the flow to honour it — but this token read happens before it.

`ReadTokenFileAsync` (`TokenStore.cs:84-92`) only catches `FileNotFoundException`, `DirectoryNotFoundException` and `JsonException`, and its own comment states: *"real IO/permission faults still propagate and must not be masked as unauthenticated."* So the throw is documented intent, not an accident — the promise is contradicted by the contract of the very first call the leg makes.

**Failure scenario (Windows):** the daemon's proactive refresh publishes a new token for this profile at that moment. `TokenStore.SaveAsync` writes its temp with `FileShare.None` and then `File.Move(overwrite: true)`, which `ReplaceWithRetryAsync`'s own comment documents as taking the target with exclusive sharing and colliding with peers. The concurrent `File.ReadAllTextAsync` here throws `IOException`/`UnauthorizedAccessException`, unhandled -> `kcap setup` aborts after login and before any hooks are installed.

**Fix:** move the load and the token guard inside the existing try — which is exactly what `PingCliSetupAsync` (`SetupCommand.cs:1033-1035`) already does with the same accessor in the same file.

src/Capacitor.Cli.Core/FirstRun/FirstRunFlowOutcomes.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// outcome.
+    ///
+    /// <para><b>Which steps are gates is the server's to say, not this file's.</b> That is what
+    /// <c>can_finish</c> carries, and taking it from there rather than restating the rule locally is
+    /// what stops an old CLI calling a flow finished whose sign-in failed. The settled test below is
+    /// deliberately the permissive one — skipped and failed both count — because nothing after the
+    /// gate blocks finishing, and a flow whose import failed is over, not stuck.</para>
+    /// </summary>
+    public static bool IsFinished(FirstRunFlowResponse view) =>
+        view.CanFinish && KnownSteps.All(step => IsSettled(view, step));
```
realtonyyoung, line 107: **Finding 4 — `IsFinished` requires the `Done` step to carry an outcome; wire-name mapping is exact-case.**

*(Flagged unverified: the server routes are in neither this checkout nor `/Users/tony/dev/kcap-server`'s main, so this is a contract risk the reviewer could not settle. Please confirm against a real payload.)*

`Done` reads like the terminal value of the `step` field ("The step the browser is on"), not a gate with its own outcome. If the server's `steps` dictionary only carries the three gates, `StatusOf(view, Done)` falls to `Pending` forever, `IsFinished` is never true, and a browser flow that completed in ten seconds still burns the full 30-minute budget before reporting `Abandoned` -> "The browser didn't finish setup." — the exact opposite of what happened.

Same silent-never-finishes failure from a second cause: `Step()`/`Outcome()` switch on exact PascalCase wire strings, and `StatusOf` looks up `step.ToString()` in a default (ordinal, case-sensitive) `Dictionary`. A server serializing its step/outcome enums camelCase (`"signIn"`, `"completed"`) or snake_case makes every outcome unrecognised -> `Pending` -> same 30-minute hang. The "unknown members are dropped" boundary is right, but it means a naming mismatch degrades to a half-hour wait rather than to anything diagnosable.

**Fix:** (a) confirm against the real payload that `Done` is always present in `steps` once `can_finish` is true, and (b) add at least one test that feeds a *recorded* server response through `IsFinished` rather than a hand-built `FirstRunFlowResponse` — a hand-built fixture agrees with the CLI's own casing by construction and cannot catch this.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -0,0 +1,164 @@
...
+                case FirstRunPollVerdict.Unauthenticated:
+                    return new FirstRunFlowResult.Failed("The server stopped accepting this sign-in mid-setup.");
+
+                case FirstRunPollVerdict.SlowDown:
+                    interval = interval + SlowDownStep < MaxInterval ? interval + SlowDownStep : MaxInterval;
+                    progress.PollTick();
+
+                    break;
+
+                default:
```
realtonyyoung, line 182: **Finding 5 — backoff only exists for 429; every other unhappy path polls flat-out for the full 30 minutes.**

Transport failure (status 0), 5xx, and a 200 whose body didn't parse all land on `Wait` (this `default:` branch), which leaves `interval` at 2s. Only `SlowDown` widens it.

**Failure scenario A:** the server rolling-restarts, or the laptop's wifi drops, five minutes into the wait. The CLI issues a request every ~2s for the remaining 25 minutes — ~750 failed requests — and ends on `Abandoned`. Compare `HttpClientExtensions.SendWithRetryAsync`, the repo's convention, which backs off exponentially.

**Failure scenario B:** the poll route rate-limits at 1/min and sends `Retry-After: 60`. The create path reads that header carefully (both delta and HTTP-date forms, `FirstRunFlowClient.cs:~65-72`), but `PollAsync` discards it — `FirstRunPollOutcome` carries no `RetryAfter` field at all — so the loop escalates +2s/tick to a 30s ceiling and keeps drawing 429s for the rest of the budget, never observing a single state.

**Fix:** apply `SlowDownStep` to `Wait` as well (or a separate transport-failure ramp), and carry `RetryAfter` on `FirstRunPollOutcome` so a server-specified wait wins over the local ramp.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -0,0 +1,164 @@
...
+                if (++attempt >= CreateAttempts)
+                    return new FirstRunFlowResult.Failed("Could not claim a setup link on this server.");
+
+                continue;
+            }
+
+            if (created.StatusCode == 0)
+                return new FirstRunFlowResult.Failed("Could not reach the server to start browser setup.");
+
+            if (created.StatusCode is < 200 or >= 300 || created.Body is null)
```
realtonyyoung: **Finding 6 (minor) — an unparseable body on a 2xx is reported as the server rejecting the request, quoting a success status.**

`created.Body is null` is OR'd into the non-2xx branch, so a 200/201 whose JSON this build couldn't read prints "The server did not accept a browser setup link (HTTP 200)."

`FirstRunFlowClient.ReadAsync` already distinguishes the two cases (it returns null only on a read/parse failure), and the poll side treats exactly this as a distinct condition — worth a separate message here too, e.g. "couldn't read the server's answer".

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
realtonyyoung, line 151: **Finding 7 — the poll never verifies the echoed `flow_id`.**

`RunAsync` deliberately checks `created.Body.FlowId == flowId` and fails on a mismatch (line ~87: *"a flow other than the one asked for is not an answer to the question... It cannot happen against the server this was written for, which is exactly why a disagreement is worth stopping on rather than polling an id this process never generated"*). That argument applies verbatim here: the `State` branch assigns `last = poll.Body` and runs `IsFinished(last!)` on whatever arrived, with no such check. The asymmetry is the finding — one route treats a wrong id as fatal, its twin trusts it silently.

The caching-proxy theory is explicitly **ruled out**: both requests carry `Authorization`, and RFC 9111 §3.5 forbids a shared cache from storing such a response absent an explicit `public`/`s-maxage`/`must-revalidate` directive. The realistic trigger is the one the create check already guards against — a server-side id-routing or tenant-scoping bug.

**Failure scenario:** a server-side routing defect returns flow B's state on `GET /api/first-run/flows/{A}`. B has `can_finish: true` with all four steps settled. The CLI prints "Browser setup finished." and stops polling while flow A — the one whose URL is on screen and which the user is still working through — is untouched.

Severity is low because the leg configures nothing (the outcomes-never-instructions boundary holds), so the damage is one wrong line of terminal output plus an abandoned real flow. But the fix is one line — the same `string.Equals(..., StringComparison.Ordinal)` guard, mismatch -> `Failed`/`Gone` — cheap enough that the asymmetry is harder to justify than the check.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -55,6 +56,32 @@ internal static string Indent(string message) =>
...
+
+        AnsiConsole.MarkupLine(SetupAuthProgress.Indent("Opening your browser to finish setup."));
+        AnsiConsole.MarkupLine(SetupAuthProgress.Indent($"[dim]If it didn't open:[/]  {Markup.Escape(setupUrl)}"));
+        AnsiConsole.WriteLine();
+        AnsiConsole.MarkupLine(SetupAuthProgress.Indent("[dim]Press any key to carry on here instead.[/]"));
+        AnsiConsole.WriteLine();
+        AnsiConsole.Markup(SetupAuthProgress.Indent("Waiting…"));
+    }
+
+    public void PollTick() => AnsiConsole.Write(".");
```
realtonyyoung: **Finding 8 (lowest severity) — unbounded `PollTick` dots.**

At the 2s interval over the 30-minute budget this is ~900 dots. Two concrete harms, neither aesthetic:

1. The dots bury the recovery affordance the leg's entire headless story rests on. `Opening` (just above) prints "If it didn't open: `<url>`" specifically so "a machine with no browser of its own can be finished from one on another" — then `PollTick` soft-wraps ~900 dots into 9-11 lines beneath it, scrolling that URL out of a standard 24-line terminal within the first few minutes. The user who most needs to copy the URL is the one least likely to still see it.
2. `PollTick()` fires identically on `State`, `SlowDown` and `Wait`, so the terminal renders "the server is healthy and you are mid-flow" and "the server has been unreachable for 25 minutes" as the same stream of dots. That is Finding 5 seen from the user's side: there is no signal anywhere — not in the interval, not on screen — distinguishing a working wait from a dead one.

**Fix:** a Spectre status line that rewrites in place, or a dot per Nth tick with the URL re-printed periodically. Not a blocker.

src/Capacitor.Cli.Core/FirstRun/FirstRunFlowOutcomes.cs
George-Payne: Verified against the server half (the paired kcap-server change): `FirstRunFlowView.From` always sends **all four** steps — `FirstRunSteps.InOrder` (SignIn, Agents, Import, Done) — each with an outcome, keyed by `enum.ToString()`, so the exact-case names (`Done`) match this mapping. `Done` does settle: the payoff screen writes the completed event, and the poll endpoint own comment describes a CLI poll landing "just after the browser settled Done" — the exact state this `IsFinished` exists to detect. The only forward-compat caveat is the deliberate, documented closed-set trade (an unknown new-server step is dropped rather than stalling the poll). No change needed.

alexeyzimarev (dismissed):

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -0,0 +1,229 @@
...
+        var deadline = _clock.GetUtcNow() + PollBudget;
+        var first    = true;
+
+        FirstRunFlowResponse? last = null;
+
+        while (_clock.GetUtcNow() < deadline) {
+            // Polled before the first sleep: a flow the browser has already finished — a resumed link,
+            // or a tab that was quicker than this process — should not wait out an interval to be noticed.
+            if (!first) {
+                if (await WaitForIntervalAsync(interval, last, ct)) return new FirstRunFlowResult.Dismissed(last);
```
realtonyyoung: [P2] Bound the wait by the remaining poll budget

`Retry-After` is now correctly honored beyond the local 30-second backoff cap, but this call sleeps the entire interval before the deadline is checked. If a proxy/server returns (for example) `Retry-After: 3600` on the first poll, `kcap setup` waits an hour before returning `Abandoned`, despite `PollBudget` being the documented 30-minute backstop; on a host without a watchable keyboard there is no earlier exit. Sleep for `min(interval, deadline - now)` and abandon at the deadline without issuing another poll. The current 60-second test does not exercise this interaction.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -876,6 +921,92 @@ internal static async Task<int> RunLoginStepAsync(
...
+            using (http) {
+                http.Timeout = BrowserFlowHttpTimeout;
+
+                // Ok runs the leg. NoAuthRequired is the None-provider skip again — silent, for the
+                // same reason. The rest get one line: the factory's quiet variant prints nothing, and
+                // expired / not authenticated / wrong server all share one remedy.
+                if (authStatus is not AuthStatus.Ok) {
+                    if (authStatus != AuthStatus.NoAuthRequired)
+                        await Console.Out.WriteLineAsync(
+                            "  [dim]Skipped browser setup: the stored token is not usable. Run 'kcap login' to re-authenticate.[/]");
```
realtonyyoung, line 959: [P3] Render this markup through Spectre

`Console.Out.WriteLineAsync` does not interpret Spectre markup, so this branch prints the literal `[dim]` and `[/]` tags to the user. Use `AnsiConsole.MarkupLine(...)` (as the surrounding setup output does), or remove the markup tags from this plain-text write.

realtonyyoung (changes requested):
Static review of the latest head found two actionable issues. The first allows a server-supplied delay to defeat the flow's stated 30-minute backstop; the second exposes raw Spectre tags in an error path. No build or tests were run, per request.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
Re-reviewed the latest head (13c6f4b). The previously reported findings are addressed. No build or tests were run, per request.

PR #653, opened 2026-08-24
Desktop shell: Home with repository and harness selection

alexeyzimarev:
Gives the Avalonia desktop app a Home screen: pick a repository, pick a harness, start a hosted session. First slice of the hosted-agents shell.

Linear: **AI-2194** (part of **AI-2171** — design record and decisions are in the AI-2171 comment).

> **No GitHub issue to close.** The slice issues were created directly in Linear rather than in GitHub Issues, which is backwards from this repo's convention (`CLAUDE.md`: open issues in GitHub, Linear auto-imports them). Creating one now would import as a *second* Linear issue alongside AI-2194, so this PR references Linear only. Flagging rather than papering over it.

## What this does

- **The daemon advertises what it can host.** `DaemonInfoDto` gains a trailing `SupportedVendors`, populated from the same set already sent on `DaemonConnect` — one source of truth, no second computation. `null` means *unknown* (an older daemon), never *none*.
- **The harness is remembered per repository.** `AppState.HarnessByRepo` maps repo path → vendor token. Switching repository switches the chip; each repo keeps its own. A repo with no stored choice falls back to the default, never to the previously selected vendor.
- **`HostedHarnessCatalog` builds the picker rows** from `Core.Setup.HarnessCatalog.All`, contributing only what Core has no notion of: transport family and daemon-advertised availability. A tenth vendor added to Core appears automatically.
- **Sessions start through the server.** The local `Spawn` frame resolves against the daemon's PTY launchers — `claude` and `codex` only — so a nine-vendor picker cannot use it. `ServerLaunchClient` invokes `RequestLaunchAgentV2`, which reaches every vendor via the runtime factories.
- **`HomeViewModel` + `HomeView`**, wired through the composition root and disposed on both shutdown paths. Home is the window's default tab.

## Worth a reviewer's attention

- **The hub payload binds by name under `SnakeCaseLower`.** An earlier revision of this branch used camelCase keys: it compiled, passed every test, and would have bound `daemon_name`/`repo_path` as null on every launch. `LaunchHubJson.Configure` is now shared by the client and its tests so the two cannot diverge, and a test pins the exact twelve-key set.
- **UI-thread marshalling.** `IDaemonClientService.Agents`/`.Snapshots` are pushed from a background thread; both projections `ObserveOn` before their binding operator. There is a test that genuinely fails if either is removed — it asserts the thread the bound collection is mutated on, because "does not throw" turned out not to be falsifiable here.

## Not in this slice

- **The repository list.** Only the folder picker landed, not the derived list of known repositories — so per-repo harness memory is currently reachable only by re-navigating a native dialog. Needs a decision: land it here or defer explicitly.
- **Theme.** Home is dark-only inside a theme-following shell (`RequestedThemeVariant="Default"`). On a light-mode host it renders as a dark slab in light chrome.
- Artifacts/outputs and any diff view, both deliberately out of scope.

## Tests

`Capacitor.App.Tests.Unit` 1169/1169 · `Capacitor.Cli.Core.Tests.Unit` 1996/1996 · AOT publish clean.

`Capacitor.Cli.Daemon.Tests.Unit` is 2759 pass / 1 fail / 38 skipped. The failure is `Installed_codex_schema_matches_the_vendored_pin` — installed codex 0.149.0 against an older vendored pin. Pre-existing on `main`; this branch touches no codex or schema files.

### DIFF
--- a/docs/superpowers/plans/2026-08-23-ai2194-desktop-shell-home.md
+++ b/docs/superpowers/plans/2026-08-23-ai2194-desktop-shell-home.md
@@ -1,0 +1,831 @@
+# Desktop Shell: Chrome, Home, and Repository + Harness Selection — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Turn the Avalonia app's window into a Home surface that lists hosted sessions and starts new ones, with the harness chosen per session and remembered per repository.
+
+**Architecture:** The app already observes the daemon over the local control socket (`IDaemonClientService` → `SourceCache<AgentStatusDto>`); Home renders off that cache. Starting a session goes the other way — through the *server*, because the local `Spawn` frame is PTY-only (claude/codex) while the server's `RequestLaunchAgentV2` reaches every vendor via the daemon's runtime factories. The harness catalogue comes from the daemon itself: `DaemonInfoDto` gains an additive `SupportedVendors` member carrying the same set the daemon already advertises to the server, so availability is the runtime-factory probe and never a hardcoded list.
+
+**Tech Stack:** .NET 10, Avalonia 11 + FluentTheme, ReactiveUI.Avalonia, DynamicData, `Microsoft.AspNetCore.SignalR.Client` (new to `Capacitor.App`, already in `Directory.Packages.props` at 10.0.11), TUnit.
+
+**Spec:** The design record is the AI-2171 Linear comment (decisions, with the reasoning); the visual reference is the design canvas linked from it. AI-2194 is this slice.
+
+## Global Constraints
+
+- **Nothing here is safety-bearing.** `AppState` is app-owned UX state only; the CLI's fixed-namespace marker remains the source of truth for anything else (`AppStateStore.cs:6`).
+- **AOT.** The app publishes NativeAOT. Every new serialized type joins a `JsonSerializerContext`; no reflection-based `JsonSerializer` overloads. Verify with `dotnet publish src/Capacitor.Cli/Capacitor.Cli.csproj -c Release 2>&1 | grep -E 'IL[23][01][0-9]{2}'` — must be empty.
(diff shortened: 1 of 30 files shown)


alexeyzimarev:
Note for reviewers: the Active session cards on Home are **display-only on purpose** — no click handler, no command. Opening a session needs the session workspace, which is AI-2195 (declared `blocked_by` this issue), so in this slice there is nowhere for a click to go.

Wiring the two actions `AgentActionService` already exposes (`OpenInWeb`, `RequestStop`) onto each card was considered as an interim affordance and declined: it commits the card's click gesture to a meaning AI-2195 wants to take over. Rationale and the follow-ups are recorded on AI-2195.

So an inert card is expected behaviour here, not a defect.

alexeyzimarev:
All five review findings verified against the code and fixed in `11930d79`.

| Finding | Disposition |
|---|---|
| Launch ignores cancellation | Fixed. `HomeViewModel` now takes the app's shutdown token and passes it to `ILaunchClient.StartAsync`. This was also the upstream cause of a shutdown race already noted during review. |
| Linear IDs in comments | Fixed — 12 occurrences removed; `scripts/check-linear-ids.sh` passes locally. Also removed references to internal plan artifacts ("Task 5", "task-6-brief") that a future reader cannot resolve. |
| `JsonValueKind` checked directly | Fixed via `JsonElementExtensions.IsNull`. I checked first whether a defensive accessor would weaken a wire-format assertion — `IsNull` is a direct kind check, so it does not, and `GetProperty` still distinguishes absent from explicit null. |
| Repo key casing mismatch | Fixed, with one correction to the suggestion: a blanket case-insensitive comparer would be wrong on Linux, where two paths differing only in case are genuinely different repositories. Now `OrdinalIgnoreCase` on Windows/macOS, `Ordinal` on Linux, applied on **read** because `System.Text.Json` rebuilds the dictionary with a default comparer on load. |
| Hub recreated while in use | Fixed. `StartAsync` now holds the connection gate across the invoke, so a second launch cannot dispose a connection the first is mid-invoke on. Launches are a per-click UI action, so serializing them costs nothing. |

The casing fix carries a test that asserts the platform's own answer rather than one hardcoded expectation, so it is meaningful on every CI leg. Verified it fails without the fix: reverting the comparer to `Ordinal` turns it red, restoring it turns it green.

App suite 1170/1170.

PR #656, opened 2026-08-24
[AI-2039] kcap daemon service ensure: the flow's daemon-install ladder

George-Payne:
The first-run flow's Done detour ("reach this machine from anywhere") needs one action that makes the daemon service-installed and running. The ladder existed only in the Avalonia wizard (`Capacitor.App`, being retired by AI-2053): `DaemonStepViewModel` classifies a fresh `service status --json`, `DaemonMutationLane` dispatches and classifies the mutation, and `ReasonRouting` maps `start_gate_reason=` tokens to a recovery surface. The CLI had all the primitives but nothing composing them into the ladder the flow drives. This adds that composition.

- **`kcap daemon service ensure`** — from a fresh status read, install when there is no unit or start when the unit is stopped, and report "already enabled" when the daemon is running. Ambiguous states (unknown probe, active transaction, orphan label, stale marker) fail closed to attention with a coded reason — never guessed, never mutated into.
- **Born `prompt`** — the install bakes `KCAP_CONSENT_SEED_DEFAULT=prompt` (plus the expected-server pin) exactly as the app's `MutationEnv` does, so an app-installed daemon is born `prompt`: nothing runs unattended on someone else's say-so.
- **Gated start** — on launchd, the start runs the verified transaction with the seed directive in its gate env, so a unit not baked `prompt` is refused with `verify_start_gate` (28) and one `start_gate_reason=` line, mapped machine-readably to `recovery_surface=takeover|reinstall|attention` via the pinned `ReasonRouting` table — never derived from prose.
- **Degraded end state off-macOS** — Windows/Linux get plain install/start (no gates, no rollback); `ensure --json` reports `"verified":false` so the flow's copy can say so.
- **`ReasonRouting`/`RecoverySurface` move to `Capacitor.Cli.Core`** — the CLI cannot reference the app, both reference Core; the pinned token→surface table now has one home (same rescue shape as AI-2167), with the app's references updated.
- **Machine-readable result** — `--json` emits a snake_case `ServiceEnsureJson` (service id, fresh state, action, outcome, recovery, reason, verified) through the shared `ServiceJsonContext`, including on the pre-flight refusals.

Windows answer to AI-2039's open question, established while wiring the install: the daemon itself is fully cross-platform (ConPTY, Scheduled Task service, win-x64 npm, Windows CI), hosted agents and the server→daemon SignalR path work there; what is Windows-gated is the local `kcap agent` drive and the launchd-only verified transaction. The detour therefore shows everywhere; the copy reflects plain install off-macOS.

Tests: 28 new/moved unit tests (classifier rows incl. the launchd stopped-but-installed shape, born-prompt bake, JSON render, failure→recovery mapping, dispatch rows); `ReasonRouting` tests moved with the type. Full CLI suite 3481, Core 2009, App 1157, all green; AOT publish clean.


src/Capacitor.Cli/Commands/DaemonServiceCommands.cs
```diff
@@ -287,6 +288,194 @@ async Task<int> StatusJson() {
...
+
+    /// <summary>Pure: the unit env for an ensure install — the ambient capture, overlaid with the
+    /// born-<c>prompt</c> directive and the expected-server pin, so both are deliberate unit content
+    /// regardless of what the installing shell exported. <c>prompt</c> WINS over an ambient value
+    /// (a refusal exported as <c>deny</c>/<c>allow</c> must not survive into the unit): the flow's
+    /// contract is "an app-installed daemon is born prompt", and the gate's identity half re-reads
+    /// the expect pin, so a unit without it can never pass a later gated start.</summary>
+    internal static Dictionary<string, string> EnsureUnitEnv(
+            string? profileName, string serverUrl, IReadOnlyDictionary<string, string> ambient) {
+        var env = new Dictionary<string, string>(ambient) {
```
realtonyyoung, line 461: ### Finding 1 — High
**An explicit `--profile P` can bake a unit whose effective server is still the ambient `KCAP_URL`, and the plain platforms can report that failed boot as success.**

> The ambient clone retains `KCAP_URL`, then overlays `KCAP_PROFILE`. Because URL resolution outranks profile resolution, `--profile P` does not actually pin P. This can bake expectation P alongside effective URL Q, causing a boot refusal; on the unverified Linux/Windows arms the supervisor invocation can be accepted and ensure can return `outcome:"installed"`/exit 0 although the daemon immediately refuses. If the URLs coincide, URL resolution still produces a profile-null result and token lookup can fall back to the on-disk active profile. A URL-only invocation also installs without a profile, but the later launchd start gate requires a non-empty invoking profile and can never pass.
>
> **Supporting precedence:** `src/Capacitor.Cli/Services/ServiceEnvironment.cs:15,130-140` and `src/Capacitor.Cli.Core/Config/ProfileResolver.cs:22-35`.
>
> Strip `KCAP_URL` when pinning P and fail closed (or define a coherent URL-only gate contract) when no profile name resolves.
>
> **Coverage:** None of `ServiceEnsureTests.cs:EnsureUnitEnvTests` supplies ambient `KCAP_URL`, combines it with a different explicit profile, or drives the resulting environment through profile resolution/start gating. `Does_not_pin_an_empty_profile` checks only key omission and therefore does not catch the unusable URL-only lifecycle.

*Verified independently:* `ServiceEnvironment.cs:15` literally lists `"KCAP_URL"` in the captured `Keys` array, and `UnusableUrlDiagnostic.cs:50` states that `KCAP_URL` and `--server-url` "both outrank the profile".

src/Capacitor.Cli/Commands/ServiceEnsure.cs
```diff
@@ -0,0 +1,135 @@
...
+        if (probe == LabelProbe.Unknown)
+            return new EnsureDecision(EnsureAction.Attention, "status_unknown");
+
+        // A held transaction is never mutated into — the wizard waits it out; a CLI verb reports
+        // it and lets the flow retry.
+        if (txnActive)
+            return new EnsureDecision(EnsureAction.Attention, "txn_active");
+
+        // A validated pid means the daemon is up; that is the flow's done state.
+        if (state == ServiceState.Running && daemonPid is not null)
```
realtonyyoung: ### Finding 2 — High
**Ambiguity checks lose precedence to the success arm.**

> A running daemon plus any stale transaction marker returns `AlreadyEnabled`/0 despite the unconditional documented fail-closed rule. A running loaded label whose unit file has disappeared likewise returns success and masks an installation that cannot survive a relaunch/login. The ambiguity checks must precede the success arm.
>
> **Coverage:** `Stale_marker_precedes_mutation_rows` exercises only `NotInstalled`; it does not cover marker + Running. The orphan test covers only `Installed` + no unit, not `Running` + no unit. The current tests therefore miss both precedence holes.

*Note:* `README.md:744` and `help-daemon.txt` both state the fail-closed rule **unqualified** ("a stale marker … fail\[s\] closed to attention"), so the shipped docs currently overstate the guarantee. Either the ladder or the docs needs to move.

src/Capacitor.Cli/Commands/DaemonServiceCommands.cs
```diff
@@ -287,6 +288,194 @@ async Task<int> StatusJson() {
...
+        // non-empty KCAP_PROFILE, so a bare `ensure` on launchd must still carry the resolved name.
+        var profileName = DaemonCommands.ExtractFlagValue(args, "--profile") ?? AppConfig.ResolvedProfile?.ProfileName;
+
+        var query     = manager.Query(id);
+        var daemonPid = DaemonPidProbe.ValidatedPid(store, id);
+        var txnActive = ServiceTxnLock.IsHeld(store, id);
+        var txnMarker = ServiceTxnMarker.Exists(store, id);
+
+        var decision = EnsureClassifier.Classify(
+            query.Probe, query.State, query.UnitPresent, daemonPid, txnMarker, txnActive);
```
realtonyyoung: ### Finding 3 — High
**“Already enabled” does not prove that the service job owns the validated daemon PID.**

> The success predicate establishes that some PID-file owner is a daemon and that some service job is running, but not that they are the same process. On launchd, `query.JobPid` is available and must equal `daemonPid`, matching `ServiceVerify.IsReadyAsync`'s ownership predicate. If launchd reports Running but the job PID is absent/unparseable, classify attention.
>
> **Do *not* require a non-null job PID on systemd or Windows:** those managers deliberately cannot supply it, and their documented `verified:false` already-enabled arm should retain the coarser Running + validated-daemon check. The classifier therefore needs platform/evidence-availability context rather than interpreting every null `JobPid` identically.
>
> The discarded evidence is defined at `src/Capacitor.Cli/Services/IServiceManager.cs:21` and populated at `src/Capacitor.Cli/Services/LaunchdServiceManager.cs:60-71`.
>
> **Coverage:** The classifier API has no JobPid/platform input, so `Running_with_validated_pid_is_already_enabled` cannot distinguish matching, mismatching, and unavailable ownership evidence. No existing row catches this.

This call site has `query` in hand and passes `Probe`/`State`/`UnitPresent` but never `JobPid`.

src/Capacitor.Cli/Commands/DaemonServiceCommands.cs
```diff
@@ -287,6 +288,194 @@ async Task<int> StatusJson() {
...
+        var config = await AppConfig.LoadProfileConfig();
+        return config.Profiles.TryGetValue(profileName, out var p) ? p.ServerUrl : null;
+    }
+
+    /// <summary>Wire token for the fresh-read state. An Unknown probe is reported as "unknown" rather
+    /// than falling through to a state value — the same never-masquerade rule status --json applies.</summary>
+    static string ServiceStateToken(ServiceQuery q) => q.Probe == LabelProbe.Unknown
+        ? "unknown"
+        : q.State switch {
+            ServiceState.NotInstalled => "not_installed",
```
realtonyyoung, line 372: ### Finding 4 — Medium
**Launchd’s stopped-but-installed arm emits a self-contradictory JSON state.**

> For a stopped retained plist, the ladder correctly chooses Start from `UnitPresent=true`, but the JSON reports `state:"not_installed"`. Unlike status JSON, ensure JSON omits `unit_present`, so consumers cannot recover the distinction; a success can contain `action:"start"`, `outcome:"started"`, and `state:"not_installed"`. Derive this state from unit presence or carry unit presence in the DTO.
>
> **Coverage:** `Launchd_stopped_but_installed_shape_is_start` asserts only classifier selection. The JSON tests construct DTOs directly and never exercise `ServiceStateToken`, so they do not catch the contradictory wire shape.

*Verified independently:* `LaunchdServiceManager.QueryCore` computes `var state = probe == LabelProbe.Loaded ? LaunchdUnit.StatusFromPrint(...) : ServiceState.NotInstalled;` while `unitPresent` is read independently from the plist — so the shape is real. The existing test proves the author knew about it and handled it in the classifier but not in the state token.

src/Capacitor.Cli/Services/ServiceVerify.cs
```diff
@@ -167,6 +167,13 @@ readPlist is null && plistExists is null
     /// fail-closed placeholder in dev/test builds.</summary>
     readonly Func<string, bool> _digestMatches = digestMatches ?? DaemonDigest.Matches;
 
+    /// <summary>The gate reason of the LAST Phase-A refusal, for an in-process caller (the ensure
+    /// ladder) that must map <see cref="VerifyExit.StartGate"/> to a recovery surface without
+    /// re-parsing its own stderr. Null after a passing start, a non-gate exit, or a fresh engine.
+    /// The app shells out and reads the token from child stderr instead; this is the in-process
+    /// twin of the same <c>start_gate_reason=</c> contract.</summary>
+    internal StartGateReason? LastGateReason { get; private set; }
```
realtonyyoung, line 176: ### Finding 5 — Low
**`LastGateReason` violates its own “null after a non-gate exit” contract on a reused engine.**

> The property still violates its explicit "null after a passing start \[or\] non-gate exit" contract: a second call on one engine leaves the prior refusal resident because entry and early returns never clear it. However, both ensure arms create a fresh engine, so no current production ensure path observes stale state. Final severity is **Low** as a latent reuse/API-contract defect. Reset on operation entry remains the correct hardening.
>
> **Coverage:** The supplied tests contain no repeated-call test on one `ServiceVerify` instance and no assertion of `LastGateReason` after success or a non-gate exit.

Not currently reachable — flagged because the doc comment on this very property promises semantics the code does not implement, which is the kind of thing a later caller reasonably trusts.

src/Capacitor.Cli/Commands/DaemonServiceCommands.cs
```diff
@@ -287,6 +288,194 @@ async Task<int> StatusJson() {
...
+    };
+
+    /// <summary>Shared non-success tail: the verify transaction already emitted its coded token and
+    /// <c>start_gate_reason=</c> line; the pure <see cref="EnsureFailureMap"/> derives the JSON's
+    /// recovery/reason fields, and the <c>recovery_surface=</c> line is re-emitted when a gate
+    /// refused. The exit code always passes through unchanged.</summary>
+    async Task<int> EnsureFailure(int exit, StartGateReason? gateReason, string state, string action, bool json) {
+        var (recovery, reason) = EnsureFailureMap.Map(exit, gateReason, verified: manager is LaunchdServiceManager);
+        if (recovery is not null) await Console.Error.WriteLineAsync($"recovery_surface={recovery}");
+        return await Report(new ServiceEnsureJson(id, state, action, "refused", recovery, reason), exit, json);
```
realtonyyoung: ### Finding 6 — Medium
**Refused launchd transactions always serialize `verified:false`.**

> The launchd-path boolean is computed and passed to `EnsureFailureMap`, then omitted from the DTO, so every refused verified transaction serializes `verified:false`. This contradicts the stated distinction that launchd arms use the verified transaction while off-macOS reports false. Pass the same path boolean as `Verified`; `outcome:"refused"` already communicates that verification did not succeed.
>
> **Coverage:** `EnsureFailureMapTests` do not render the DTO. `ServiceEnsureJsonTests.Recovery_and_reason_are_nullable` constructs a refused record without `Verified` but asserts only nullable recovery/reason behavior. No test drives `EnsureFailure` on a launchd manager and asserts `verified:true`.

The 7th positional parameter is simply omitted here, so it defaults to `false` — the same boolean is computed one line above and handed to `EnsureFailureMap.Map(..., verified: manager is LaunchdServiceManager)`.

src/Capacitor.Cli/Commands/ServiceEnsure.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        if (exit == VerifyExit.StartGateDrift)
+            // Drift is the gate's TOCTOU re-check refusing — a gate failure, so it surfaces the
+            // attention row, never auto-retried (same rule as the app's own table); the reason token
+            // keeps the JSON and the human line from reading empty.
+            return (RecoverySurfaceTokens.Token(RecoverySurface.Attention), VerifyExitToken(exit));
+
+        // Not a gate refusal. Verified runs carry their verify_* token; plain runs never wear the
+        // verify prefix (exit 1 is lock contention or a manager error, not a verify outcome).
+        return (null, verified ? VerifyExitToken(exit) : "plain_failure");
```
realtonyyoung, line 139: ### Finding 8 — Medium
**Known coded install/boot refusal reasons are discarded by ensure JSON outside exits 28/29.**

> A fresh launchd ensure-install with a package digest mismatch returns exit 21 and emits the specific `package_inconsistent` evidence, but JSON reports only `reason:"verify_viability"`, no `recovery_surface:"reinstall"`. Likewise, an attributable `consent_seed_unwritable` boot refusal returns the readiness-timeout exit and emits its coded reason, but JSON collapses it to `verify_readiness_timeout` instead of the existing `RecoverySurface.Storage` route; the server-expectation mismatch induced by finding 1 is similarly collapsed instead of routing to takeover. A flow consuming the advertised JSON contract cannot safely choose the recovery already encoded in `ReasonRouting`. **Carry the verified engine's structured viability/boot-refusal reason to `EnsureFailureMap`; do not parse stderr.**
>
> The lost producers are `src/Capacitor.Cli/Services/ServiceVerify.cs:402-411` (`AttributeReadinessTimeout` emits `refusal_reason=...`) and the gated install viability branch around `ServiceVerify.cs:700` (emits `viability_reason=package_inconsistent`).
>
> **Coverage:** `EnsureFailureMapTests` covers only `StartGate` and `StartGateDrift`. It has no row for Viability with a coded viability reason, or ReadinessTimeout with an attributed boot-refusal reason, and the current `Map` signature cannot express either input.

*Verified independently:* `ServiceVerify.cs:407` is `Say($"refusal_reason={evidence.Token}");` and `:710` is `Say($"viability_reason={GateReasonToken(StartGateReason.PackageInconsistent)}");` — both stderr-only, with no in-process property equivalent to the `LastGateReason` seam this PR adds for the gate path.

realtonyyoung (commented):
## Peer review — automated code-review flow (reviewer vendor: `codex`, model `gpt-5.6-sol`)

Ran a 3-round structured code-review flow against this PR. The reviewer signed off after round 3 with **seven actionable findings** (1 High×3, Medium×3, Low×1). One round-1 finding was **withdrawn** after context, and findings fixed or superseded mid-flow are not reposted — what follows is only what was still open at sign-off.

Each finding is an inline thread below, anchored to an added line, so it can be resolved individually.

| # | Severity | Summary |
|---|----------|---------|
| 1 | **High** | `EnsureUnitEnv` keeps ambient `KCAP_URL`, so `--profile P` does not actually pin P |
| 2 | **High** | Classifier success arm precedes the `orphan_label` / `stale_marker` fail-closed checks |
| 3 | **High** | "Already enabled" never proves the service job owns the validated daemon pid |
| 4 | Medium | `ServiceStateToken` ignores `UnitPresent` — a successful start reports `state:"not_installed"` |
| 5 | Low | `LastGateReason` is never cleared on entry, violating its own documented contract |
| 6 | Medium | `EnsureFailure` omits `Verified`, so refused launchd transactions always serialize `verified:false` |
| 8 | Medium | Coded `viability_reason` / `refusal_reason` evidence is discarded outside exits 28/29 |

**Withdrawn (no action needed):** finding 7, "the new verb is absent from the Getting Started surface." The reviewer withdrew it once shown `README.md:153` in full — that section points a human at `kcap daemon service install` and links to the daemon section, while `ensure` is explicitly flow-driven/machine-facing and *is* documented at `README.md:744` and in `help-daemon.txt`. Reviewer's words: *"Adding it to onboarding would blur the supported human workflow rather than repair missing documentation."*

**Findings 6 and 8 were deliberately kept separate** at the reviewer's call: *"Finding 6 drops the verified-path boolean from every refusal DTO; finding 8 drops specific cause/recovery evidence that the engine already determined. They have different fixes and different consumer impact."*

<sub>Reviewer caveats, carried through verbatim: the PR head `f922298` and the branch-only test/design files were not present in the reviewer's launch worktree, so it treated the submitted diff and test descriptions as authoritative and verified the referenced base implementations locally. **Tests were not run.** I independently confirmed every base-code claim against `main` before relaying — `ServiceEnvironment.cs:15` really does list `KCAP_URL` among captured keys; `LaunchdServiceManager.QueryCore` really does yield `Probe=Absent, State=NotInstalled, UnitPresent=true` for a stopped-but-installed service; `ServiceVerify.cs:407` and `:710` really are stderr-only `Say(...)` emissions with no in-process equivalent.</sub>

<sub>Posted as a `COMMENT` review, not an approval or a change request.</sub>

src/Capacitor.Cli/Commands/DaemonServiceCommands.cs
George-Payne: **Fixed in 4d368a5.** `EnsureUnitEnv` now strips ambient `KCAP_URL` whenever a profile is pinned (the pin is the sole URL authority — URL resolution outranks profile resolution, so baking both would point the daemon at the wrong server). And on launchd, `ensure` now fails closed with `no_profile_configured` before any mutation when no profile name resolves: a unit baked without one can never pass the start gate's identity half, so installing it would have reported a success that can never be gated-started. URL-only remains coherent off-macOS (no gate), covered by `Url_only_install_keeps_the_ambient_kcap_url` + `Pinning_a_profile_strips_an_ambient_kcap_url` and the dispatch row `Launchd_install_without_a_profile_fails_closed`.

src/Capacitor.Cli/Commands/ServiceEnsure.cs
George-Payne: **Fixed in 4d368a5.** `EnsureClassifier.Classify` now evaluates the repair rows — orphan label (`Installed` **or** `Running` with no unit file) and stale marker — *before* the success arm, so a stale marker or an orphaned running label can no longer ride a validated pid into `AlreadyEnabled`. Added rows `Stale_marker_precedes_a_validated_running_daemon` and `Running_label_without_a_unit_file_is_orphan`; the docs (README/help) already stated the unqualified rule, so the ladder moved to match them.

src/Capacitor.Cli/Commands/DaemonServiceCommands.cs
George-Payne: **Fixed in 4d368a5.** The classifier now takes `jobPid` + `jobPidAvailable`; on launchd, `AlreadyEnabled` additionally requires `jobPid == daemonPid` (the same ownership predicate `ServiceVerify.IsReadyAsync` uses), and a Running label with an absent/unparseable/mismatched job pid classifies `running_unconfirmed` attention. systemd/Windows pass `jobPidAvailable:false` and keep their documented coarser `verified:false` arm. Rows: `Launchd_running_with_job_pid_matching_the_validated_daemon_is_already_enabled`, `Launchd_running_with_a_different_job_pid_is_unconfirmed`, `Launchd_running_without_a_job_pid_is_unconfirmed`, `Managers_without_a_job_pid_keep_the_coarser_already_enabled_check`.

src/Capacitor.Cli/Commands/DaemonServiceCommands.cs
George-Payne: **Fixed in 4d368a5.** `ServiceStateToken` now derives from unit presence: `NotInstalled` + `UnitPresent` reads `installed`, so a stopped-but-installed launchd unit no longer serializes `state:"not_installed"` beside `action:"start"`. `ServiceStateToken` is now internal and directly tested (`ServiceEnsureStateTokenTests`), including the launchd stopped-but-installed shape.

src/Capacitor.Cli/Services/ServiceVerify.cs
George-Payne: **Fixed in 4d368a5.** `LastGateReason` (and the two new evidence properties) are reset at the entry of `StartVerifiedAsync`/`InstallVerifiedAsync`, so a reused engine never carries a prior refusal into a later call — the documented "null after a passing start, a non-gate exit, or a fresh engine" contract now holds. Covered by `Last_gate_reason_is_cleared_at_the_next_operation_entry` (gate refusal, then a lock-contended second call observes null).

src/Capacitor.Cli/Commands/DaemonServiceCommands.cs
George-Payne: **Fixed in 4d368a5.** `EnsureFailure` now passes `Verified: manager is LaunchdServiceManager` into the DTO, so a refused launchd transaction serializes `verified:true` (the verified transaction did run), while plain off-macOS refusals stay `verified:false`. Wire covered by `Refused_launchd_transaction_serializes_verified_true`.

src/Capacitor.Cli/Commands/ServiceEnsure.cs
George-Payne: **Fixed in 4d368a5.** The engine now surfaces both coded reasons in-process — `LastViabilityReason` (set at the gated-install `package_inconsistent` abort) and `LastBootRefusalToken` (set by the attributed readiness-timeout tail, threaded through `AttributeReadinessTimeout`) — and `EnsureFailureMap` routes them: viability `package_inconsistent` → `recovery_surface=reinstall`; attributed boot refusals (e.g. `consent_seed_unwritable` → storage, `server_expectation_mismatch` → takeover) via `ReasonRouting.ForBootRefusal`, instead of collapsing to the generic `verify_viability`/`verify_readiness_timeout`. Engine producers asserted in `Gated_install_with_bad_binary_digest_aborts_viability_with_reason_line` (now also checks `LastViabilityReason`) and `Gated_install_readiness_timeout_with_matching_marker_attributes_refusal_reason` (now also checks `LastBootRefusalToken`); the map rows are in `EnsureFailureMapTests`.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
Re-reviewed the current head after the prior findings were addressed. No actionable findings. Static review only; no build or tests run.

PR #657, opened 2026-08-24
[AI-2167] Let the flow fix a broken kcap PATH

George-Payne:
### What and why

The Agents screen's PATH warning (AI-2032) is the one justified coral in the flow: if the login
shell cannot find `kcap`, hooks run and record nothing, silently. The design offers **"fix it for
me"** and **"show me the line"**; the second already ships. The first needs two things that only
existed in `Capacitor.App` — the Avalonia supervisor this project retires — so this PR rescues
them into `Capacitor.Cli.Core` and adds the named capability the flow drives.

### 1. The rescue: `LoginShellProbe` + `PathShimInstaller` → Core

Both classes (plus the process seam they depend on) move from `Capacitor.App` to
`Capacitor.Cli.Core`:

- `IProcessRunner` + `ProcessResult`/`RunOptions`/enums + the production `ProcessRunner`
  implementation → `Capacitor.Cli.Core` (one copy shared by app and CLI).
- `ILoginShellProbe` + `LoginShellProbe` → `Capacitor.Cli.Core.Setup` (beside
  `AgentDetection`/`HarnessInventory`, whose inventory the flow already reads).
- `PathShimInstaller` + `ShimPreflight`/`ShimOutcome`/`ShimResult` → same namespace.

The app keeps working via `using`; its tests move with the classes. The installer's
destination-override seam and `Preflight` become public because Core has no
`InternalsVisibleTo` for the app (the app's shim coordinator drives preflight against temp
destinations in tests). The rescue shape is established for this retirement: types leave the
app before AI-2053 deletes it, and the app picks them up via `using`.

### 2. `kcap daemon shim ensure` — the flow's PATH-fix capability

The server→CLI lane carries **values, not paths or commands** (retirement spec §6.1), so the fix
cannot be a server-supplied path or shell line. It is a named CLI capability the CLI composes
itself:

- resolves its own binary path (`Environment.ProcessPath`) — never a server-supplied path;
- probes the interactive login shell (`$SHELL -lic`, fallback `-lc`, sentinel-parsed);
- on a **positive** absence, links `/usr/local/bin/kcap` to the running CLI (macOS only;
  osascript admin prompt; non-forcing `ln -s`; then **re-probes** so success is never reported on
  the symlink alone — the spec §5 rule);
- every ambiguous row **fails closed** with a coded reason, never guessed:
  - unknown probe → `probe_unknown`
  - a different filesystem entry already at the destination → `conflict` (never overwritten)
  - non-macOS → `unsupported_platform` (no osascript-based shim exists; copy reflects the plain
    "show me the line" shape)
- `--json` emits the outcome + coded reason the flow keys off (never prose), with the actionable
  `detail`/`sudo_fallback` on `installed_not_on_path`/`failed`. Exit 0 only when the terminal now
  resolves `kcap`.

Design doc: `docs/superpowers/specs/2026-08-24-ai2167-let-the-flow-fix-a-broken-kcap-path-design.md`

### Out of scope

- Server/flow wiring (AI-2048, AI-2156) — the verb exists; who calls it and when rides those.
- The probe-result field in the create-flow body — that is AI-2156's file; this PR only moves the
  probe to Core so that field can be reported there later.
- Everything else in `Capacitor.App` (AI-2053).

### Test plan

- Core suite: 2084 pass (incl. the four moved classes: `LoginShellProbeTests`,
  `PathShimInstallerTests`, `ProcessRunnerTests`, `StreamingRunnerTests`).
- CLI suite: 3457 pass / 19 skipped (incl. 23 `DaemonShimCommandsTests`: classifier rows,
  JSON contract, human output, dispatch).
- App suite: 1101 pass (incl. all classes touching the moved types).
- Daemon suite: 2771 pass under CI's parallelism cap (the environment was under contention from a
  concurrent agent compile; a serial-capped run is fully green and my diff touches no daemon
  files).
- AOT publish clean (no IL3050/IL2026).
- Manual smoke: `kcap daemon shim ensure --json` on Linux → single-line
  `{"outcome":"refused","reason":"unsupported_platform"}`, exit 1.


src/Capacitor.Cli/Commands/DaemonShimCommands.cs
```diff
@@ -0,0 +1,198 @@
...
+        // before the first probe or prompt — this verb's own ladder principle is fail-closed, and
+        // an unknown flag must not be silently ignored into a mutation. (-h/--help never reach
+        // here: Program.cs intercepts them for the whole command group.)
+        if (args.Any(a => a != "--json"))
+            return Usage();
+
+        var json = args.Contains("--json");
+        isMacOs ??= OperatingSystem.IsMacOS();
+
+        var target = (resolveTarget ?? (() => Environment.ProcessPath))();
```
realtonyyoung: [P1] Preserve the npm launcher when installing the shim

For the documented global npm installation, the process running this code is the platform NativeAOT binary that `npm/kcap/bin/kcap.js` spawned, so `Environment.ProcessPath` resolves to that binary rather than the npm launcher. Linking `/usr/local/bin/kcap` here therefore bypasses the launcher for every subsequent command. That breaks `kcap update`: `kcap.js` is the component that intercepts `update` and runs npm, while the native `UpdateCommand` explicitly handles direct-binary invocation by only printing `Run kcap update` and returning. After this repair, `/usr/local/bin/kcap update` can never actually update. Please resolve a CLI-controlled launcher target that preserves the installation's wrapper semantics (or make the native update path perform the update) instead of always linking the native process image.

src/Capacitor.Cli/Commands/DaemonShimCommands.cs
```diff
@@ -0,0 +1,198 @@
...
+                await Report(new ShimEnsureJson(Capability, target, true, true, "install", "installed"), 0, json),
+
+            ShimOutcome.InstalledButNotOnPath =>
+                await Report(new ShimEnsureJson(Capability, target, true, false, "install", "installed_not_on_path",
+                    Detail: result.Detail), 1, json),
+
+            ShimOutcome.Cancelled =>
+                await Report(new ShimEnsureJson(Capability, target, true, false, "install", "cancelled"), 1, json),
+
+            _ =>
```
realtonyyoung, line 171: [P2] Preserve the coded conflict outcome after the outer preflight

The preflight at line 116 cannot make the later install atomic. `PathShimInstaller.InstallAsync` repeats the preflight and returns `ShimOutcome.Failed` if an entry appears in between; likewise, the non-forcing `ln -s` can lose a race after both checks and return a generic failure. Both paths land here as `outcome:failed`, `reason:null` (and the latter may offer a sudo command that is guaranteed to hit the same existing entry), even though this command's machine contract promises `outcome:refused`, `reason:conflict` for that state. Carry a structured conflict result out of the installer, and classify an `ln -s` failure as conflict when a fresh preflight now sees one, so the flow receives the coded recovery row regardless of when the entry appears.

realtonyyoung (changes requested):
Static review found two correctness issues in the flow-facing shim contract. Per request, I did not build or run tests.

George-Payne:
Both review findings are addressed in b505405:

**P1 — npm launcher.** The shim now links to the npm launcher when this CLI is part of an npm-global install. `ResolveLinkTarget` derives the launcher path (`node_modules/@kurrent/kcap/bin/kcap.js`) from the running native binary's own location — the wrapper package is a sibling of the platform package — so no environment lookup is involved, and it falls back to the running binary for standalone installs. `/usr/local/bin/kcap update` therefore keeps the launcher-driven npm upgrade instead of landing on the native binary's "Run kcap update" hint.

**P2 — coded conflict row.** After a failed install the verb re-runs the destination preflight: an entry that appeared mid-flight (between the outer preflight and the installer's own checks, or a non-forcing `ln -s` that lost the race) now surfaces `outcome:refused, reason:conflict` with the "left untouched" detail — the coded row the wire contract promises — instead of a generic `failed`. A failed install with no foreign entry at the destination still reports `failed`.

Both are covered by new tests (launcher layout / standalone fallback / no-process-path; failed→conflict race / failed stays failed). Full CLI suite green, AOT publish clean.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
Re-reviewed the current head. The two prior findings are addressed.

PR #663, opened 2026-08-25
[AI-2216] Report the machine on create, and act on the Agents decision

George-Payne:
AI-2216

## What & why

Both halves of the first-run Agents lane existed and neither was connected to the other: the create sent only `flow_id` and `machine`, and the poll's `agents` was never read. A real `kcap setup` therefore showed "we could not read the coding agents on this machine", and turning four harnesses on installed nothing. The CLI now reports per-vendor detection when it creates the flow, and applies the returned decision in Step 4 instead of asking again.

## Where to look

`SetupDecisions.WithBrowserAnswer` is the single place an answer becomes install options; `HarnessCatalogConformanceTests` pins every catalogue vendor to an arm of it, so a harness added later cannot round-trip through the screen and then install nothing.

`Options.ToolsIndependentOfCapture` keeps the per-axis split to the browser. `--skip-<vendor>-hooks` stays a whole-vendor opt-out: scripts pass it expecting no writes, and `--no-prompt` gives them no moment to learn otherwise.

Both detection signals travel separately rather than OR'd — the screen names the signal it saw, and the vendors disagree about which one they have.

## Verification

`Capacitor.Cli.Core.Tests.Unit` 2245/2245. `Capacitor.Cli.Tests.Unit` 3547 passed, 19 skipped, 1 failed — a macOS launchd test that refuses a group-writable temp dir and cannot pass in this dev container. `Capacitor.App.Tests.Unit` 1088/1088. AOT publish clean for CLI and daemon.

Removing any vendor's arm from `WithBrowserAnswer` fails the conformance pin naming that vendor. Neutralising the skills audience fails `Shared_skills_are_not_written_for_a_vendor_the_browser_declined` while the terminal-path case still passes.


src/Capacitor.Cli/Commands/SetupDecisions.cs
```diff
@@ -41,6 +44,55 @@ public static bool DecideInstallAgents(CodingAgentsStep.DetectedAgents d, bool n
...
+            SkipCursorMcp      = Tools("cursor", options.SkipCursorMcp),
+            SkipCopilotMcp     = Tools("copilot", options.SkipCopilotMcp),
+            SkipGeminiMcp      = Tools("gemini", options.SkipGeminiMcp),
+            SkipKiroMcp        = Tools("kiro", options.SkipKiroMcp),
+            SkipPiMcp          = Tools("pi", options.SkipPiMcp),
+            SkipOpenCodeMcp    = Tools("opencode", options.SkipOpenCodeMcp),
+            SkipAntigravityMcp = Tools("antigravity", options.SkipAntigravityMcp),
+
+            // The browser answered every prompt this step would raise, so it must not raise one.
+            NoPrompt = true,
```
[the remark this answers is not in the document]
George-Payne: Disclosed on screen: 
image

src/Capacitor.Cli/Commands/SetupDecisions.cs
```diff
@@ -41,6 +44,55 @@ public static bool DecideInstallAgents(CodingAgentsStep.DetectedAgents d, bool n
...
+            SkipCodex       = Skip("codex", options.SkipCodex),
+            SkipCursor      = Skip("cursor", options.SkipCursor),
+            SkipCopilot     = Skip("copilot", options.SkipCopilot),
+            SkipGemini      = Skip("gemini", options.SkipGemini),
+            SkipKiro        = Skip("kiro", options.SkipKiro),
+            SkipPi          = Skip("pi", options.SkipPi),
+            SkipOpenCode    = Skip("opencode", options.SkipOpenCode),
+            SkipAntigravity = Skip("antigravity", options.SkipAntigravity),
+
+            SkipCursorMcp      = Tools("cursor", options.SkipCursorMcp),
```
realtonyyoung: [P1] Preserve whole-vendor skip flags on the browser path

If setup is invoked with `--skip-cursor-hooks` (and likewise Copilot, Gemini, Pi, OpenCode, or Antigravity), the corresponding `options.SkipCursor` is already true but `SkipCursorMcp` is false unless the separate MCP flag was also passed. This fold enables `ToolsIndependentOfCapture` and leaves the MCP skip false when the browser selected tools, so the downstream handler reinterprets the explicit CLI opt-out as “capture off, tools on” and writes that vendor’s MCP config. That contradicts the whole-vendor/no-write guarantee documented below. Please carry each pre-existing whole-vendor hook skip into its MCP skip before applying the browser’s independent axes, and cover the case where only `skipCursor: true` is supplied.

realtonyyoung (changes requested):
Requesting changes for the flag-precedence issue noted inline: on the browser path, a whole-vendor `--skip-<vendor>-hooks` opt-out can still mutate that vendor’s MCP configuration.

Static review only, as requested; I did not build or run tests.

src/Capacitor.Cli/Commands/SetupDecisions.cs
```diff
@@ -41,6 +44,60 @@ public static bool DecideInstallAgents(CodingAgentsStep.DetectedAgents d, bool n
...
+            SkipGemini      = Skip("gemini", options.SkipGemini),
+            SkipKiro        = Skip("kiro", options.SkipKiro),
+            SkipPi          = Skip("pi", options.SkipPi),
+            SkipOpenCode    = Skip("opencode", options.SkipOpenCode),
+            SkipAntigravity = Skip("antigravity", options.SkipAntigravity),
+
+            SkipCursorMcp      = Tools("cursor", options.SkipCursor, options.SkipCursorMcp),
+            SkipCopilotMcp     = Tools("copilot", options.SkipCopilot, options.SkipCopilotMcp),
+            SkipGeminiMcp      = Tools("gemini", options.SkipGemini, options.SkipGeminiMcp),
+            SkipKiroMcp        = Tools("kiro", options.SkipKiro, options.SkipKiroMcp),
```
realtonyyoung: [P1] Preserve Kiro’s intentionally independent MCP flag

Kiro is the exception to the whole-vendor rule applied here. Its existing handler and regression test (`Kiro_mcp_registered_when_hooks_skipped_by_flag`) explicitly define `--skip-kiro-hooks` as skipping only the invasive agent clone/default-agent change; MCP remains enabled unless `--skip-kiro-mcp` is also passed. Folding `options.SkipKiro` into `SkipKiroMcp` changes that contract only when a browser answer exists and silently drops Kiro tools the browser selected. Please keep this arm based on `options.SkipKiroMcp` alone and add the browser-path Kiro case to the regression coverage.

realtonyyoung (changes requested):
The original flag-precedence issue is fixed for the affected vendors, but the follow-up also couples Kiro’s intentionally independent hook and MCP flags. I’ve documented that regression inline and am requesting changes on the updated head.

Static re-review only, as requested; I did not build or run tests.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
Static re-review complete. The requested fixes address the prior findings. Per request, I did not build or run tests.

PR #665, opened 2026-08-25
[AI-2228] De-flake the attach detach/dial race test

George-Payne:
AI-2228

## What & why

`Detach_racing_the_dial_prevents_the_daemon_from_ever_seeing_an_attach` deadlocks against its own fixture on one of the two interleavings it exercises, and times out at its 5s backstop.

When the detach wins, `_stream` is still null, `DetachAsync` claims `Detached`, and the run settles at once — the case the test is for. When the dial wins, `DetachAsync` takes its success path: it writes a graceful Detach and deliberately leaves the transport open so a terminal frame can still beat `Detached`, leaving `RunCoreAsync` parked in `FrameCodec.ReadAsync` awaiting the peer. The fixture accepted only after `await runTask`, so the client waited for a server that was waiting for the client.

`AgentAttachClient` is not at fault and is unchanged.

## Where to look

The accept moves to run concurrently with the dial, and the connection is closed once an Attach has been observed — the EOF a daemon produces once it holds the Detach frame. `AcceptAndPumpInboundAsync` only reads, so accepting alone would not settle the client.

Assertion semantics are untouched: attempts that delivered an Attach are still skipped, and `sawNoAttach` must still be set by a detach-wins attempt or the test fails as vacuous.

## Verification

Reproduced on the 6th full-assembly run of `Capacitor.Cli.Core.Tests.Unit` (2245 tests): `TimeoutException` at `AgentAttachClientTests.cs:387`. Six further runs after the change, all clean.

Six clean runs against a one-in-six flake is not decisive on its own — the argument is that the deadlock is structural for the dial-wins interleaving, and this removes it. The class in isolation never reproduces it; contention across the assembly is required.

### DIFF
--- a/test/Capacitor.Cli.Core.Tests.Unit/LocalIpc/AgentAttachClientTests.cs
+++ b/test/Capacitor.Cli.Core.Tests.Unit/LocalIpc/AgentAttachClientTests.cs
@@ -384,18 +384,29 @@ public class AgentAttachClientTests {
             var runTask = Task.Run(() => client.RunAsync(80, 24, CancellationToken.None));
             await client.DetachAsync();
 
-            var outcome = await runTask.WaitAsync(TimeSpan.FromSeconds(5));   // must settle promptly, never hang
-
-            // Best-effort: when the dial itself got aborted before completing, nothing ever
-            // reached the listener's backlog, so an unbounded Accept() would hang forever waiting
-            // for a connection that will never arrive -- bound it and treat a timeout as "no
-            // attach possible", which is exactly the outcome under test.
+            // Accepted CONCURRENTLY with the run, and never after it. When the DIAL wins this race
+            // the client has written a graceful Detach and is parked in its read loop waiting for the
+            // peer to answer -- which is correct: DetachAsync deliberately leaves the transport open
+            // so a terminal frame can still beat Detached. Accepting only once the run had finished
+            // therefore waited for a client that was waiting for us, and the attempt deadlocked
+            // against its own fixture until the timeout below fired. Rare on a fast machine, where
+            // the detach almost always wins; reliable enough under CI contention to redden the leg.
+            //
+            // Best-effort still: when the dial WAS aborted before completing, nothing ever reached
+            // the listener's backlog, so bound the accept and read a timeout as "no attach possible"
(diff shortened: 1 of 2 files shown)


test/Capacitor.Cli.Core.Tests.Unit/LocalIpc/AgentAttachClientTests.cs
```diff
@@ -384,19 +384,30 @@ public async Task Detach_racing_the_dial_prevents_the_daemon_from_ever_seeing_an
...
+            //
+            // Best-effort still: when the dial WAS aborted before completing, nothing ever reached
+            // the listener's backlog, so bound the accept and read a timeout as "no attach possible"
+            // -- which is the outcome under test.
             using var acceptCts = new CancellationTokenSource(TimeSpan.FromMilliseconds(300));
             try {
                 await server.AcceptAndPumpInboundAsync(acceptCts.Token);
                 await server.WaitForReceivedAsync(FrameType.Attach, acceptCts.Token);
             } catch (OperationCanceledException) { }
             var receivedAttach = server.SnapshotReceived().Any(f => f.Type == FrameType.Attach);
```
realtonyyoung: [P2] Quiesce the inbound pump before deciding no Attach arrived

`AcceptAndPumpInboundAsync` starts its reader as a fire-and-forget task using the same 300 ms token as `WaitForReceivedAsync`. When that token expires, this catch can resume and snapshot `Received` before the pump has observed cancellation or drained an Attach already written to the socket; `CloseConnection` does not join the pump either. Under the scheduler contention this test is intended to survive, the Attach can therefore be recorded just after `receivedAttach` is computed, causing the attempt to set `sawNoAttach` and pass falsely. Expose and await pump completion, or otherwise establish EOF/cancellation quiescence, before taking the snapshot so the negative assertion is authoritative.

realtonyyoung (changes requested):
Static review found one remaining false-pass race in the test fixture. Per request, I did not build or run tests.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
Re-reviewed the current head. The inbound-pump race is addressed.

PR #667, opened 2026-08-25
Answer the create-a-workspace prompts with --org and --slug

George-Payne:
AI-2163

## What & why

`kcap setup` can authenticate anywhere since the device grant landed, but creating a workspace still needs a terminal: the zero-workspace fork asks three questions through Spectre prompts, and Spectre throws from inside a prompt rather than returning. `--org "<name>" --slug <slug>` supply those answers up front, so the fork raises nothing and the command is scriptable end to end.

Both or neither — the slug becomes a permanent public hostname, so nothing derives one from the name in a run with nobody watching. Every combination that would take the flags and then not act on them is refused: alongside `--server-url` or a positional tenant, alongside `--github`, a value that is blank or is the next flag, the `--org=` spelling, a slug that could never be a hostname, and a non-interactive session without `--no-prompt` (which would create the workspace, then throw at the first step the flags do not answer).

## Where to look

`WrongWorkspaceError` compares the landed workspace by **slug**, not by URL. Only the zero-workspace fork consults the provisioner, so an account that already has one never reaches it, and the run would otherwise configure a workspace nobody named — silently, with hooks and imports behind it. Profile names are the comparison because the server names the workspace it creates and the profile is stamped from the url it returns, which need not be the origin the CLI would guess.

Two contract changes ride along. `ITenantPicker` returning null now means the picker has already said why, so discovery adds no second line of its own; without that, a headless run gets per-workspace guidance followed by a contradicting "No tenant selected." And a headless multi-workspace discovery declines with guidance instead of throwing `NotSupportedException`, which reaches `kcap login --discover` too.

## Verification

Suites run in the devcontainer: `SetupCommandTests` 69, `SetupFacadeParityTests` 16, `TenantProvisionerHeadlessTests` 16, `SetupDecisionsTests` 22, `LoginFacadeParityTests` 10, `TenantPickerHeadlessTests` 2; Core auth 110, `SetupFunnelTests` 7; App `WizardAuthBridgesTests` 25, `SignInStepViewModelTests` 28.

`dotnet publish -c Release` emits no IL2026/IL3050.

Each guard was checked by breaking it and watching the named test fail: the slug comparison, the flag-value parse, the taken-slug exit, and both `RunDiscoveryAsync` wiring arguments — deleting either left the feature dead with the suite green until these tests existed.


src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -1134,6 +1240,12 @@ internal static IReadOnlyList<string> BrowserAgentsSummary(FirstRunAgentsAnswer
                     return null;
                 }
 
+                if (WrongWorkspaceError(requested, cfg.ActiveProfile, active.ServerUrl) is { } landedElsewhere) {
```
realtonyyoung: [P2] Validate the requested workspace before the auth commit

This guard runs only after `DiscoverAsync` returns `Committed`. At that point the WorkOS publication path has already merged the discovered profiles, changed `ActiveProfile`, stamped auth, and saved tokens. An account that already belongs to another workspace therefore exits with the new safety error but still has its local active profile and credentials changed. Please reject the mismatched workspace through the pre-commit boundary, or otherwise before any config/token publication, so the promised stop is side-effect free.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -1096,11 +1195,18 @@ internal static IReadOnlyList<string> BrowserAgentsSummary(FirstRunAgentsAnswer
...
+
         var provisioner = chosen == AuthProvider.WorkOS
-            ? new SpectreTenantProvisioner(new TenantProvisioningClient(new HttpClient()), ProvisioningEndpoint.Url)
+            ? new SpectreTenantProvisioner(
+                new TenantProvisioningClient(new HttpClient()), ProvisioningEndpoint.Url,
+                isInteractive: () => canPrompt, requested: requested)
             : null;
 
-        var result = await NewFacade(provisioner).DiscoverAsync(chosen, forceDevice, CancellationToken.None);
+        var result = await NewFacade(provisioner, new SpectreTenantPicker(() => canPrompt))
```
realtonyyoung: [P2] Keep `--no-prompt` out of the tenant picker

The picker is gated only by terminal capability. On a TTY, `kcap setup --org ... --slug ... --no-prompt` with multiple existing workspaces still opens the Spectre selection prompt, defeating unattended execution; selecting one also reaches the later commit-before-validation problem. Include `--no-prompt` in this prompt allowance, or bypass selection entirely when a requested workspace cannot be created because discovery returned existing tenants.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -137,6 +137,24 @@ public static async Task<int> HandleAsync(string[] args) {
         var skipClaude       = skipClaudeFlag || legacyPluginScope == "skip";
         var legacyProjectScope = legacyPluginScope == "project";
 
+        var (requestedWorkspace, workspaceArgError) = ParseRequestedWorkspace(args, serverUrlArg is not null);
```
realtonyyoung: [P2] Detect a present but valueless server flag

Conflict detection is based on the parsed value rather than flag presence. `--server-url` with no following value is therefore treated as absent, so pairing it with valid `--org`/`--slug` can continue into irreversible workspace creation instead of rejecting the conflicting modes. The unsupported `--server-url=...` spelling is likewise invisible here. Track presence separately and reject malformed values before considering workspace creation.

src/Capacitor.Cli/Commands/SpectreTenantProvisioner.cs
```diff
@@ -125,33 +219,43 @@ public async Task<ProvisionOffer> OfferCreateAsync(WorkOSTokenSource tokens, Can
...
-        var retry = $"Re-run [cyan]kcap setup {Markup.Escape(slug)}[/]";
-        return await AnsiConsole.Status().StartAsync($"Provisioning {slug}.kcap.ai — this can take a few minutes…", async ctx => {
+        // Two different re-runs. A workspace still being built will exist, so naming it is right; one
+        // that failed does not, and `kcap setup <slug>` treats a positional as an existing server and
+        // never reaches creation at all. The scripted forms name --no-prompt, without which the first
+        // prompt after discovery throws on the session the guidance was printed to.
+        var waitPlain    = Scripted ? $"Re-run kcap setup {slug} --no-prompt" : $"Re-run kcap setup {slug}";
+        var waitMarkup   = $"Re-run [cyan]kcap setup {Markup.Escape(slug)}[/]";
+        var createPlain  = requested is { } r
+            ? $"Re-run kcap setup --org \"{r.OrgName}\" --slug {slug} --no-prompt"
```
realtonyyoung: [P2] Quote retry arguments without changing their values

`OrgName` is inserted between literal double quotes without shell escaping. Names containing quotes, dollar expansions, backticks, backslashes, or newlines either fail to round-trip or execute substitutions when the suggested command is copied. The forbidden branch also calls `ToLowerInvariant()` on this entire command, changing the organization name. Build the command with a platform-appropriate argument-quoting helper and lowercase only the surrounding prose.

realtonyyoung (changes requested):
Four P2 findings: the requested-workspace safety check occurs after auth/config publication, `--no-prompt` can still reach tenant selection, a present but valueless server flag can be ignored before creation, and retry guidance does not safely preserve the organization argument. Details are inline.

PR #669, opened 2026-08-25
Gate Antigravity nudges on the invocation counter

### LINKED ISSUE #668: Antigravity re-injects the work-items nudge on every turn
`kcap hook --antigravity PreInvocation` fires once per turn, and the memory index is fenced against that by a once-per-conversation lease. The work-items nudge is not: `WorkItemsNudgeEmitter.Resolve` is a pure function of the session id, composed at the output layer after the lease is decided, so every turn writes it again as another `injectSteps` `userMessage` — a step Antigravity persists for the conversation.

The result is the nudge accumulating in the transcript, once per invocation, for the life of the conversation.

Reproduced against 0.11.33 with a synthetic payload (same conversation id, `invocationNum` 1/2/3):

```
invocationNum=1 -> 929 bytes of injectSteps
invocationNum=2 -> 929 bytes
invocationNum=3 -> 929 bytes
```

Counting the nudge in real `transcript_full.jsonl` files under `~/.gemini/antigravity-cli/brain/`:

| conversation | injections |
| --- | --- |
| 443d31c1-fc46-46ad-9574-82278d38eec3 | 228 |
| 8e387648-8b8c-4e16-8c10-98336e2dff3d | 199 |
| 595d36fd-8523-462e-ba6c-97037989bae2 | 31 |

The memory index appears exactly once in the same transcripts, which is what the lease buys.

Antigravity is not the only harness with a repeating start callback — Kiro, OpenCode and Pi also declare `CallbackMayRepeat: true` and resolve the nudge unconditionally — so the same shape is worth checking there. The harness-setup nudge shares the call site but carries its own 6-hour throttle, which bounds it.

Coldaine:
Closes #668 — Linear id omitted: the issue was filed minutes ago and has not been imported yet.

## What & why

Antigravity's `PreInvocation` is a per-turn callback (`CallbackMayRepeat: true`). The memory index is fenced against that by the once-per-conversation lease; the work-items and harness nudges are not — they are resolved at the output layer, after the lease is decided, as pure functions of the session id. Every turn therefore wrote another `injectSteps` `userMessage`, which Antigravity persists for the conversation. This gates both on `invocationNum`, the one payload field that separates the callbacks. An absent or non-numeric counter reads as first, so a payload change degrades to emitting rather than to silence.

## Where to look

The gate lives at the hook's output site, not inside the emitters, so the lease lanes keep their current acquire/complete/retry behaviour untouched.

Kiro, OpenCode and Pi also declare `CallbackMayRepeat: true` and resolve the nudge unconditionally. They are left alone here — each would need its own counter or marker — and #668 records that.

## Verification

Same conversation id, three invocations, before and after (stdout bytes):

| invocationNum | 0.11.33 | this branch |
| --- | --- | --- |
| 1 | 929 | 929 |
| 2 | 929 | 0 |
| 3 | 929 | 0 |

`AntigravitySessionStartMemoryTests`: 22/22 pass. Release publish is clean of IL2026/IL3050.

### DIFF
--- a/src/Capacitor.Cli/Commands/Harness/AntigravityHookCommand.cs
+++ b/src/Capacitor.Cli/Commands/Harness/AntigravityHookCommand.cs
@@ -207,9 +207,14 @@ sealed class AntigravityHookCommand(ConfigRoot config, ProfileContext profiles, 
         // subtracts HookBudget.Safety — do NOT subtract it again. Written even when the watcher-spawn
         // gate below returns early — a withheld watcher must not suppress injection.
         var fragment = await SessionStartMemoryHookSupport.AwaitBounded(memoryTask, budget);
-        var workItemsNudge = HarnessNudgeEmitter.Combine(
-            WorkItemsNudgeEmitter.Resolve(HarnessId.Antigravity, sessionId, activeProfile?.DisableWorkItemsNudge is true, home),
-            HarnessNudgeEmitter.ResolveFragmentForHook(activeProfile?.DisableHarnessNudge is true, config, home));
+        // Nudges are unleased pure functions of the session id, so on this repeating callback they
+        // must be gated by the vendor's own per-conversation counter: without the gate every turn
+        // re-injects them as another persistent userMessage step.
+        var workItemsNudge = IsFirstInvocation(payload)
+            ? HarnessNudgeEmitter.Combine(
+                WorkItemsNudgeEmitter.Resolve(HarnessId.Antigravity, sessionId, activeProfile?.DisableWorkItemsNudge is true, home),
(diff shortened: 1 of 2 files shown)


realtonyyoung:
Hi @Coldaine , thanks for the PR! May you please sign the CLA and we're happy to accept the pull: https://cla-assistant.io/kurrent-io/kcap-cli?pullRequest=669

alexeyzimarev:
@Coldaine thanks! Do you ming accepting the CLA?

Coldaine:
Hi @realtonyyoung @alexeyzimarev,

My apologies for the delay! I've signed the CLA and cleanly rebased the branch on top of the latest main. All unit tests pass and this should be ready for merge whenever you have a moment. Thanks!

PR #674, opened 2026-08-26
Stamp private under --private, rather than saying nothing

George-Payne:
AI-2222

## What & why

`kcap import --private` uploaded org-visible sessions and pulled them back only at the end of the run. An omitted `default_visibility` is not "no default": the server's generated column coalesces an absent one to `org_public`. Six of the nine sources omitted it under `--private` and left privacy to the closing PUT pass — minutes of exposure on a large import, and permanent for any session whose PUT fails, since those failures go to stderr and are swallowed by design. `ImportContext.VisibilityStampFor` is now the only code that decides what a session-start carries, and `HandleImport` resolves the same rule for the chain path, which has no `ImportContext`.

## Where to look

The two halves are asymmetric on purpose: the Step 3 default is a *creation* default and lands on `New` alone, while `private` is a floor, and re-asserting a floor on a replay can only narrow what is already there. Every site is therefore a widening and none a narrowing. `SetVisibilityNoneForAll` stays, and one path still needs it: a resumed chain session posts no session-start, so no stamp reaches it, and it joined `importedSessionIds` only on `OnSessionEnded` — so a resume whose session-end POST failed was privatised by nothing. `chainScopeSessionIds` captures chain sessions up front under `forcePrivate`, as the routed sources already were.

## Verification

`ImportVisibilityTests` 55 pass (48 before, plus a `forcePrivate` × `AlreadyLoaded` row and the chain-resume case). Full unit suite 3641 pass, integration 232 pass.

Five mutations, each killed: reverting `ForcePrivate ? "private"` fails 20 tests; dropping the `New`-only guard on the Step 3 default fails 12; reverting `chainDefaultVisibility` fails 1; dropping Antigravity's repair-branch stamp fails 1; dropping `chainScopeSessionIds` from the privatise union fails 1.


src/Capacitor.Cli/Commands/IImportSource.cs
```diff
@@ -41,17 +41,28 @@ internal sealed record ClassifyContext(
...
+    /// <summary>
+    /// The <c>default_visibility</c> to stamp on a session-start, or null to leave the field off.
+    /// <b>An omitted stamp is not "no default"</b> — the server coalesces an absent one to
+    /// <c>org_public</c> — so force-private must say <c>private</c> out loud.
+    ///
+    /// <para>Asymmetric on purpose: the Step 3 default is a <i>creation</i> default, so it lands on
+    /// New alone, while <c>private</c> is a floor that a replay can only be narrowed by.</para>
+    /// </summary>
+    public string? VisibilityStampFor(ImportCommand.ClassificationStatus status) =>
+        ForcePrivate                                        ? "private"
```
realtonyyoung, line 65: [P1] Do not treat replay stamps as a privacy floor

`default_visibility` does not narrow an existing historical-import session: the server's import-overlap branch deliberately leaves that column unchanged, and this interface itself documents below that a replayed session-start is only a create-time hint. Consequently, a routed `Partial` import can append content while the prior session remains org-visible; if the import then fails, most vendors are absent from both `importedSessionIds` and `privateScopeSessionIds`, so the closing PUT never reaches them. Apply `visibility=none` to every existing routed session before replaying content instead of relying on this stamp.

realtonyyoung (changes requested):
One privacy-blocking finding.

src/Capacitor.Cli/Commands/ImportCommand.cs
```diff
@@ -1342,14 +1349,29 @@ ImportOutcome                         outcome
...
+        // discarded — which is why scopedSessionIds above, and not this, is what privacy rests on
+        // for anything this run only revisits.
+        var chainDefaultVisibility = forcePrivate ? "private" : defaultVisibility;
+
+        // Status carries both filters by this point: the scope filter ran before classification, and
+        // an excluded source had its status flipped to Excluded — so these three statuses are exactly
+        // the sessions the user selected AND the server already has. Neither a chain resume (which
+        // posts no session-start) nor a routed replay outside the child-content scope reaches the
+        // privatize set any other way.
+        if (forcePrivate) {
```
realtonyyoung: [P1] Privatize existing sessions before replaying content

This captures existing sessions only for the closing PUT. Because the server discards `default_visibility` on an import-overlap start, a routed `Partial` replay (or an `AlreadyLoaded` replay that attaches child content) still uploads into the previously org-visible session and leaves that content visible until the entire import finishes. The updated spec now calls this out of scope, but it is the exposure window this PR's problem statement says it fixes. Apply `visibility=none` before replaying content for an existing session; keep the closing pass as recovery/defense in depth.

realtonyyoung (changes requested):
The update fixes permanent exposure after failed/revisited imports, but the privacy window remains.

src/Capacitor.Cli/Commands/ImportCommand.cs
```diff
@@ -1342,14 +1349,52 @@ ImportOutcome                         outcome
...
+            var existing = classifications
+                .Where(c => c.Status is ClassificationStatus.Partial
+                                     or ClassificationStatus.AlreadyLoaded)
+                .Select(c => c.SessionId)
+                .Distinct(StringComparer.Ordinal)
+                .ToList();
+
+            if (existing.Count > 0) {
+                display.BeginPhase("Making existing sessions private");
+                await SetVisibilityNoneForAll(httpClient, baseUrl, existing);
```
realtonyyoung: [P1] Do not upload when the privacy floor failed

`SetVisibilityNoneForAll` deliberately logs and swallows each non-success/exception, so this await does not establish that every `Partial`/`AlreadyLoaded` session is private. If a pre-import PUT fails, the code immediately proceeds into both import phases and publishes new content into that still-public session until the closing retry (or permanently if it also fails), preserving the privacy window on the exact failure path the change needs to close. Have this preflight return the failed IDs and exclude/abort those imports, or use a fail-closed variant here; the best-effort behavior remains appropriate only for the closing compatibility pass.

realtonyyoung (changes requested):
Not yet: the pre-import privacy write is ordered correctly on success, but its failure still permits the upload to proceed into a public session.

realtonyyoung (approved):

realtonyyoung:
NO FINDINGS

PR #675, opened 2026-08-26
Report import discovery, and run the decision the browser makes

George-Payne:
AI-2231

## What & why

The server lane and the Import screen shipped with nothing on the CLI side feeding or reading them, so against a real `kcap setup` the screen waited for ever and any answer given on it was recorded and ignored. This is both halves.

**Reporting** needs counts per repository *and* per window: "how many sessions will this selection import" is a cell, and neither margin of a table gives you one. `ImportDiscoverySummary` now buckets both from the same pass, and its windows are keyed off the same constant the report travels under, so `--discover`'s own windows and the screen's picker are one list rather than two that correspond by convention.

**Acting** is two `HandleImport` passes, because `--private` is per invocation, each scoped to the chosen window and vendor flags, with the shared pass followed by an explicit `visibility=org` write. Setup's step 6 then reports rather than prompts — it can only offer the current repository, and the screen just chose several.

## Where to look

- **The vendor filter picks which sources get scanned**, not which counts get subtracted afterwards — that is what makes every reported figure already scoped. **Only an explicit refusal drops a vendor:** the server normalises an untouched harness out of the decision, so refused and never-offered are identical on the wire, but this machine knows what it reported. `FirstRunMachineReport.Detected` is the set the screen could offer; anything outside it was never offered to refuse, and dropping it would silently discard that history.
- **Polling pauses for the import**, because two live Spectre renderables cannot share a terminal. Both lanes add their elapsed time back to the poll budget: that budget exists to catch a terminal nobody is sitting at, and a disk scan or an upload is work.
- **The decision's timestamp is a cursor, not a flag.** The server advances it only when the answer changes, so going Back and widening the window runs the wider import while re-confirming runs nothing. Stamped before the run, so a throw leaves a failed upload to `kcap import` rather than to the next tick.
- **`SetVisibilityForAll`** is what makes "everyone in your workspace" true: the profile default produces `default:org`, admitted only where the repository owner matches the configured org, which is empty on every provider but `GitHubApp`.
- **`NoReadableVendors`** is the one otherwise-silent failure — repositories chosen but no vendor this build can read, where the scan finds nothing and the summary would claim a successful import of history that never moved. An unknown window or titles answer voids the decision instead, since both name what to do with everything selected and neither has a safe guess; an unknown level costs one repository and is counted.
- The ticket says the report body cap is 32 KiB. It is **96 KiB** (`FirstRunFlowEndpoints.MaxImportBytes`), sized so 200 maximum-length repositories cannot be refused, so the CLI respects the repository cap and needs no byte cap of its own.

## Verification

Unit suites: Core 2394, Cli 3652, Daemon unchanged, App 1161, Integration 232 — all pass. `dotnet publish -c Release` clean, no IL2026/IL3050.

Eight mutations, each killed: removing the Agents gate before the scan, scanning a refused vendor, treating a never-offered vendor as refused, dropping the report retry, collapsing the decision cursor to a flag, stamping it after the run instead of before, spending the poll backstop on either lane, and reporting a missing window count as zero.

The wire is asserted in both directions against WireMock — every key the server reads by name, plus a decision parsed out of a canned poll response.


src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -107,6 +132,96 @@ public void WaitEnded() {
...
+
+    public async Task ImportAsync(FirstRunImportAnswer answer, DateOnly today, CancellationToken ct) {
+        var since = answer.Since(today);
+
+        // One pass per level, because --private is per invocation. Ordered narrowest first so a run
+        // interrupted between them has uploaded the private history rather than the shared.
+        foreach (var level in (FirstRunImportLevel[])[FirstRunImportLevel.OnlyMe, FirstRunImportLevel.Shared]) {
+            if (answer.At(level) is not { Count: > 0 } chosen) continue;
+
+            var exit = await new ImportCommand(config, profiles).HandleImport(
```
realtonyyoung: [P2] Record thrown imports as failures too

`Failed` is set only when `HandleImport` returns a non-zero code. If this await throws (for example during discovery/auth/I/O), `BrowserFirstRunFlow` catches and suppresses the exception, but this lane leaves `Failed == false`; Step 6 then prints the green “Imported …” summary, and the exception also aborts the remaining visibility pass. Catch here (or return an explicit result through the lane interface) and mark the run failed before propagating/continuing.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -214,6 +227,134 @@ async Task<FirstRunFlowResult> PollAsync(string serverUrl, string flowId, Cancel
...
+        // Nothing to scan, so nothing to run: the caller says why instead of announcing an import of
+        // history that could not move.
+        if (answer.NoReadableVendors) return TimeSpan.Zero;
+
+        var began = _clock.GetUtcNow();
+
+        progress.Importing(answer.Choices.Count, SessionsInScope(state.Discovered, answer));
+
+        try {
+            await importing.ImportAsync(answer, DateOnly.FromDateTime(_clock.GetUtcNow().UtcDateTime), ct);
```
realtonyyoung: [P2] Reuse the date that backed the discovery report

This resolves the relative window from the clock when the browser decision is executed, while discovery computed its window counts earlier via `ImportCommand.DiscoveryWindows()`. If discovery happens before UTC midnight and the user confirms after it, the UI count was built with one `--since` date and the actual import uses the next day, omitting the boundary day. Capture the discovery date in the lane state/report and reuse it here; passing one date to both visibility passes only fixes the later half of this drift.

src/Capacitor.Cli/Commands/ImportCommand.cs
```diff
@@ -1659,22 +1685,30 @@ await Parallel.ForEachAsync(
...
         // Loaded/Failed/AlreadyLoaded/SentChildContent accounting used for import counts.
-        if (forcePrivate) {
-            var toPrivatize = new HashSet<string>(importedSessionIds, StringComparer.Ordinal);
-            toPrivatize.UnionWith(privateScopeSessionIds);
-
-            if (toPrivatize.Count > 0) {
-                display.BeginPhase("Marking imported sessions private");
-                await SetVisibilityNoneForAll(httpClient, baseUrl, [.. toPrivatize]);
+        if (ExplicitVisibility(forcePrivate, shareWithOrg) is { } explicitVisibility) {
+            var touched = new HashSet<string>(importedSessionIds, StringComparer.Ordinal);
```
realtonyyoung, line 1746: [P2] Apply Shared visibility to already-loaded sessions

For the Shared pass, `touched` contains only `importedSessionIds`: chain sessions classified `AlreadyLoaded` never enter that bag, routed no-op replays are suppressed, and `privateScopeSessionIds` is populated only when `forcePrivate` is true. A rerun can therefore count an existing private session in discovery, let the user choose “Shared,” and finish green while leaving it owner-only. Track the selected/scoped already-loaded IDs for `shareWithOrg` as well so the browser's visibility choice is actually applied.

realtonyyoung (changes requested):
Three actionable findings remain at the current head: thrown import failures are still reported as success, the selected relative window can still shift between discovery and execution, and already-loaded sessions are omitted from the new Shared visibility update.

Static review only; I did not build or run tests.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -107,6 +132,132 @@ public void WaitEnded() {
...
+                Failed = true;
+
+                AnsiConsole.MarkupLine(
+                    $"  [yellow]![/] That history did not import: {Markup.Escape(ex.Message)}. "
+                  + "Run [cyan]kcap import[/] to retry it.");
+
+                continue;
+            }
+
+            if (exit != 0) {
```
realtonyyoung: [P2] Read the import result instead of its exit code

`HandleImport` deliberately returns `0` after computing and printing `FinalCounts` even when `ProbeError`/`Errored` is nonzero, so the normal partial- or all-session failure path never enters this branch. `SetVisibilityForAll` also logs and swallows failed Shared writes. In both cases this lane leaves `Failed == false` and Step 6 adds a green “Imported …” summary after the failure lines. Have `HandleImport` return/callback a structured outcome (including import and explicit-visibility failures) and derive `Failed` from that rather than the process exit code.

realtonyyoung (changes requested):
The three prior findings are addressed structurally, but the success signal still misses the importer's normal failure path.

realtonyyoung (approved):
NO FINDINGS

PR #676, opened 2026-08-26
Pick a workspace in the browser when org SSO finds more than one

George-Payne:
AI-2026

## What & why

When org SSO discovery returns two or more workspaces, the pick moves from a Spectre prompt to a page on the auth proxy. The CLI stays the authenticated party throughout: it prepares the pick with its own bearer, opens the browser at an opaque handle, and collects the answer by presenting a secret it generated and only ever sent the SHA-256 of. So the browser never authenticates, and nothing travelling through it is sufficient to read the result — no `return_url`, no `state`, no loopback listener.

One workspace still auto-selects and zero still offers to create one, so this appears only when there is a genuine choice.

## Where to look

`BrowserTenantPicker` refuses the browser wherever one cannot act on the pick, and each refusal falls back to the terminal rather than erroring — a failed pick ends the whole of `kcap setup`, so a degraded pick is always the better outcome.

Which channel signed in is **observed, not predicted**. `ChooseWorkOSFlow` reads `forceDevice` alone, but the browser leg falls through to a device code on the escape-hatch key, on `BrowserLaunchException`, and on a headless console — so `ViaDeviceGrant` is stamped where the device grant actually answers, and the picker reads that.

The returned key resolves against **discovery's** tenant array, never prepare's. Profiles are created by iterating discovery's list, so a key present only in prepare's would name a profile nobody created and dead-end the setup.

## Verification

`Capacitor.Cli.Core.Tests.Unit` 2363 passed, `Capacitor.App.Tests.Unit` 1161 passed, `Capacitor.Cli.Tests.Unit` 3649 with one pre-existing environment failure (`LaunchdStartStopTests.WriteAndBootstrap_…`, which refuses a group-writable service directory; the dev container's umask is `0002`, so `mkdir -p` yields `775`).

Inverting the channel wiring (`ViaLoopback = auth.ViaDeviceGrant`) fails two `WorkOSDiscoveryTests` cases; before those were added it failed none.

`dotnet publish -c Release` emits no IL2026/IL3050.


src/Capacitor.Cli/Commands/BrowserTenantPicker.cs
```diff
@@ -0,0 +1,174 @@
...
+            return await FallBackAsync(tenants, context, ct);
+        }
+
+        var url = $"{context.ProxyUrl}/cli/v1/picker/{prepared.Handle}";
+
+        // Printed as well as opened: launching cannot confirm a browser appeared, and a user whose
+        // browser never came up should not discover the way out ten minutes later.
+        Progress.Notice("Pick a workspace in your browser, or press a key to choose here.");
+        Progress.Notice(url);
+        launcher.TryOpen(url);
```
realtonyyoung: [P2] Fall back when the browser cannot be launched

`IBrowserLauncher.TryOpen` explicitly returns `false` when there is no browser handler, but this result is discarded. In that case the picker still enters the polling loop and can hold setup until the proxy deadline (or until the user notices and presses a key), instead of honoring the class's failure-to-terminal fallback contract. Check the return value and abandon/fall back immediately when launch fails.

realtonyyoung (changes requested):
One actionable fallback issue.

realtonyyoung (approved):
NO FINDINGS

PR #680, opened 2026-08-26
Apply the visibility answer the Agents screen already recorded

George-Payne:
AI-2282

## What & why

The Agents step asks who may read the sessions this machine records from now on, records it on `FirstRunAgentsDecidedEvent`, and serves it on the poll as `default_visibility`. No CLI read it: the field was absent from `Capacitor.Cli.Core/FirstRun`'s wire models, so it was dropped at deserialisation, and `kcap setup`'s step 3 then prompted unconditionally and wrote its own answer over it — the one place in the flow that asked a question and discarded the answer.

It now rides the Agents answer, since it comes from the same decision and the same timestamp, and step 3 spends it instead of asking again.

## Where to look

- **The two nulls are not the same, and this is the part worth reading.** The field is null both when the step is unanswered and when it was answered and left unset. Only the first should reach the prompt: the prompt's cursor starts on `org_public`, so a Return on a re-run would widen an existing `private` on a question the user had already answered. An answered-but-unset screen therefore re-writes what the profile already holds — the lane's contract for a null answer, and a no-op downstream. `SetupCommand.DecideVisibility` is the single place that decides, extracted because `HandleSetupAsync`'s interactive branches have no coverage at all (every `HandleAsync_*` test drives `--no-prompt`, which never reaches the browser leg).
- **Validated against `AppConfig.ValidVisibilities`, never forwarded.** This is the one field on the lane that persists *locally*: it lands in profile config and is stamped on every session afterwards, so a stop a newer server invented would be written to a file this build owns and read back by something whose meaning for it cannot be checked here.
- **No precedence rule against `--default-visibility`.** That flag is read only under `--no-prompt`, and the browser leg is skipped there entirely, so the two can never both be present. Interactively the flag is ignored today, before and after this change; making it live is a separate behaviour change.
- **The stop labels moved into one `VisibilityLabel`** so the prompt's converter and the browser-answer line cannot describe the same stop differently.

## Verification

Core 2425, Cli 3718, App 1161, Integration 233 — all pass, bar two unrelated timing tests that pass in isolation (`Callback_fault_losing_to_dispose_is_logged_once_and_run_settles_detached`, local IPC; and `Codex_StdoutHandshake_UnaffectedByALargeUnreachableSpoolBacklog`). `dotnet publish -c Release` clean, no IL2026/IL3050.

Five mutations, each killed. On the boundary: ignoring the validation, and never reading the field — both needed reshaping to compile, since dropping the reference trips IDE0005/IDE0051 as errors, which is analyzer protection rather than coverage. On the rule: an answered-but-unset screen falling through to the prompt, an unsettled screen treated as answered, and the kept branch inventing a fallback instead of carrying the profile's value.

The field is round-tripped through the source-generated JSON context against WireMock, present and absent — nothing else covered that, since every other test builds `FirstRunFlowResponse` directly, so a naming or AOT-binding slip would have left the profile untouched for ever with the suite green.


src/Capacitor.Cli.Core/FirstRun/FirstRunFlowModels.cs
```diff
@@ -209,6 +209,21 @@ public sealed record FirstRunFlowResponse {
...
+    /// <summary>
+    /// The default session visibility the same decision chose, as a canonical
+    /// <c>default_visibility</c> value.
+    ///
+    /// <para><b>Null is not a value</b>, and it is null in two situations that are not the same: the step
+    /// was answered and no audience set, and the step was never answered at all. Only the first says
+    /// anything about the profile — see <c>SetupCommand.DecideVisibility</c>, which separates them by
+    /// whether the step settled, because the second has told the machine nothing.</para>
+    ///
+    /// <para><b>It can name a stop this tenant does not currently offer</b>, because the server's fold
```
realtonyyoung: [P3] Keep the wire contract aligned with closed-set validation

This paragraph says a future/unsupported visibility is still written, but `FirstRunFlowOutcomes.Visibility` deliberately rejects anything outside `AppConfig.ValidVisibilities` and converts it to null, after which setup preserves the current profile. That closed-set boundary is also explicitly tested. Please document the actual drop-and-preserve behavior; the current contract states the opposite and could lead a later caller to bypass the validation.

realtonyyoung (changes requested):
Static review found one actionable documentation-contract inconsistency: unsupported future visibility values are documented as writable even though the implementation deliberately drops them at the closed-set boundary. I left the concrete correction inline. Per request, I did not build or run tests.

src/Capacitor.Cli.Core/FirstRun/FirstRunFlowModels.cs
George-Payne: Agreed, and it was the wrong way round rather than merely vague - fixed in 2c38a35. The paragraph now states drop-and-preserve and names `AppConfig.ValidVisibilities` as the boundary, so the doc points at the validation instead of inviting a bypass of it.

The behaviour was already what you describe and is pinned by tests (an unknown stop and an empty string both degrade to null); only the doc disagreed. I checked the spec and CHANGES for the same claim - both phrase it as the counterfactual argument for validating and then say a dropped value degrades to null, so they were already right.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
NO FINDINGS

PR #685, opened 2026-08-27
Report the import outcome to the first-run flow

George-Payne:
AI-2307

## What & why

The flow's `import-outcome` route folded a report into `FirstRunFlowState.ImportOutcome` and had no
caller, so the Done screen's counts caption was unreachable — and since the outcome is also the signal
that the run finished, the screen cannot tell a working import from a stalled one. This is the caller.
The route itself is still open as kcap-server#1671, so this is written against the contract on that
branch and degrades on a 404 like any other newer-server route.

## Where to look

- **Null is not `(0,0,0)`.** A run that lost a pass reports *nothing* — its sessions are unaccounted and
  three counts cannot say so, while the surviving pass's figures alone would state a clean import. An
  empty `Choices` likewise has two causes, and `IsDecline` separates a real decline from a decision
  whose every level was unreadable here.
- **The held report's retry is deliberately uncredited against the poll budget**, unlike the scan and
  the import: it runs every tick for as long as the report is refused, so crediting it would let a
  server that never accepts it stretch the 30-minute backstop into hours.
- **Status retry is opt-in per call site** — every hook, watch, daemon and MCP path shares
  `SendWithRetryAsync` and their budgets assume one attempt. An exhausted budget returns the status
  rather than throwing, because those call sites catch `HttpRequestException` only.

## Verification

Core 2459, Cli 3727, App 1161 — all pass. `dotnet publish -c Release` clean, no IL2026/IL3050. The
daemon suite fails 12–29 in this container on both this branch and an untouched `origin/main` worktree
(three runs each, overlapping ranges): contention between subprocess-heavy classes, left to CI.

Sixteen mutations, fifteen killed and one shown equivalent — swapping `answer.DecidedAt` for
`view.ImportDecidedAt`, which are the same value by construction. The retry predicate is pinned from
both sides: retrying a 4xx kills six tests, not retrying a 429 kills two, defaulting the flag on kills
the opt-in test.

The wire shape is round-tripped through the source-generated context against WireMock, present and
absent. Nothing else covered it — every other test in that file builds the request object directly, so
a naming or AOT-binding slip would leave the screen waiting for ever with the suite green.

### DIFF
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -7,6 +7,52 @@ Not release notes. Each entry is written as of the change that produced it and i
 code moves on; where an entry disagrees with the code, the code wins.
 
 
+## The import outcome reaches the first-run flow
+
+The flow's `import-outcome` route folded a report into `FirstRunFlowState.ImportOutcome` and had no
+caller, so the Done screen's counts caption was unreachable — and since the outcome is also the signal
+that the run finished, the screen could not tell a working import from a stalled one.
+
+`FinalCounts` already carried the route's three counts, so nothing is re-derived. Two things it did not
+carry: a session the `--private` preflight held back never reaches the upload and so appears in none of
+the three, and folds into `failed` (re-running retries exactly it); and a pass that threw leaves its
+sessions unaccounted, which **three counts cannot express**, so nothing at all is reported for that run.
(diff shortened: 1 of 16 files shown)


realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
NO FINDINGS

PR #686, opened 2026-08-27
[AI-2323] Document the Codex desktop-app hooks trust step in Settings → Hooks

realtonyyoung:
Users who only use the Codex desktop app never see the CLI trust prompt, so the Capacitor hooks kcap installs into `~/.codex/hooks.json` stay inert and no sessions are captured. Every place that describes the trust step said only "Codex will prompt on its next launch", which is exactly what the desktop app never does.

This adds the desktop-app path — trust the kcap hooks under **Settings → Hooks** in the app — to `kcap setup --help`, `kcap plugin --help`, and the two runtime notes printed after a Codex hooks install (setup wizard step + `kcap plugin install --codex`).

Companion website change: kurrent-io/kcap-web#200 (same fact, docs pages).


realtonyyoung:
Qodo finding 1 (README omits desktop trust path): addressed in 2317ec9f — both README trust mentions now cover the desktop-app Settings → Hooks path, and help-setup.txt gained the PATH-detection caveat (`kcap plugin install --codex` when setup did not detect Codex).

PR #692, opened 2026-08-28
[AI-2342] Mirror the browser's progress in the terminal

George-Payne:
Replaces the browser leg's row of dots with a tick per screen above a spinner line naming what the browser is waiting on. The poll loop already held a full flow response every two seconds; nothing new crosses the wire.

```
  Opening your browser to finish setup.
  If it didn't open:  https://acme.kcap.ai/setup?s=b7f3…

  ✓ Signed in
  ✓ Agents: Claude Code, Pi
  ⠋ Choose what to import in the browser
    t to carry on here  ·  ctrl+c to stop
```

- **`FirstRunFlowProgress.cs`** — `PollTick()` becomes `Waiting(step, healthy)` and `Settled(step, outcome, detail)`, carrying the closed sets `FirstRunFlowOutcomes` already defines rather than a wire type the renderer could read anything out of. `detail` has one source, `FirstRunAgentsAnswer.Labels`.
- **`BrowserFirstRunFlow.cs`** — the loop maps wire to vocabulary and announces a step when it is new, when its outcome changes, or when its decision stamp advances. Keyed on the stamp rather than presence because going Back and re-answering leaves the step settled: a tick naming the abandoned answer would be the only statement made about what step 4 installs. Never on a stamp's *absence* — the stamp is absent exactly when the decision is, so a view carrying neither must not be read as an answer becoming empty.
- **`Waiting`'s `healthy` flag** — the `SlowDown` and default arms both called `PollTick()`, so a 503ing server and a user reading a screen rendered identically. An unhealthy poll now says the server has gone quiet instead of naming a screen nobody is looking at.
- **The escape hatch** — narrowed from any key to `t`, following `OAuthLoginFlow.EscapeHatchKey`'s own pattern (named const, case-insensitive, offered only where there is a keyboard). Not Enter or Space, which get pressed by accident; not `q` or `x`, which read as "quit setup". Every other key is consumed rather than left buffered, or one stray byte is re-examined on every 200ms slice for half an hour.
- **Withdrawing the offer** — handover is one-directional: press `t` and the terminal spends what settled, but nothing reads the flow again. So the advertisement goes once a screen past the gate has an outcome. The mechanism stays live throughout, since it is also the only way out of a thirty-minute wait and a closed tab still needs it.
- **`TerminalWaitLine.cs`** — hand-rolled rather than `AnsiConsole.Status()`. A Spectre live region owns the console for the whole of one delegate, and the import runs inline mid-wait rendering its own bars; two live renderables cannot share a console. This stops and restarts as often as that takes, which is what makes the existing `Importing()` / `ImportEnded()` pair sufficient. Lines are truncated against a live `Console.WindowWidth`, and the row move writes CR as well as LF: either a wrap or a bare LF on a console without newline auto-return costs the erase arithmetic a row it does not know about.
- **`InteractiveLifetime.cs`** — the four exits there all reach `Environment.Exit`, so one `ProcessExit` registration gives the cursor back after any of them. A pre-existing bug: the import's own progress regions already left it hidden, and putting a region across the whole wait turns that into the common case.
- **`SetupCommand.BrowserAgentsSummary`** — stops restating the harnesses the live tick now names, keeping only the unrecognised-vendor warning, whose remedy is a command to run there. `BrowserImportSummary` is deliberately untouched: step 6 has nothing else to say, and what it reports is the run's outcome, including a partial import, where the live line is the choice being acted on.
- **`README.md`** — the transcript and the surrounding prose, including that the offer's withdrawal is a terminal-only property.

The rationale is in `docs/superpowers/specs/2026-08-28-ai2342-terminal-progress-design.md`.

Not in scope: the import's slot rows still name sessions by raw id. Naming them by anything human also affects plain `kcap import`, and session titles do not exist at import time — they are generated afterwards, and `SessionMetadata` has no title field — so it would have to use the repository or the first prompt.


src/Capacitor.Cli/TerminalWaitLine.cs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    void Draw() {
+        Erase();
+
+        if (!_running || _text is null) return;
+
+        var width = Width();
+
+        AnsiConsole.Markup($"  [cyan]{Frames[_frame]}[/] {Markup.Escape(Clip(_text, width - 5))}");
```
realtonyyoung, line 134: [P2] Preserve row accounting below the fixed prefix width

When `Console.WindowWidth` is 1–3, clipping only `_text` does not prevent the fixed four-cell `"  ⠋ "` prefix from wrapping. The offer has the same problem with its four-space prefix, while `_drawn` still records only one or two rows; the next `Erase()` can therefore clear the wrong terminal rows. The 80-column fallback has the same bookkeeping risk when width cannot be measured. Please disable in-place drawing when the width is unknown/too small, or make the prefixes themselves width-aware so every emitted row is guaranteed not to wrap.

realtonyyoung (changes requested):
One actionable issue found in the static review; see the inline comment on TerminalWaitLine. The fixed spinner prefixes can still wrap below their own width (and when width measurement falls back), invalidating the row count used by Erase(). No build or tests were run, as requested.

src/Capacitor.Cli/TerminalWaitLine.cs
George-Payne: Fixed in 11bfa480 — you were right that clipping the text was only half of it.

Rather than making the prefixes width-aware I took the other option you offered: the block is now drawn only when the width is known *and* at least `MinWidth`. Below that, or when the width cannot be read at all, nothing is drawn and the permanent lines stand on their own — which is exactly the redirected-output path, so the degraded mode already existed and was already tested. A later widening resumes drawing. Width-aware prefixes would have kept a spinner on a 3-column terminal that nobody could read.

The 80-column fallback is gone rather than patched: there is no safe number to stand in for an unreadable width, since one wider than the terminal wraps and one narrower is the same lie the clip exists to prevent. It returns `int?` now. The cursor is also only hidden once a row is actually drawn, so a terminal that never hosts the block never loses its cursor.

I have made the width injectable and added four cases — too narrow, unmeasurable, widening back, and narrowing mid-wait. That was overdue: this is the third time this row-accounting hazard has bitten (a cached `Profile.Width`, then a `Math.Max` floor that lied about narrow terminals, now the prefix) and none of them was reachable by a test, because the wrap happens inside Spectre's writer where the row count is the only thing observable.

src/Capacitor.Cli/TerminalWaitLine.cs
```diff
@@ -0,0 +1,182 @@
...
+    bool    _running;
+    bool    _hidden;
+
+    /// <summary>How many rows the block currently occupies. The one piece of state a wrong erase would
+    /// corrupt, so it is readable rather than inferred.</summary>
+    internal int Drawn => _drawn;
+
+    /// <summary>Whether anything is drawn in place. False makes every member here a no-op, which is
+    /// what the caller reads to decide whether a transition needs saying as a plain line instead.</summary>
+    public bool Enabled => tty;
```
realtonyyoung: [P2] Expose the width-disabled state to the renderer

When `_measure()` is null or below `MinWidth`, `Draw()` emits no block, but `Enabled` remains true. `SpectreFirstRunFlowProgress` uses `Enabled` to decide whether to print the plain handover offer and one-shot unreachable warning, so on this degraded TTY path it suppresses those fallback lines and draws nothing instead. Please make current drawability visible to the caller (or otherwise emit the fallback copy) while preserving recovery if the terminal later widens.

docs/superpowers/specs/2026-08-28-ai2342-terminal-progress-design.md
```diff
@@ -0,0 +1,220 @@
...
+frozen. It can be stopped and restarted as often as the import takes, which is what makes the existing
+`Importing()` / `ImportEnded()` pair enough and leaves the flow's own `finally { WaitEnded(); }` as the
+teardown. **Nothing had to wrap `RunAsync`.**
+
+Lines are **truncated rather than wrapped**: a wrapped line costs a row the cursor arithmetic does not
+know about, and the block would then erase the wrong rows. `Drawn` is exposed for the same reason — an
+erase that gets the count wrong takes a line the caller already committed to the scrollback with it.
+
+Four ways that arithmetic can be lost, all fixed rather than accepted, and the width took three passes to
+get right — worth recording, because each pass fixed half of it.
```
realtonyyoung: [P3] Remove change-history narration

“Took three passes” and “each pass fixed half” narrate review and implementation history, which the repository comment policy explicitly bans across every file type. Keep the durable width constraints and failure modes, but remove the account of how many iterations produced them.

realtonyyoung (changes requested):
Two remaining findings on the current head: the width-disabled path suppresses both the pinned block and the caller fallback copy, and the design note includes prohibited change-history narration. Details are inline.

src/Capacitor.Cli/TerminalWaitLine.cs
George-Payne: Fixed in b6ee269e. You are right, and the property was worse than under-exposed: its own doc comment said "what the caller reads to decide whether a transition needs saying as a plain line instead", while the implementation was `=> tty`. Documented as drawability, implemented as a TTY check — so on a narrow or unmeasurable terminal all three call sites read true, the block drew nothing, and the two plain fallbacks were suppressed. Neither shown. My design note claimed that path fell back to the redirected-output behaviour; it did not, because that behaviour is those very lines.

`Pinned` now answers for the terminal as it is now (`tty && width known && >= MinWidth`), so a widening resumes drawing and a narrowing hands the lines back to the caller, as you asked.

One case beyond the one you raised, which the same gap creates: a terminal wide at `Opening` puts the offer *in* the block, so if it later narrows the offer disappears with no plain line ever having been printed. So rather than restoring the print in `Opening` I moved it to the per-poll refresh behind a one-shot latch, which covers narrow-from-the-start and narrowed-later with one mechanism. Latched because a line already scrolled past cannot be withdrawn.

I also dropped the branch in `Say`: a permanent line goes through the block either way, and the branch was a second place holding the same assumption that got this wrong.

Tests: `Pinned` across tty/narrow/unmeasurable and a live narrowing, plus the offer decision as a predicate over its four inputs. The writing itself is still uncovered — it goes through Spectre's console — so what is pinned is the decision, not the output.

docs/superpowers/specs/2026-08-28-ai2342-terminal-progress-design.md
George-Payne: Removed in b6ee269e — the pass count is process, not design, and a future reader cannot act on it. The durable width constraints stay, and I have added what was missing: why drawability is exposed to the caller rather than inferred.

One note for the record rather than a disagreement: the policy you cite governs comments (it says it applies to `.editorconfig`, `.props` and YAML as much as to C#), and `CLAUDE.local.md` routes reasoning *to* the spec as its proper home — so a design note is not barred from carrying argument. It is that a count of iterations is not argument. Same conclusion, narrower ground.

src/Capacitor.Cli/TerminalWaitLine.cs
```diff
@@ -0,0 +1,194 @@
...
+    /// <summary>
+    /// Whether a pinned block is being drawn <b>right now</b>: false with output redirected, and false
+    /// too on a terminal too narrow to host one or whose width cannot be read.
+    ///
+    /// <para>The caller reads this to decide whether a line has somewhere to change in place or has to
+    /// be said outright, so it has to answer for the current terminal rather than for the process — a
+    /// property meaning "is a TTY" cannot answer it, and one that only looked like it did suppressed
+    /// the caller's plain-line fallback on exactly the terminals that needed it.</para>
+    /// </summary>
+    public bool Pinned => tty && Hostable();
```
realtonyyoung: [P2] Base the fallback on the draw that actually happened

`Pinned` takes a fresh width sample after `Show` has already called `Draw`, so it can disagree with the block on screen. For example, if `Draw` observes an unreadable width and emits nothing but this subsequent read returns 80, `Refresh` suppresses the plain offer even though `_drawn` is zero—the original missing-fallback path remains. The reverse ordering can print a fallback while a block was drawn. Please expose the result/state of the completed draw under the existing lock rather than independently re-measuring here; the next draw can still recover when the terminal widens.

src/Capacitor.Cli/TerminalWaitLine.cs
```diff
@@ -0,0 +1,194 @@
...
+    void Draw() {
+        Erase();
+
+        if (!_running || _text is null) { ShowCursor(); return; }
+
+        // Nothing is drawn rather than drawn wrong. The caller then says its lines outright instead,
+        // which it decides from `Pinned`. A later widening resumes.
+        if (!Hostable()) { ShowCursor(); return; }
+
+        var width = _measure()!.Value;
```
realtonyyoung: [P2] Use one width snapshot for the draw

`Hostable()` reads the live width, then this line reads it again. The terminal can resize—or the second query can become unreadable—between those calls: a first value of 80 followed by null throws on `.Value`, while 80 followed by 3 draws the fixed prefix at a width already known to wrap. Capture one nullable width and validate/use that same value for the entire draw.

src/Capacitor.Cli/TerminalWaitLine.cs
```diff
@@ -0,0 +1,194 @@
...
+    internal int Drawn => _drawn;
+
+    /// <summary>
+    /// Whether a pinned block is being drawn <b>right now</b>: false with output redirected, and false
+    /// too on a terminal too narrow to host one or whose width cannot be read.
+    ///
+    /// <para>The caller reads this to decide whether a line has somewhere to change in place or has to
+    /// be said outright, so it has to answer for the current terminal rather than for the process — a
+    /// property meaning "is a TTY" cannot answer it, and one that only looked like it did suppressed
+    /// the caller's plain-line fallback on exactly the terminals that needed it.</para>
```
realtonyyoung: [P3] Remove the old-implementation narration from comments

This paragraph explains how the former TTY-only property failed, which is change history prohibited by the repository comment policy. The same pattern was added in `SetupCommand.cs` (“the dots this change removed”), `SetupCommandTests.cs` (“what a Pinned that only meant is a TTY produced”), and `TerminalWaitLineTests.cs` (“a version meaning is a TTY suppressed…”). Please retain only the durable drawability and fallback invariants at those sites.

realtonyyoung (changes requested):
Re-review of current head `b6ee269e`: the caller can still disagree with what `Draw` actually rendered because width is sampled independently, and `Draw` itself reads the live width twice. I also found new change-history narration in source and test comments. Details are inline. Static review only; I did not build or run tests.

src/Capacitor.Cli/TerminalWaitLine.cs
George-Payne: Fixed in 573127d3. `Pinned` is now the row count read under the block's own lock, so it reports the draw that happened rather than guessing whether one would. You are right that re-measuring answered a different question from the one the caller asks, and could disagree with the screen in either direction — I had reintroduced the missing-fallback path through the back door. The next draw is what recovers a widened terminal, and the frame timer keeps that under 100ms.

src/Capacitor.Cli/TerminalWaitLine.cs
George-Payne: Fixed in 573127d3 — one sample now serves the whole draw. This was worse than a wrong width: `_measure()!.Value` after a `Hostable()` that had already consumed a read would throw on a second read of null, and it does so inside a frame-timer callback where nothing catches it. So a resize at the wrong moment could take the process down, not just draw at a stale width. Test added that feeds 80 then null then 3 through one draw; it throws without the fix.

src/Capacitor.Cli/TerminalWaitLine.cs
George-Payne: Removed in 573127d3, at all four sites — the invariants stay, the account of what preceded them goes. Fair hit, and not a small one: I agreed with you about the design note and then seeded the same pattern across four files in the very next commit.

realtonyyoung:
NO FINDINGS

realtonyyoung (dismissed):
The current head resolves the width-sampling and fallback-state races and removes the historical source/test narration. Static review only; I did not build or run tests.

George-Payne:
Merged `main` in (rather than rebasing, per the repo's own preference) to clear the conflict with #693. Two conflicts, and one needed a decision rather than a pick-a-side, so flagging it for the re-review since it is in the file you looked at most closely.

**`InteractiveLifetime.cs`.** #693 introduced `Interrupt()` as the single funnel every interrupt path takes — Ctrl+C, both signals, the watchdog — with a doc comment saying anything a command needs to say on its way out has to be said there. That is exactly the seam my `AppDomain.ProcessExit` registration was working around, so I moved the cursor restore into `Interrupt()` and deleted the parallel mechanism rather than keeping both. It runs first, before `RunBeforeExit` spends its budget on the network, so a hanging notice cannot cost the shell its cursor.

What that gives up is that `ProcessExit` also fired on a normal `Main` return — already covered by the leg's own `finally { WaitEnded(); }`, and it never covered a crash, since the CLR does not raise `ProcessExit` for an unhandled exception. So the redundancy was buying nothing.

**`BrowserFirstRunFlowTests.cs`** was a clean split: your `OnWaitEnded` hook kept, `PollTick` replaced by `Waiting`/`Settled`.

I also read the auto-merged `RunAsync` rather than trusting it, since #693 rewrote the region this PR changed. Ordering composes: `Opening` → armed interrupt → poll → `finally { WaitEnded(); }` → `notice.SendAsync`, so the pinned block is down before any relinquish output.

Green after the merge: `BrowserFirstRunFlowTests` 102, `FirstRunInterruptRelinquishTests` 12, `TerminalWaitLineTests` 18, `SetupCommandTests` 94, whole solution builds. CI had passed on the pre-merge head and is re-running from scratch on this one.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
The merge-conflict resolution preserves the reviewed terminal-progress behavior and composes correctly with the interrupt-relinquish path from main. Static review only; I did not build or run tests.

PR #693, opened 2026-08-28
Tell the server when this machine stops listening

George-Payne:
AI-2343

## What & why

When the browser setup leg ends without finishing, the browser has no way to know: it keeps offering
decisions nothing will act on, and the PATH-fix button records a request that stays outstanding for ever
while the screen says it asked your machine. This adds `POST /api/first-run/flows/{id}/relinquish` to the
flow client and sends it from one place in `RunAsync`, keyed off the leg's result type — so a closed tab
and a failed leg are covered without a special case for the keypress.

Two reasons cross, because the browser renders opposite remedies for them. A dismissal is the only
handover: the terminal carries on with everything the browser settled, so it is the one exit where the
page must not send anyone back to `kcap setup`. `Abandoned` and `Failed` are `stopped`. `Finished` must
not send it at all, and that guard is the load-bearing one.

## Where to look

`FirstRunInterruptRelinquish` is a process-global static, which is what an interrupt needs: Ctrl+C,
SIGTERM, SIGHUP and the parent-liveness watchdog all reach `Environment.Exit` from a handler that runs no
`finally`, and the handlers are installed process-wide with nothing command-scoped to reach through. The
leg arms it for its own duration; `Disarm` compares before clearing so a second leg's registration is not
dropped by the first leg's handle going out of scope. Tests touching it carry bare `[NotInParallel]`.

## Verification

`BrowserFirstRunFlowTests` 86/86 and `FirstRunFlowClientTests` 32/32. `dotnet publish -c Release` clean
of IL2026/IL3050. Driven end to end against a live tenant: a flow relinquished under an open tab renders
the terminal page, and both reason tails were checked in the browser, including an unreadable token
falling back to `stopped`.

### DIFF
--- a/src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
+++ b/src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
@@ -118,10 +118,57 @@ public sealed class BrowserFirstRunFlow(
         progress.Opening(setupUrl);
         launcher.TryOpen(setupUrl);
 
+        // Armed for the leg's whole duration, because an interrupt runs no finally: Ctrl+C, SIGTERM and
+        // SIGHUP all reach Environment.Exit from a handler, and this is the only thing between that and a
+        // browser left offering decisions nobody will act on.
+        using var armed = FirstRunInterruptRelinquish.Arm(
+            token => RelinquishAsync(serverUrl, flowId, FirstRunRelinquishReasons.Stopped, token));
+
+        FirstRunFlowResult result;
+
         try {
-            return await PollAsync(serverUrl, flowId, report, ct);
+            result = await PollAsync(serverUrl, flowId, report, ct);
         } finally {
             progress.WaitEnded();
+        }
+
+        // ONE call, keyed off the result type, so the closed tab and the failed leg are covered without a
+        // special case for the keypress.
+        //
+        // Finished must not relinquish, and that guard is the load-bearing one: the flow is over on its own
+        // terms and the browser is rendering the payoff. Expired needs nothing either — the server already
+        // refuses a flow past its lifetime, and the page says so.
+        if (ReasonFor(result) is { } reason) await RelinquishAsync(serverUrl, flowId, reason, ct);
+
+        return result;
+    }
+
+    /// <summary>
+    /// What to tell the browser about this ending, or null when there is nothing to tell it.
+    ///
+    /// <para><b>A dismissal is a handover and nothing else is.</b> It is the one exit where the terminal
+    /// carries on with everything the browser settled, so it is the one where the page must not send anyone
+    /// back to <c>kcap setup</c>. A backstop that elapsed and a leg that failed both leave nothing
+    /// running.</para>
+    /// </summary>
+    static string? ReasonFor(FirstRunFlowResult result) => result switch {
+        FirstRunFlowResult.Dismissed => FirstRunRelinquishReasons.Handover,
+        FirstRunFlowResult.Abandoned or FirstRunFlowResult.Failed => FirstRunRelinquishReasons.Stopped,
+        _ => null
+    };
+
+    /// <summary>Says the machine has gone. Swallows everything: this runs as the leg ends, and a setup that
+    /// otherwise worked must not report a failure because one best-effort POST did not land.</summary>
+    async Task RelinquishAsync(string serverUrl, string flowId, string reason, CancellationToken ct) {
+        try {
+            await channel.RelinquishAsync(serverUrl, flowId, reason, ct);
+        } catch (OperationCanceledException) when (ct.IsCancellationRequested) {
+            // Propagated, as every other await in this class does: a caller's cancel is not this method's
+            // to swallow, and the poll would already have thrown on one.
+            throw;
+        } catch (Exception) {
+            // The channel already degrades a transport failure to a status code, so reaching here means
+            // something unexpected — and there is still nothing useful to do about it.
         }
     }
 
--- a/src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
+++ b/src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
@@ -124,46 +124,50 @@ public sealed class BrowserFirstRunFlow(
         progress.Opening(setupUrl);
         launcher.TryOpen(setupUrl);
 
-        FirstRunFlowResult result;
+        // ONE write decides the reason, and it is the poll's own assignment — so an interrupt either sees
+        // a leg still waiting (the machine is dying: `stopped`) or the reason its result names, never a
+        // half-decided state. Computing the reason into a second variable would reopen exactly that gap.
+        FirstRunFlowResult? settled = null;
 
-        // Scoped to the POLL, and that is load-bearing rather than tidiness. An interrupt runs no finally
-        // — Ctrl+C, SIGTERM, SIGHUP and the parent watchdog all reach Environment.Exit from a handler — so
-        // something has to be armed while the leg waits. But this callback can only ever say `stopped`, so
-        // leaving it armed past the poll would let it race the reason the result actually names: a
-        // dismissed leg would gamble `handover` against `stopped`, and a FINISHED one — which must never
-        // relinquish — could do so at all.
-        using (_interrupts.Arm(
-                   token => RelinquishAsync(serverUrl, flowId, FirstRunRelinquishReasons.Stopped, token))) {
-            try {
-                result = await PollAsync(serverUrl, flowId, report, ct);
-            } finally {
-                progress.WaitEnded();
-            }
+        using var notice = _interrupts.Arm(token =>
+            ReasonToSend(Volatile.Read(ref settled)) is { } reason
+                ? RelinquishAsync(serverUrl, flowId, reason, token)
+                : Task.CompletedTask);
+
+        try {
+            Volatile.Write(ref settled, await PollAsync(serverUrl, flowId, report, ct));
+        } finally {
+            progress.WaitEnded();
         }
 
-        // ONE call, keyed off the result type, so the closed tab and the failed leg are covered without a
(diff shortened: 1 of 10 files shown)


src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -118,10 +124,60 @@ public async Task<FirstRunFlowResult> RunAsync(
...
+        // dismissed leg would gamble `handover` against `stopped`, and a FINISHED one — which must never
+        // relinquish — could do so at all.
+        using (_interrupts.Arm(
+                   token => RelinquishAsync(serverUrl, flowId, FirstRunRelinquishReasons.Stopped, token))) {
+            try {
+                result = await PollAsync(serverUrl, flowId, report, ct);
+            } finally {
+                progress.WaitEnded();
+            }
+        }
```
realtonyyoung, line 139: [P2] Make interrupt handoff atomic

Ending the `using` before the result-specific send does not close the race. `RunBeforeExit` can read `_pending` first, then this scope can dispose, after which the captured `stopped` callback still runs and races a dismissed `handover` or a finished result. In the opposite ordering, an interrupt after disposal finds no callback and `Environment.Exit` can abort the normal relinquish POST. Please make the interrupt and normal paths contend for one atomically consumed transition carrying the selected reason, rather than using read/disarm followed by a separate direct send.

test/Capacitor.Cli.Core.Tests.Unit/FirstRun/BrowserFirstRunFlowTests.cs
```diff
@@ -1521,4 +1586,175 @@ public async Task Neither_lane_spends_the_backstop_it_was_not_waiting_through()
...
+        h.Channel.Creates.Enqueue(new(404, null));
+
+        var result = await Run(h);
+
+        await Assert.That(result).IsTypeOf<FirstRunFlowResult.Unavailable>();
+        await Assert.That(h.Channel.Relinquished).IsEmpty();
+    }
+
+    /// <summary>It is the last thing the leg does, so a failure has nowhere to go: what it costs is the
+    /// browser falling back to the behaviour it had before the route existed.</summary>
```
realtonyyoung: [P3] Remove the remaining historical narration

This comment still describes behavior “before the route existed,” which the repository comment policy explicitly bans. Please state only the durable behavior this test pins: a refused best-effort relinquish leaves the leg result unchanged and the browser waits until the flow expires.

realtonyyoung (changes requested):
Two actionable findings from the static review: the interrupt-to-normal relinquish handoff is still non-atomic, and one newly added test comment still contains prohibited historical narration. Details are attached inline. No build or tests were run, as requested.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
George-Payne: You're right, and the read-then-invoke ordering is exactly why — `RunBeforeExit` took the callback into a local and ran it afterwards, so narrowing the scope only shrank the window rather than removing it. Both orderings you name were live.

Fixed in 5f82133 with a single atomically-consumed send:

- `FirstRunNotice` holds one `Interlocked` claim. Either path may win; the winner sends, the loser returns.
- The loser **waits out** the send in flight (same 1s budget) instead of exiting through it, which closes the "interrupt after disposal aborts the POST" ordering.
- An interrupt that wins sends whatever the reason is at that instant, and that is the honest outcome: the process is being killed, so nothing is carrying on however the poll ended.

On "carrying the selected reason": the reason is now read from the poll's own assignment (`Volatile.Write(ref settled, await PollAsync(...))`), so there is **one** write rather than a decided result plus a separately stored reason. An interrupt therefore observes either a leg still waiting — `stopped` — or the reason the result names, never a half-decided state. That also removes the residual window a compute-into-a-second-variable shape would have left on `Finished`.

Verified by reverting the production change and re-running: `An_interrupt_racing_the_legs_own_send_produces_exactly_one` and `An_interrupt_that_wins_the_claim_is_the_only_send` both go red against read-then-send, green with the claim. `FirstRunInterruptRelinquishTests` covers the arbitration directly, including the loser's wait and the budget bound.

test/Capacitor.Cli.Core.Tests.Unit/FirstRun/BrowserFirstRunFlowTests.cs
George-Payne: Fixed in 5f82133, using your wording: it now states that a refused best-effort relinquish leaves the leg's result unchanged and the browser waits until the flow's own lifetime ends it.

My earlier sweep for that phrase covered `src/` and missed the test. I've since run it over the whole branch diff — `before … existed`, `no longer`, `previously`, `used to`, `originally`, `today that is`, `this replaces`, `moved from` — and there are no other instances.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -118,11 +124,65 @@ public async Task<FirstRunFlowResult> RunAsync(
         progress.Opening(setupUrl);
         launcher.TryOpen(setupUrl);
 
+        // ONE write decides the reason, and it is the poll's own assignment — so an interrupt either sees
+        // a leg still waiting (the machine is dying: `stopped`) or the reason its result names, never a
+        // half-decided state. Computing the reason into a second variable would reopen exactly that gap.
+        FirstRunFlowResult? settled = null;
+
+        using var notice = _interrupts.Arm(token =>
+            ReasonToSend(Volatile.Read(ref settled)) is { } reason
```
realtonyyoung: [P2] Keep the interrupt reason independent of the settled result

If `PollAsync` has published `Dismissed` but an interrupt wins the claim before `notice.SendAsync`, this shared callback reads `Dismissed` and posts `handover`; `Environment.Exit` then kills the process, so the browser is told the terminal took over when nothing is running. The interrupt-first test fires inside `PollAsync`, before publication, and only covers the null/`stopped` case. The claimant needs to supply its own reason: interrupts always `stopped`, while the normal path maps the settled result (including the no-op finished case), all behind the same atomic claim.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -118,11 +124,65 @@ public async Task<FirstRunFlowResult> RunAsync(
         progress.Opening(setupUrl);
         launcher.TryOpen(setupUrl);
 
+        // ONE write decides the reason, and it is the poll's own assignment — so an interrupt either sees
+        // a leg still waiting (the machine is dying: `stopped`) or the reason its result names, never a
+        // half-decided state. Computing the reason into a second variable would reopen exactly that gap.
```
realtonyyoung: [P3] Keep review history out of the source comment

“Reopen exactly that gap” describes the history of the race and its fix, which the repository comment policy prohibits. Please retain the durable synchronization invariant here, but remove the reference to the earlier implementation defect.

realtonyyoung (changes requested):
Two remaining findings on the current head: an interrupt can still send the normal-path handover reason after `Dismissed` is published, and the new synchronization comment includes prohibited review history. Details are inline.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
George-Payne: You're right, and the shared callback contradicted my own stated intent — I designed for "an interrupt means `stopped`" and then had both paths read the reason from `settled`, which is precisely the `Dismissed` → `handover` case you describe. Fixed in d5dc641.

Each claimant now supplies its own reason, behind the same atomic claim:

- `FirstRunNotice` takes `send(reason, ct)` plus an `interruptReason()` resolver. The leg passes `ReasonFor(settled)`; the interrupt path calls the resolver.
- `InterruptReason` never returns `handover`.
- A null reason claims without sending, which is how a result that must not be relinquished shuts the interrupt out as well as itself.

One deliberate step past "interrupts always `stopped`": the interrupt consults the published result to decide **whether** there is anything to say, never **which** reason. Literally always-`stopped` would let an interrupt landing just after `Finished` replace the payoff screen with "your machine stopped waiting", and `Finished` must not relinquish at all. So `InterruptReason` is `stopped` when the leg would have sent something, and null when it would not. Happy to drop the refinement if you'd rather have the simpler rule.

Your point about the test was the sharper half: `An_interrupt_that_wins_the_claim_is_the_only_send` fires inside `PollAsync`, so it only ever covered the null → `stopped` arm. Added `An_interrupt_after_a_dismissal_still_says_the_machine_stopped`, which fires from `WaitEnded` — after publication, before the leg's send, i.e. your exact window. Verified by reverting to a shared reason: it goes red with `Expected to be equivalent to [stopped]`, green with the split. `An_interrupt_after_a_finish_says_nothing` covers the other arm, and `The_claimant_supplies_the_reason` pins the rule directly in both orderings.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
George-Payne: Gone in d5dc641 — the block was rewritten for the reason split, and the comment now states only the invariant: the poll's own assignment is the single write that publishes the result, so an interrupt sees either a leg still waiting or a settled one, never a half-decided state.

I've re-run the sweep over the whole branch diff for that family (`reopen`, `earlier version/implementation`, `defect`, `no longer`, `previously`, `used to`, `originally`, `before … existed`, `today that is`, `this replaces`, `moved from`) and there are no other instances.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
The current head resolves the claimant-specific reason race and removes the historical source narration. Static review only; I did not build or run tests.

PR #694, opened 2026-08-28
Give each harness its own module and carry its identity as an enum

### LINKED ISSUE #691: Adding a harness edits five shared sites before any vendor code
## Problem

Adding a harness edits shared code in five places — a `HarnessPaths` property and its factory line, an `AgentDetectionResult` field, an arm of `AgentDetection.Detect`, and a `HarnessCatalog` row — before a line of vendor-specific code is written. The layout rule in CLAUDE.md asks for the opposite: a new `Harness/<Vendor>/` directory plus one registration site per assembly.

Two symptoms show the shape is wrong rather than merely verbose:

* **`KnownHarness.IsWired` takes the whole bundle.** A vendor's wiring check needs one vendor's layout and receives all nine, because the delegate cannot name the type it wants.
* **`KnownHarness.Select` exists only to index a positional record.** Every consumer already loops the catalogue — `HarnessInventory`, `HarnessNudge`, `FirstRunMachineReport`, the desktop app's Agents and Import steps — so the selector is a workaround for the result not being keyed.

The codebase already has the shape this wants, three times over: `IImportSource`, `IHostedAgentLauncher` and the reviewer runtime factories are each an interface with one implementation per vendor, registered in a list. Paths and detection are the outlier.

## Approach

One interface per question a caller asks, one type per vendor answering only the questions it can, one list where they register. A vendor's paths become a private field of its module — nothing outside asks for `ClaudePaths`. Identity and capability declarations are static, so a vendor is described rather than instantiated to be asked; the per-process layout stays an instance, since it is composed from `UserHome` plus that vendor's own override.

Consumers become queries over the registry rather than switches: the PATH pass, the status line and the nudge, import sources, skills sync. Absence of a capability is the absence of an interface, not a sentinel — Cursor has no PATH probe by design, Claude and Codex have no on-disk marker, and today those are `[]` and `false` with a comment.

`HarnessPaths` retreats to a registry internal and then disappears; `AgentDetectionResult` stops naming every vendor as a member.

## Payoff

* adding a vendor is a directory plus a registration line, with nothing in shared code enumerating vendors
* `KnownHarness.Select`, `HarnessWiring`'s hop through the catalogue, and the nine-member bundle all go
* a capability a vendor cannot implement becomes a signal to add a capability rather than widen an existing one
* no reflection and no open generics, so it costs nothing under NativeAOT

## Out of scope

`PluginCommand`'s nine near-parallel install branches. Behind a wire/unwire capability they are the largest payoff and the largest risk; they want their own design pass.

Design: `docs/superpowers/specs/2026-08-27-harness-modules-design.md`.

Inok:
Closes #691 — AI-2348

## What & why

Adding a harness meant editing five shared sites before any vendor code: a catalogue, a detection
switch, a wiring table, an install-flag list, a wizard order. Each vendor now owns its module —
paths, detection signals, wiring probe — and one `HarnessRegistry` instance is built at the entry
point and passed down, so shared code never names a vendor. Identity travels as the `HarnessId`
enum; the lowercase id strings live only where something outside picks the spelling: hook URLs, the
transcript-batch tag, the offer-ledger keys on disk, the session-memory token, the lines we print.

## Where to look

`HarnessNames` is a single C# 14 extension block — `id.VendorId`, `id.Flag`, static
`HarnessId.From`. A second block declares a second generated member named `extension`, which CA1708
rejects, so the set-level helper stays a classic extension method. The registry caches no answer on
purpose: the daemon holds one for its lifetime and must still see a vendor installed after it was
built.

## Verification

Five suites green: Core 2468, Cli 3759, Daemon 2776, App 1229, Integration 233 — 0 failed.
`dotnet publish -c Release` reports no IL2026/IL3050. `ImportVendorSelectionOutputTests` pins the
two explicit-selection stderr lines and fails when a vendor id reaches them as `Skipping OpenCode`
rather than `Skipping opencode`.


realtonyyoung (approved):
Static review only, as requested; I did not build or run tests.

realtonyyoung:
NO FINDINGS

PR #700, opened 2026-08-30
[AI-2280] Drop the stale Claude-Code-only qualifier from mcp workitems help

realtonyyoung:
## What & why

`kcap --help` still marks `mcp workitems` as "(Claude Code only)", but the workitems MCP server registers on every supported harness — the qualifier tells non-Claude users the tool is unavailable to them when it is not. One line in the help resource.


realtonyyoung:
NO FINDINGS

PR #704, opened 2026-08-31
Let the first-run flow enable the daemon as a service

George-Payne:
AI-2048

## What & why

`kcap daemon service ensure` shipped with no caller in the product. The first-run flow's Done screen is where the ask belongs, so this adds a `daemon_service` machine capability the browser can request, and reports on the create whether the service is already enabled — the browser cannot see a service on someone's laptop, so without that there is no offer to draw.

The ladder is split from its verb: `EvaluateEnsureAsync` decides and mutates, `Ensure` reports, and both the terminal and the flow reach the machine through the one ladder, so a screen cannot render an outcome the verb would not have printed.

## Where to look

`EnsureFlowMap`. The flow's outcome and reason sets are closed and the server refuses a token it does not know, after which the CLI retries for ever and the screen waits on an answer that already happened — so the ladder's reasons collapse under one rule: `refused` means nothing was mutated, `failed` means a transaction ran and did not land. Two rows are not what a first pass writes. Viability refuses rather than fails, because it is proven before anything destructive and no retry can help an unusable pinned URL. And the start gate lands on both sides of the rule: its family refuses, declining ahead of the marker write, while its TOCTOU re-check fails, firing after a bootout.

The reported fact is offerability rather than state — false only where `ensure` has a verb it could actually run, since the classifier is half of the refusal surface and the arms add three more after it.

## Verification

- `Capacitor.Cli.Tests.Unit`, filtered to the shim and ensure classes: 44 passed.
- `Capacitor.Cli.Core.Tests.Unit`, filtered to the flow classes: 102 passed.
- `dotnet publish -c Release`: no IL2026/IL3050.
- The mapping is pinned by exhaustiveness rather than by example: a new `VerifyExit` member or `StartGateReason` fails the tests with nowhere to go, instead of inheriting the tail and a retry button.


src/Capacitor.Cli/Commands/ServiceEnsure.cs
```diff
@@ -154,6 +155,60 @@ public static (string? Recovery, string? Reason) Map(
...
+
+        // Rows the ladder will not touch, plus the whole start-gate family: that gate refuses in its
+        // first phase, before the marker write, so nothing was mutated — and every reason it gives means
+        // something else owns or has invalidated the install, which no retry can change.
+        "status_unknown" or "orphan_label" or "stale_marker" or "running_unconfirmed"
+            or "directive_missing" or "directive_invalid" or "identity_mismatch" or "foreign_binary"
+            or "evidence_unreadable" or "package_inconsistent" =>
+            Refused(FirstRunMachineActionReasons.NeedsAttention),
+
+        _ => new(FirstRunMachineActionOutcomes.Failed, null),
```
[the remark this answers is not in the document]
George-Payne: The mechanism is right, the consequence is not: the retry is not lost. `DaemonRetryable` on the browser side includes `failed`, so the card still draws `Try again` — what differs between `failed` and `refused/service_busy` is only the sentence.

And `failed` is the better bucket for this exit. Off launchd, exit 1 from the plain arms is lock contention *or* a manager error, and nothing distinguishes them:

- As `failed` the card says "That did not work, and nothing is running. Your terminal has the details." True of both, and the terminal genuinely holds either the lock timeout or the systemctl/schtasks error.
- As `service_busy` a manager error would read "Something else was changing the service" — false, and it invents a cause.

Making them separable means giving the plain arms their own exit code for a lock timeout, which changes `kcap daemon service`'s published exit contract; that is a wider change than this PR should carry.

On launchd the case is already precise: the verified engine returns `verify_contended` for both the lock timeout and the install arm's ownership refusal, and that maps to `refused`/`service_busy` — with copy that deliberately avoids naming a transaction, since the ownership row is a unit appearing rather than one running.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -1420,15 +1452,16 @@ async Task<BrowserFlowAnswers> RunBrowserFlowStepAsync(string serverUrl, string
 
                 var report = FirstRunMachineReport.EvaluateCurrent(
                     config, HarnessRegistry.FromEnvironment(home),
-                    Environment.MachineName, await LoginShellFindsCliAsync());
+                    Environment.MachineName, await LoginShellFindsCliAsync(),
+                    await DaemonServiceCommands.FlowServiceEnabledAsync(config, profiles, home));
```
realtonyyoung: [P1] Bind the service action to the profile selected by this setup run

`profiles` is the immutable context resolved at process startup, but this code runs after discovery may have activated a different tenant and before Step 5 saves the chosen daemon name. If setup starts on profile A and discovery selects profile B, both `FlowServiceEnabledAsync` and the later `FlowEnsureAsync` still inspect and mutate A; on a fresh run, choosing a non-default daemon name later can likewise leave the just-installed default service detached from the saved profile. Please build the service action from the tenant/profile and daemon name selected by this setup run, or defer the action until those settings have been committed.

realtonyyoung (changes requested):
Static review of 161d0863. I found one blocking setup-context issue and left it inline. No builds or tests were run.

src/Capacitor.Cli/Commands/SetupCommand.cs
George-Payne: Agreed, and thank you — the second half in particular. Fixed in 4fdb1fc6 by deferring rather than re-resolving.

Re-resolving at action time would only have fixed the tenant half: the poll loop still runs before Step 5, so the daemon name would still have been the startup one and the orphan-service case survived. So the capability is no longer performed in the browser leg at all.

- `SetupMachineActions` advertises `path_shim` only. The loop already leaves an unadvertised capability outstanding rather than reporting it, which is exactly the "asked, waiting" state the browser renders — so nothing was needed on the server side for this.
- A new `SetupDaemonService` step runs after the profile write, polls the flow once (fresh, since the browser can still be open and the press can land during the steps in between), performs any outstanding `daemon_service` request, and reports the outcome.
- It is handed `saved` — the `ProfileContext` built from the values setup just wrote, not a re-resolution. That is the same reasoning the existing comment there gives: CLI, environment and repository precedence can each land somewhere other than what this run chose.
- The report is retried a bounded three times, because unlike the poll loop there is no next tick; the action itself runs once.

The browser copy shifted with it — it now says the work happens "as the rest of setup finishes" rather than "once the import finishes", which is both what it waits for and true whether or not an import is running.

Worth noting one thing your review changed beyond the code: the design spec had recorded this same ordering as a *safety* property, on the grounds that the fact and the action agree because neither sees the new config. Agreeing on the wrong value is not safety, and the spec now says so.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -866,6 +886,23 @@ await ConfigMutator.MutateAsync(config, c => {
         var saved = new ProfileContext(
             new(serverUrl, activeName, defaultProfile, null), await AppConfig.LoadProfileConfig(config));
 
+        // Here rather than in the browser leg, and after the write above rather than before it: the unit
+        // bakes the profile, the expected server and the daemon name, and `saved` is the only context that
+        // carries what this run actually chose.
+        if (browserAnswers.FlowId is { } browserFlowId) {
+            try {
+                using var deferred = new HttpClient();
```
realtonyyoung: [P1] Authenticate the deferred flow client

`FirstRunFlowClient` expects its supplied `HttpClient` to already carry the caller bearer, but this raw client has no authentication handler or token. Every first-run flow route is authenticated, so the poll receives 401; `SetupDaemonService.RunAsync` then sees a null body and quietly returns without enabling the service. Please construct this client through `HttpClientExtensions.CreateClientWithAuthStatusAsync(config, saved, serverUrl, autoRetryUnauthorized: true)` (or an equivalent authenticated path).

src/Capacitor.Cli/Commands/SetupCommand.cs
realtonyyoung: Deferring the mutation fixes the execution half, but the create report still calls `FlowServiceEnabledAsync(config, profiles, home)` with the immutable startup `profiles`. That stale value controls whether the browser offers/requests the daemon-service action at all: selecting profile B after starting on profile A can still suppress or misstate the offer based on A. Please compute offerability from the profile selected by this setup run (or represent it without consulting pre-setup profile state).

realtonyyoung (changes requested):
Two blockers remain in the revised first-run service setup: the deferred flow client is unauthenticated, and the browser action offer is still derived from the immutable startup profile context. Details are inline.

src/Capacitor.Cli/Commands/SetupCommand.cs
George-Payne: Fixed in ad6f0382 — and this one was the worse of the two, because it fails silently. The poll's 401 gives an empty body, `RunAsync` reads that as "nothing was asked", and returns without enabling anything. No error, no line in the terminal, and the browser goes on saying it asked.

It now goes through `CreateClientWithAuthStatusAsync(config, saved, serverUrl, autoRetryUnauthorized: true)` — the same choke point the browser leg uses — and built from `saved` rather than the startup context, so the bearer is the one for the profile setup landed on.

src/Capacitor.Cli/Commands/SetupCommand.cs
George-Payne: You are right, and chasing it turned up something worse than a stale read: the fact could not be made honest at all, so it is gone.

Computing offerability from the profile this run selected is not available at the create. The create runs before login picks a tenant, and the service *id* comes from `DaemonNameResolver.Resolve([], profiles.DaemonName)` — a daemon name setup does not collect until Step 5, after the browser leg. So the fact was a claim about a service whose identity is chosen after the screen has already been shown. There is no point in the flow where it becomes true.

Your second branch — represent it without consulting pre-setup profile state — leaves two options. A platform-only fact ("does this OS have a service manager") is profile-independent but true for macOS, Linux and Windows alike, so it would be a wire field, an event field and a fold that decide nothing. The other is to drop it, which is what ad6f0382 does: the offer is unconditional, and the ladder decides at action time where the profile is known.

The cost is that someone who already has a service-installed daemon is offered it and gets `already_enabled` back, which reads honestly but is a wasted click. We took that deliberately: someone running `kcap setup` has very likely not set it up.

Gone with it: `daemon_service_enabled` on the create body, the field on the machine report, and `FlowServiceEnabledAsync`. The server side drops the matching wire field, event field and fold.

src/Capacitor.Cli/Commands/SetupCommand.cs
realtonyyoung: The factory is the right choke point, but it does not guarantee that the returned client is authenticated; that is why it returns `AuthStatus`. Discarding the status here preserves the same silent path for `Expired`, `Missing`, or `WrongServer`: the client is left unauthenticated, the poll receives 401, and `RunAsync` treats the empty body as no request. Please require `AuthStatus.Ok` before polling and surface the same re-auth warning as the browser leg when it is not usable.

src/Capacitor.Cli/Commands/SetupCommand.cs
realtonyyoung: Removing the stale fact is sound only with the matching server change, but the current paired server PR (#1758 at `a1d8dcd9`) still computes `DaemonOfferable` as `state.MachineFacts?.DaemonServiceEnabled is false`. This CLI now omits the field, so it folds as null and that server draws no daemon-service offer at all; the feature becomes unreachable rather than unconditional. Please update the paired server branch and make the compatible merge or release ordering explicit, or preserve the field until the server no longer gates on it.

realtonyyoung (changes requested):
Re-review of `ad6f0382`: the original raw-client and stale-profile implementations are removed, but two blockers remain. The deferred path discards a non-Ok authentication status and can still fail silently, and the current paired server PR still requires the removed field to be explicitly false before it offers the action. Details are in the existing threads. No builds or tests were run.

src/Capacitor.Cli/Commands/SetupCommand.cs
George-Payne: Right, and that was a half fix — I took the choke point and then discarded the one thing it exists to tell me. `AuthStatus` is the return value precisely because the factory hands back a client either way.

Now requires `AuthStatus.Ok` before polling, and prints the same shape of line the leg does when it is not — naming the recovery, which here is `kcap login` followed by `kcap daemon service ensure`, since the browser is gone by then and the request will not be retried. `NoAuthRequired` stays silent, matching the leg's treatment of the None provider.

src/Capacitor.Cli/Commands/SetupCommand.cs
George-Payne: The server side is already updated — you read `a1d8dcd9`, which predates it. `fdc91eaf7` on that branch removes `DaemonOfferable` entirely, so the offer no longer consults any reported field: `DaemonDetour` is now `!Finished || DaemonAction is not null`. The field is gone from the create request, the event, the fold and the view on that side too.

On making the ordering explicit — fair, nothing said it. It is:

1. This PR merges to `kcap-cli` `main`.
2. The server branch bumps the `src/cli` pin to that commit.
3. The server PR merges.

That order is forced by something other than this field, which is worth stating: the server side carries a test comparing its copy of the machine-action closed sets against the CLI's, and the two are duplicated because neither project may reference the other. It can only pass once the pin points at a CLI that has the new vocabulary, so it lands in the bump commit rather than in the server PR.

There is no window where a released pair disagrees about the removed field: it is absent on both sides, and a CLI that omits it talks to a server that never reads it.

realtonyyoung: Confirmed against paired server PR #1758 at `fdc91eaf`: the reported field and gate are removed, the daemon detour is now offered independently of machine facts, and the documented merge order keeps the pinned closed-set contract aligned. This resolves the cross-repository compatibility finding.

src/Capacitor.Cli/Commands/SetupCommand.cs
realtonyyoung: [P1] Treat `NoAuthRequired` as a usable client

The status check still leaves the deferred action unreachable on servers configured with the `None` provider. `AuthStatus.NoAuthRequired` explicitly means “the client is usable as-is”, but this branch calls `SetupDaemonService.RunAsync` only for `Ok` and silently does nothing for `NoAuthRequired`. The browser request therefore remains outstanding even though the client can poll successfully. Please run the deferred action for `Ok` **or** `NoAuthRequired`, and reserve the warning/skip path for the unusable statuses.

realtonyyoung (changes requested):
One authentication edge case remains: the deferred action silently skips a `NoAuthRequired` client even though that status is usable. I documented the detail in the existing authentication thread.

src/Capacitor.Cli/Commands/SetupCommand.cs
George-Payne: Right — fixed in d50945e1. It now runs for `Ok` or `NoAuthRequired`, and the warning is reserved for the statuses that genuinely cannot poll.

The mistake was copying the browser leg's shape without its reason. The leg skips on `NoAuthRequired` because it has nothing left to do there; the deferred action has a request waiting for an answer, so the same skip strands it on a server that would have replied.

That is the third time in this review the defect has been the same shape — a path that goes quiet instead of failing — which is worth naming given the feature's whole outcome vocabulary exists to stop the screen doing exactly that. Thank you for staying on it.

realtonyyoung: Re-reviewed at `d50945e1`. The deferred path now treats both `Ok` and `NoAuthRequired` as usable and runs the action, while the remaining statuses take the warning path. This resolves my finding.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
Re-reviewed the latest fix. No actionable findings.

PR #717, opened 2026-08-31
Stop a non-PTY hosted agent on /quit instead of forwarding it

### LINKED ISSUE #711: Chat input /quit reaches non-PTY hosted agents as a prompt instead of stopping them
## What happens

Typing `/quit` into a hosted agent's chat does nothing for Copilot (ACP) and Pi (RPC) — the agent keeps running (and may even answer the text). On PTY vendors (Claude, Codex) the same input appears to work, because the bytes reach a TUI that interprets `/quit` itself.

## Why

Nothing in the input path interprets slash commands: the app/dashboard composer sends the text verbatim, the daemon's SendInput forwards it verbatim, and ACP/RPC runtimes deliver it to the model as an ordinary prompt. `/quit` "working" is a PTY-vendor accident, not a feature — so the same gesture silently means two different things depending on vendor transport.

## Proposal (needs a product decision)

Either intercept quit-like commands (`/quit`, `/exit`) at one seam — the daemon's SendInput is the single place every surface routes through — and translate them into the existing graceful-stop path, or have the composers reject them with a hint pointing at the Stop control. Silently forwarding them to the model is the one behavior that can't be right.

realtonyyoung:
Closes #711 — AI-2380

## What & why

Typing `/quit` (or `/exit`) into a hosted agent's chat did nothing on ACP/RPC vendors — nothing in the input path interprets slash commands, so the text reached the model as an ordinary prompt (Kiro role-plays "Quitting"; Copilot answers it as a question). It only "worked" on PTY vendors by accident of their TUI. The daemon's SendInput handler — the one seam every surface routes through — now translates an exact `/quit`/`/exit` (trimmed, case-insensitive) into a stop for runtimes with no terminal, committed onto the same serial lane a server stop rides. PTY runtimes keep receiving the text verbatim: their TUI owns the command's meaning.

## Where to look

Prose that merely mentions the command (`/quit the loop and try plan B`) is deliberately still forwarded — only input that IS the command stops the agent.

## Verification

New `SendInputQuitCommandTests`: four quit variants watched fail (timeout waiting for a stop that never came), then pass — the stop funnel runs to `TerminateAsync` and the input is never delivered (activity clock unadvanced); prose-mention and PTY-verbatim pass-through pinned. Full daemon suite 2872 total, 0 failed; AOT publish clean of IL3050/IL2026.

### DIFF
--- a/src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
+++ b/src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
@@ -3499,6 +3499,15 @@ internal partial class AgentOrchestrator : IAsyncDisposable {
         }
     }
 
+    /// <summary>True only for input that IS the command — exactly <c>/quit</c> or <c>/exit</c>
+    /// after trimming — never for prose that merely mentions one.</summary>
+    internal static bool IsQuitCommand(string text) {
+        var trimmed = text.AsSpan().Trim();
+
+        return trimmed.Equals("/quit", StringComparison.OrdinalIgnoreCase)
+            || trimmed.Equals("/exit", StringComparison.OrdinalIgnoreCase);
+    }
+
     async Task HandleSendInput(SendInputCommand cmd) {
         var (agentId, text, attachmentIds) = cmd;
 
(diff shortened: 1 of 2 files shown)


realtonyyoung:
**Codex review flow — one P1, dispositioned.**

The reviewer flagged that this `/quit` path acknowledges a stop that may not have happened: `StopAgentCoreAsync` sends `Completed` and `AgentRunStopped("user", null)` before the graceful window and `TerminateAsync`, the boolean it returns is discarded on the unsequenced lane, and no caller on that lane has a failure surface.

That is accurate, and it is the shared stop path's existing behaviour — reached identically by the ordinary stop verb on `main`. This PR routes `/quit` into that path deliberately rather than forking a second shutdown, so it inherits the behaviour without introducing it. Reordering the status/event emission behind confirmed termination and giving the unsequenced lane a result-bearing path is a change to the stop protocol, not to `/quit`, so it is filed as #723 rather than folded in here.

Issue #714, opened 2026-08-31
Cursor hosted session dies silently and the UI keeps a zombie: input dropped, Stop inert


realtonyyoung:
## What the user sees

A hosted Cursor session stops reacting: `/quit` and `/exit` do nothing, and the Stop button does nothing either. The session looks alive the whole time.

## What actually happened (daemon log, cursor-agent 2026.08.11-e8db854, kcap 0.11.35)

```
12:54:54 ACP turn started for agent 4f2b0f19 (vendor=cursor)
12:55:13 ACP blocking request issued: kind=permission
12:55:25 ACP blocking request resolved: kind=permission decision=cancelled
12:55:30 ACP blocking request issued: kind=permission
12:55:33 ACP blocking request resolved: kind=permission decision=cancelled
12:55:49 ACP turn ended for agent 4f2b0f19 (vendor=cursor)
12:58:57 ACP hosted agent session ended: agentId=4f2b0f19 ...      <- no "Stopping agent", no exit-code line
13:06:26 SendInput dropped: agent 4f2b0f19 not found on this daemon (1 agents registered)
13:13:36 SendInput dropped: agent 4f2b0f19 not found on this daemon (2 agents registered)
```

The session ended daemon-side at 12:58:57 — ~3 minutes after its turn ended, with no stop requested and no exit code logged (both cancelled permission requests precede it; possibly the child exited on its own). For the next 15+ minutes the UI still offered the session: the user's inputs were dropped with agent-not-found (a warn in the daemon log, nothing user-visible), and Stop had nothing to act on.

## Three questions to answer

1. What ended the session at 12:58:57, and why does that path log neither a stop nor an exit code?
2. Why didn't the end propagate — the surface kept rendering a live, composable session after the daemon lost the agent?
3. Why is a dropped SendInput silent to the sender? "Agent not found" is precisely the answer the user needed at 13:06.

realtonyyoung:
[...]

PR #733, opened 2026-08-31
Answer a refused permission with the agent's reject option

### LINKED ISSUE #732: A user's Deny is sent to the agent as a cancelled turn, not a refused tool call
## What happens

`AcpInteractionBridge.MapPermissionDecision` produces `selected` only for a recognized affirmative outcome; everything else, a user's **Deny** included, maps to ACP `cancelled`.

Those are different statements. `cancelled` tells the agent the prompt turn went away; a `selected` reject option tells it *this tool call was refused* and the turn continues. An agent that reads two cancelled permission requests as its turn being pulled out from under it may well stop rather than carry on — and in the incident behind #714, a cursor-agent session ended silently about three minutes after exactly that pair of cancelled permission requests.

## Fix

Answer a refusal with the agent's own reject option (`reject_once` preferred over `reject_always`), and keep `cancelled` for the case where the agent offered no reject option at all — with no way to say no, saying nothing is honest.

The fail-safe property has to survive: a refusal must never be able to select an allow, whatever option id arrives with it, and an ambiguous reject set must not be guessed at — a standing refusal is exactly where a wrong pick persists past the call.

Found while root-causing #714.

realtonyyoung:
Closes #732 — AI-2383 (part)

## What & why

A user's Deny reached the agent as ACP `cancelled`. That says the prompt turn went away, not that this call was refused, and an agent can reasonably respond to the first by stopping — the cursor session in #714 ended silently three minutes after a pair of cancelled permission requests. A refusal now selects the agent's own reject option, and `cancelled` is kept for the case where no reject option was offered.

## Where to look

The fail-safe property is the part worth checking. A negative outcome resolves only among reject-kind options, so an allow id arriving on a refusal resolves to nothing and falls back to `cancelled`; `reject_once` is preferred over `reject_always`, and two of the same kind with no explicit choice is left ambiguous rather than guessed — a standing refusal is where a wrong pick outlives the call.

## Verification

Five new pins in `AcpInteractionBridgeTests`: the reject option is selected, no reject offered gives `cancelled`, a deny naming an allow id gives `cancelled`, two `reject_always` with no choice gives `cancelled`, and `reject_once` wins over `reject_always`. Daemon unit suite 2898 total, 0 failed. AOT publish of the daemon clean of IL3050/IL2026.

### DIFF
--- a/src/Capacitor.Cli.Daemon/Acp/AcpInteractionBridge.cs
+++ b/src/Capacitor.Cli.Daemon/Acp/AcpInteractionBridge.cs
@@ -557,52 +557,17 @@ internal sealed partial class AcpInteractionBridge(
     partial void LogUnattendedPermissionFrameForbidden(string agentId, string toolTitle, string toolKind);
 
     /// <summary>
-    /// Maps a resolved <see cref="AcpInteractionDecision"/> to the ACP outcome result shape.
+    /// Maps a resolved <see cref="AcpInteractionDecision"/> to the ACP outcome result shape, by two
+    /// allowlists and no fallthrough: an outcome on neither list is <c>cancelled</c>, so a typo, a
+    /// corrupted decision, or a vocabulary the daemon has not been taught cannot grant or refuse
+    /// anything. Both lists are literal copies of the server's canonical values — the daemon does not
+    /// reference the server-only <c>InterruptOutcomes</c> type, per the AOT/trim constraints.
     ///
-    /// <b>Spec-review Finding 2 (security-critical, fail-safe-by-default):</b> this method uses
-    /// an EXPLICIT ALLOWLIST of recognized affirmative outcome strings (<c>"allow"</c>/
-    /// <c>"allow_once"</c>/<c>"allow_always"</c>/<c>"answered"</c> — the daemon does not reference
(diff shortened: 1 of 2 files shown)


realtonyyoung:
**Codex review flow — one P1, taken, plus the question it could not answer.**

**Explicit-id path skipped the addressability checks.** Correct: it matched by id and returned, so a blank or duplicated id produced a `selected` that names no single option — and with two rejects sharing an id, the one returned depends on array order, which can turn a `reject_once` refusal into a `reject_always`. Both paths now go through one check, and it looks across *all* offered options rather than just the rejects: an id shared with an allow is unaddressable for the same reason and the wrong resolution there grants what was refused. Pinned by two new tests.

**The vocabulary question I flagged.** Checked against the server: `InterruptOutcomes` is `allow`/`allow_once`/`allow_always`/`deny`/`cancel`/`answered`/`timeout`. `deny` is the only negative outcome it sends, so the set is now that one entry — the same shape `AffirmativeOutcomes` already has, mirroring the canonical list rather than inventing wire values. `cancel` and `timeout` stay out deliberately: those really are the turn going away.

Daemon unit suite 2900 total, 0 failed; AOT clean.

PR #767, opened 2026-09-04
Beat while the browser waits, so a silent machine is visible

George-Payne:
AI-2483 — no GitHub issue; found in-house and filed in Linear only.

## What & why

The browser is told this machine has gone by a relinquish the leg sends on its way out. That is a statement, so anything that stops the process making statements sends nothing at all — a kill, a lost network, a shut lid — and the setup page then holds its step until the flow's twelve-hour lifetime ends it.

A beat on its own timer turns that silence into something the server can observe. It runs from a `PeriodicTimer` on a detached task, not from the poll loop: the loop blocks for the whole of an import and hands the elapsed time back to its own deadline, so liveness derived from polling would call the machine gone during the one stretch it is working hardest.

## Where to look

Nothing inspects the result. A beat is a network call on the machine whose network may be exactly what is failing, so a status code is never read and a throw does not end the loop — a run of failures is the signal, and only the server can read it.

Stopping does not wait for a beat in flight — the cancel aborts it. Nothing is owed to it: the relinquish that follows states the ending, and the browser reads a stated ending ahead of an inferred one.

The beat says the process is alive, never that its work is progressing — a wedged leg goes on beating. The server half, which decides what silence renders as, is a separate PR.

## Verification

`Capacitor.Cli.Core.Tests.Unit` 2960 and `Capacitor.Cli.Tests.Unit` 3984, 0 failed. `dotnet publish -c Release` clean of IL2026/IL3050.

`The_beat_starts_outside_the_poll_it_has_to_outlive` pins the beat's placement rather than observing beats during an import: that observation needs a continuation to run while the poll thread is parked inside a synchronous hook, which a fake clock cannot arrange and wall time only wins on an idle runner.


test/Capacitor.Cli.Core.Tests.Unit/FirstRun/BrowserFirstRunFlowTests.cs
```diff
@@ -1801,6 +1815,49 @@ public async Task Neither_lane_spends_the_backstop_it_was_not_waiting_through()
         await Assert.That(result).IsTypeOf<FirstRunFlowResult.Finished>();
     }
 
+    // ---- saying the machine is still here ----
```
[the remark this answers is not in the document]
George-Payne: Keeping this one. It is a navigational divider, not an explanatory comment, and it is the file's existing convention: `// ---- saying the machine has gone ----` and `// ---- the interrupt path ----` already sit below it in the same class. Removing only the divider this PR adds would leave the file inconsistent, and the two it matches are not part of this change.

The first finding was right and is fixed: the `OperationCanceledException` catch restated its own type, and the class doc already carries the reason a cancel is swallowed here rather than propagated.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -131,6 +131,10 @@ public async Task<FirstRunFlowResult> RunAsync(
             (reason, token) => RelinquishAsync(serverUrl, flowId, reason, token),
             interruptReason: () => InterruptReason(Volatile.Read(ref settled)));
 
+        // Says this machine is still here on its own timer, so a death that sends nothing — a kill, a
+        // lost network, a shut lid — is still something the browser can see. Disposed with the leg.
+        using var beat = FirstRunHeartbeat.Start(channel, serverUrl, flowId, _clock);
```
rsergeenko, line 132: **The beat can outlive the relinquish.**

`using var beat` sits above the `try`, so it is still live during `await notice.SendAsync(ReasonFor(settled!), ct)` at the end of `RunAsync`: the relinquish POST and the next 5s tick race, and the server can process the beat last.

The `Environment.Exit` path is worse. `InteractiveLifetime.Interrupt` and the `ProcessExit` handler in `Program.cs` call `FirstRunInterruptRelinquish.RunBeforeExit(ExitNoticeBudget)`, and — as that handler's own comment says — `Environment.Exit` runs no `finally` and no `using` disposal. So nothing stops the timer: it keeps firing for the whole notice budget *after* the relinquish is sent.

Failure scenario: the user Ctrl+Cs 4.9s after the last beat. The relinquish lands, a beat lands ~100ms later, and a server that reads a beat as presence flips the setup page from "this machine has stopped" back to waiting — exactly the stuck state this PR removes.

Suggested: `beat.Dispose()` before `notice.SendAsync`, and make "a relinquish is terminal, a later beat cannot resurrect the flow" an explicit part of the contract. The in-flight case cannot be closed client-side.

rsergeenko, line 132: **A `Finished` flow goes silent while this machine is still acting on the browser's request.**

`ReasonFor(Finished)` is `null` by design, and the beat stops the moment `RunAsync` returns — but `SetupCommand` then runs `SetupDaemonService.RunAsync` under the *same* `flowId` (it re-polls precisely because the browser can still be open) and can spend tens of seconds in `launchctl`/`systemctl`.

Silence used to mean nothing; now it is the signal. So: the user finishes in the browser, presses the daemon-service button, and the terminal installs the service while the page renders the machine as gone.

Either the beat needs to cover the deferred leg, or `Finished` needs a stated ending.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,98 @@
...
+    static async Task BeatAsync(
+            IFirstRunFlowChannel channel, string serverUrl, string flowId, TimeProvider clock,
+            TimeSpan interval, CancellationTokenSource stopping) {
+        var ct = stopping.Token;
+
+        using var timer = new PeriodicTimer(interval, clock);
+
+        try {
+            while (!ct.IsCancellationRequested) {
+                await SendOneAsync(channel, serverUrl, flowId, ct);
```
rsergeenko: **One slow beat silences the machine for the whole HTTP timeout.**

`SendOneAsync` is awaited before `timer.WaitForNextTickAsync`, and this client's timeout is `SetupCommand.BrowserFlowHttpTimeout` = 15s against a 5s interval. A single black-holed connection — laptop wake, NAT rebind, i.e. the conditions this feature targets — produces a ≥15s server-visible gap, three consecutive missed beats, which contradicts `Interval`'s own claim that a single dropped beat is not a verdict.

Suggested: bound each beat with a linked CTS at or below `interval`.

src/Capacitor.Cli.Core/FirstRun/FirstRunFlowClient.cs
```diff
@@ -209,6 +219,22 @@ public async Task<FirstRunRelinquishOutcome> RelinquishAsync(
...
 
+    /// <inheritdoc/>
+    public async Task<FirstRunHeartbeatOutcome> HeartbeatAsync(
+            string serverUrl, string flowId, CancellationToken ct) {
+        try {
+            using var req = new HttpRequestMessage(
+                HttpMethod.Post,
+                $"{Base(serverUrl)}/api/first-run/flows/{Uri.EscapeDataString(flowId)}/heartbeat");
+
+            using var resp = await http.SendAsync(req, ct);
```
rsergeenko, line 234: **The heartbeat is the only route that ignores 429 / `Retry-After`.**

Every other method here either surfaces `RetryAfter(resp)` or is driven by a loop that backs off. The beat reads nothing and keeps posting every 5s for up to the 30-minute poll budget (~360 requests). On a rate-limited tenant that unbacked-off traffic can keep the *poll* route in penalty, degrading the interactive wait the beat exists to protect.

"Never inspected for success" is a defensible design choice; ignoring an explicit throttling instruction is a different one.

test/Capacitor.Cli.Core.Tests.Unit/FirstRun/FirstRunHeartbeatTests.cs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    [Test]
+    public async Task Disposing_it_stops_the_beats() {
+        var channel = new FakeChannel();
+        var clock   = new FakeTimeProvider();
+
+        var beat = FirstRunHeartbeat.Start(channel, Server, Flow, clock, Beat);
+
+        await ReachesAsync(channel, clock, 2);
```
rsergeenko: **This test can pass vacuously.**

`await ReachesAsync(channel, clock, 2);` discards its `bool`. If beats never start — the regression this class guards — `settled` is 0 or 1 and the final `IsEqualTo(settled)` still holds, so the test stays green on a heartbeat that never beat.

Assert the result, as the two neighbouring tests do:

```csharp
await Assert.That(await ReachesAsync(channel, clock, 2)).IsTrue();
```

test/Capacitor.Cli.Core.Tests.Unit/FirstRun/BrowserFirstRunFlowTests.cs
```diff
@@ -1801,6 +1816,54 @@ public async Task Neither_lane_spends_the_backstop_it_was_not_waiting_through()
...
+    /// <para><b>Pinned structurally rather than by observing beats during an import.</b> That test needs
+    /// the beat's continuation to run while the poll thread is parked inside the lane's synchronous
+    /// hook, and a fake clock cannot make it: missed ticks collapse into one catch-up on resume, so the
+    /// count does not follow the advances. Driving it by wall time instead is a race that a loaded
+    /// runner loses. The regression this guards against is someone moving the beat inside the loop,
+    /// and where it starts is exactly what says whether they did.</para>
+    /// </summary>
+    [Test]
+    public async Task The_beat_starts_outside_the_poll_it_has_to_outlive() {
+        var source = await File.ReadAllTextAsync(FlowSourcePath());
```
rsergeenko: **This placement test depends on the source tree being present at the compile-time path.**

`FlowSourcePath()` uses `[CallerFilePath]` and walks up for `Capacitor.slnx`; run from a published test bundle or a container stage carrying only binaries, `File.ReadAllTextAsync` throws and the test fails for reasons unrelated to the invariant.

It also only matches text: starting the beat before `PollAsync` but disposing it early would still pass. Worth pinning the behaviour instead — a beat observed while the poll is blocked — or at least degrading to inconclusive when the source is absent.

rsergeenko (commented):
Code review of the first-run heartbeat. Six findings, all verified against this head commit; the first two reintroduce the false state the feature exists to remove, the rest are smaller.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
George-Payne: Agreed, and fixed for the scheduled case: `beat.Dispose()` now runs in the `finally` before `notice.SendAsync`. It is synchronous, so it costs the leg's exit nothing.

You are right that the `Environment.Exit` path cannot be closed from here — no `using` runs at all — and neither can a beat already in flight. So the ordering is not the guarantee; the guarantee has to be that a relinquish is terminal regardless of what arrives after it. That is how the server behaves (the relinquished arm is rendered ahead of any liveness state, and silence is never allowed to reopen a flow), but nothing pinned it. It does now: kurrent-io/kcap-server#1822 asserts a relinquished flow keeps its own tail with the tracker reporting `Listening`.

George-Payne: The premise is right — `SetupDaemonService.RunAsync` reuses `browserFlowId` after `RunAsync` has returned and the beat has stopped, and it can spend a long time in `launchctl`/`systemctl`.

The symptom does not reach the browser, though, because of the server half (kurrent-io/kcap-server#1819, not visible from this PR). Silence is suppressed whenever the flow already has an ending of its own: `Finished` is covered by an `IsFinished()` check, and `Dismissed`/`Abandoned`/`Failed` all send a relinquish first, which is terminal. Those are the only outcomes that reach the deferred leg, since it runs exactly when `browserAnswers.FlowId` is set.

The finished case had a real hole of a different shape — nothing publishes when a flow finishes, so a second tab's fold reads Active for ever — and that is fixed there by re-folding from the store before claiming lost contact, with a test.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Fixed. Each beat now runs under a linked source bounded by the interval, so the client's 15s timeout can no longer decide how long the machine looks absent. `A_beat_is_bounded_by_the_interval` hangs a beat forever and asserts both that the loop keeps beating and that the hung request was actually cancelled rather than abandoned.

src/Capacitor.Cli.Core/FirstRun/FirstRunFlowClient.cs
George-Payne: Agreed — "never inspected for success" and "ignores an explicit throttle" are different choices, and I had conflated them.

`FirstRunHeartbeatOutcome` now carries `RetryAfter`, populated on a 429 the way the other routes do it, and the loop goes quiet for the delay the server named (60s if it named none). Success is still not inspected. `A_throttled_beat_goes_quiet_for_as_long_as_it_was_told` pins both halves: silent through the window, beating again after it.

test/Capacitor.Cli.Core.Tests.Unit/FirstRun/FirstRunHeartbeatTests.cs
George-Payne: Correct, and fixed — it now asserts the result. Thank you for catching it; this is the second vacuous assertion in this change, so I have stopped trusting a passing negative test and started reverting the fix to watch it fail before believing it.

test/Capacitor.Cli.Core.Tests.Unit/FirstRun/BrowserFirstRunFlowTests.cs
George-Payne: Both objections stand, and the second one is the decisive one: starting the beat early and disposing it early would have passed.

So it is gone, replaced by the behavioural test you suggested. The reason I went structural was that the earlier behavioural version blocked the poll loop's thread with `Thread.Sleep` while waiting on a continuation elsewhere, which is flaky by construction. The fix is for the lane to park in an `await` instead: `Drive` keeps pumping the fake clock while the poll loop is suspended inside the lane, so beats accumulate deterministically with no wall clock. Verified it bites by adding an early `beat.Dispose()` and watching it fail.

On the pattern itself, for the record rather than as an argument: source-scanning guards are established here — `TestSourceScan` in the server's TestHelpers, plus `SetupJoinMintPlacementGuardTests` and `ReviewerModelVendorNeutralityGuardTests` in this repo. Worth knowing that the same source-tree dependency applies to those, since `TestSourceScan.RepoRoot()` throws if it cannot find the root.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,120 @@
...
+    /// failure, and the poll shares this tenant's budget.</para>
+    ///
+    /// <para>Swallows everything, including a cancel: this runs on a detached task, so an escaping
+    /// exception has no caller to reach. The loop reads the token itself.</para>
+    /// </summary>
+    static async Task<DateTimeOffset> SendOneAsync(
+            IFirstRunFlowChannel channel, string serverUrl, string flowId, TimeProvider clock,
+            TimeSpan interval, CancellationToken ct) {
+        try {
+            using var bound   = new CancellationTokenSource(interval, clock);
```
rsergeenko: **The per-beat bound now cancels a credential rotation.** (New — introduced by this fix.)

The beat shares the setup client, built at `SetupCommand.cs:1477` with `autoRetryUnauthorized: true`, so `UnauthorizedRetryHandler` sits in its pipeline. A beat has `Content == null`, so `CanResend` is true, and a 401'd beat reaches `RecoverForServerAsync(profile, url, applied.AccessToken, cancellationToken)` — with the 5s-bounded token.

Inside `RefreshWithCrossProcessLockAsync` that token is honoured throughout: the lock-acquire loop alone allows **15s** (`ThrowIfCancellationRequested` + `Task.Delay(100, cancellationToken)`), and `SaveAsync(profile, refreshed, cancellationToken)` passes it straight into `writer.WriteAsync(..., ct)`. Meanwhile the WorkOS rotate POST itself is *not* cancellable — the delegate is the ct-less `RefreshWorkOSAsync` overload — which makes the bad outcome more likely rather than less.

Scenario: the token is invalidated mid-setup, the beat sees the 401 first, wins the profile lock, WorkOS rotates the refresh token, and the 5s bound trips before the persist completes. The refresh token is spent server-side and never written to disk: the user is logged out mid-setup and has to re-run `kcap login`.

Before this PR every request on that client was serial poll traffic bounded only by the 15s `HttpClient` timeout — which matches the refresh lock's own deadline. The beat is the first sub-deadline canceller on that path.

Either keep the beat off the 401-refresh path (its own client without the handler, or a request whose 401 is simply ignored) or don't hard-cancel a beat that is inside a refresh.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -131,9 +131,22 @@ public async Task<FirstRunFlowResult> RunAsync(
...
             Volatile.Write(ref settled, await PollAsync(serverUrl, flowId, report, ct));
         } finally {
+            // Stopped before the notice below, not left to the `using`: a beat scheduled after the
+            // relinquish says this machine is here, having just said it had gone. Synchronous, so it
+            // costs the leg's exit nothing.
+            //
+            // Closes the SCHEDULED beat only. One already in flight, and every beat on the
+            // Environment.Exit path — which runs no `using` at all — can still arrive late, so a
+            // relinquish has to be terminal on the server rather than merely first.
+            beat.Dispose();
```
rsergeenko, line 159: **Still open from the previous round: liveness goes silent while the same `flowId` is still being worked on.**

`ReasonFor(Finished)` states nothing, and the beat is now stopped here — but `SetupCommand.cs:919` then runs `SetupDaemonService.RunAsync` against that same `browserFlowId`: a fresh authenticated client, a poll, `DaemonServiceCommands.FlowEnsureAsync` spawning systemd/launchd children, and the report retried up to three times.

The browser is holding the outstanding daemon-service request for all of it, and at a 5s interval any staleness window the server picks will be crossed by a slow `systemctl`/`launchctl` — so the page renders the machine as silent exactly while it is installing what the page asked for.

The beat needs to span the deferred step: hoisted to the caller, or a second `Start` around `SetupDaemonService.RunAsync`.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -131,9 +131,22 @@ public async Task<FirstRunFlowResult> RunAsync(
             (reason, token) => RelinquishAsync(serverUrl, flowId, reason, token),
             interruptReason: () => InterruptReason(Volatile.Read(ref settled)));
 
+        // Says this machine is still here on its own timer, so a death that sends nothing — a kill, a
+        // lost network, a shut lid — is still something the browser can see.
+        using var beat = FirstRunHeartbeat.Start(channel, serverUrl, flowId, _clock);
```
rsergeenko, line 132: **The stop-before-relinquish ordering covers only the graceful leg.**

Acknowledging the comment below — but worth being explicit that the uncovered path is the *common* one. On Ctrl+C/SIGTERM/watchdog, `InteractiveLifetime.Interrupt` calls `FirstRunInterruptRelinquish.RunBeforeExit(1s)` from the signal-handler thread while the leg is still parked in `PollAsync`. Nothing disposes `beat`, so its 5s `PeriodicTimer` can issue a fresh beat inside that 1s budget (roughly one interrupt in five) or after the relinquish POST returns — on top of the in-flight case.

As written, correctness there rests on a server property that lives in another repo ("a relinquish has to be terminal on the server"), so this PR cannot demonstrate it. A holder captured by the `send` callback would let the notice stop the beat before it posts, making the ordering hold on both paths without that dependency.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,120 @@
...
+            using var either  = CancellationTokenSource.CreateLinkedTokenSource(ct, bound.Token);
+
+            var outcome = await channel.HeartbeatAsync(serverUrl, flowId, either.Token);
+
+            return outcome.StatusCode is 429
+                ? clock.GetUtcNow() + (outcome.RetryAfter ?? ThrottleBackoff)
```
rsergeenko: **`RetryAfter` is honoured unclamped.**

A `429` naming `Retry-After: 3600` silences liveness for longer than the leg's entire 30-minute budget. The doc comment's justification — a throttled tenant reading as silent is honest — assumes the whole tenant is throttled, but the heartbeat is its own route with its own limiter: the poll can keep succeeding every 2s while the page tells the user their machine has gone.

Clamping the quiet window to a few intervals would keep the two halves consistent, the way `Backoff` clamps the poll to `MaxInterval`.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
rsergeenko: **The bound equals the interval, so a beat that needs more than one interval can never land at all.**

On a link where the POST legitimately takes >5s — cold TLS after a wake, a tethered connection — every beat is cancelled just before it would have succeeded, and the machine reads as permanently gone while the poll, on the same 15s client timeout, works fine.

A bound of two intervals still caps the multi-beat gap the previous round was about without making success impossible.

src/Capacitor.Cli.Core/FirstRun/FirstRunFlowClient.cs
```diff
@@ -209,6 +223,24 @@ public async Task<FirstRunRelinquishOutcome> RelinquishAsync(
...
+
+            var status = (int)resp.StatusCode;
+
+            return new(status, status is 429 ? RetryAfter(resp) : null);
```
rsergeenko, line 238: **No status ends the loop.**

Against a server whose `/heartbeat` route is not deployed yet — CLI and server versions move independently — every beat 404s and the loop still posts one every 5s for up to 30 minutes: ~360 authenticated no-op POSTs per setup run, which can itself trip the tenant limiter that the new 429 handling exists to protect the poll from.

The create path already treats 404 as its availability oracle; the beat could stop on 404/405 for the same reason.

rsergeenko (commented):
Re-review after `4a40eaa`.

**Resolved from the previous round:** the serial-beat gap (per-beat CTS), 429/`Retry-After` (the `quietUntil` gate is correct — the loop keeps waking on its tick and only skips the send), the vacuously-passing `Disposing_it_stops_the_beats`, and the `[CallerFilePath]` source-scraping test, now replaced by `The_beat_goes_on_while_the_import_holds_the_poll`. `beat.Dispose()` before the relinquish fixes the graceful leg. Suite is green (8/8 and 103/103).

**Still open:** the `Finished`/daemon-service gap, and the interrupt path. **New:** the per-beat bound is now applied to a request that can rotate a credential — that one I'd treat as blocking, since its blast radius (a logged-out user) is worse than the gap it closes.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Confirmed and fixed — and thank you, this was the right call to block on. I verified every link: the delegate handed to `RefreshWithCrossProcessLockAsync` is the ct-less `RefreshWorkOSAsync` overload, the lock loop calls `ThrowIfCancellationRequested` against its 15s deadline, and line 576 is `await SaveAsync(profile, refreshed, cancellationToken)`. So the rotation completes server-side and the persist is what gets cancelled. My round-1 fix put the first sub-deadline canceller on that path, exactly as you say.

The fix is to stop cancelling rather than to pick a longer bound: the loop now `Task.WhenAny`s the send against a delay and takes the next tick if the send is still running, leaving the request — and any refresh inside it — alone to finish on the client's own timeout. No second client needed.

That also answers your separate point about the bound equalling the interval: a beat that legitimately needs longer than 5s now lands rather than being cancelled just before it would have succeeded. The cost is an occasional overlapping beat, which is an idempotent no-op POST.

`A_hung_beat_holds_neither_the_loop_nor_a_cancel` pins both halves, and I checked it fails when the linked source is put back.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
George-Payne: Fixed rather than argued this time. The beat now spans the deferred leg: `SetupCommand` starts one around `SetupDaemonService.RunAsync` on the same flow.

Gated on a new `BrowserFlowAnswers.FlowStillLive`, true only for `Finished`. Writing it turned up something I had missed — the deferred leg runs for `Abandoned` and `Dismissed` too, and both of those already relinquished, so an unconditional beat there would have said this machine was back moments after it said it had gone. That is worse than the silence it would be covering.

I still think the symptom was masked server-side (finished and relinquished both suppress the silent tail, and those are the only endings that reach this code), but you are right that a liveness signal should not depend on another repo hiding its gaps.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
George-Payne: Taken, and implemented as you suggested. The beat is now created before the notice and disposed *inside* the send callback, so whichever path sends — the leg's own or the signal handler's — stops the timer before it posts. The ordering no longer rests on a property this repo cannot demonstrate.

What remains uncovered is only a request already in flight when the relinquish is sent, which cannot be closed from here without cancelling it — and cancelling is precisely what your other finding rules out. Server-side terminality is pinned separately in kurrent-io/kcap-server#1822.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Agreed — the doc comment's justification assumed a tenant-wide throttle, and the heartbeat has its own limiter, so it was wrong. Clamped to two minutes, with the reasoning corrected on the constant. `A_throttle_longer_than_the_leg_is_clamped` drives an hour-long `Retry-After` and asserts the beat resumes anyway.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Fixed by the same change as the rotation finding: the loop stops waiting rather than cancelling, so a beat that needs longer than one interval completes and lands. See the reply on that thread.

src/Capacitor.Cli.Core/FirstRun/FirstRunFlowClient.cs
George-Payne: Agreed, and done — 404 and 405 now end the loop, on the same reasoning the create path uses for its availability oracle. `A_missing_route_stops_the_beat` covers both statuses and I confirmed it fails without the check.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,158 @@
...
+    /// that legitimately needs longer than one interval — cold TLS after a wake, a tethered link — still
+    /// land, instead of being cancelled just before it would have succeeded.</para>
+    ///
+    /// <para>Success is not inspected. A throttle and an absent route are, because both say something no
+    /// later beat can discover for itself.</para>
+    /// </summary>
+    static async Task<BeatVerdict> SendOneAsync(
+            IFirstRunFlowChannel channel, string serverUrl, string flowId, TimeProvider clock,
+            TimeSpan interval, CancellationToken ct) {
+        var beat = Observed(channel.HeartbeatAsync(serverUrl, flowId, ct));
```
rsergeenko: **The beat is still issued with the stop token, so the credential loss is unchanged.**

`ct` here is `_stopping.Token`, and `Dispose()` still calls `_stopping.Cancel()`. So the hazard this method's own doc spells out is still live, verbatim: a beat that 401s enters `UnauthorizedRetryHandler.SendAsync` → `TokenStore.RecoverForServerAsync` → `RefreshWithCrossProcessLockAsync`, WorkOS spends the single-use refresh token (uncancellable), and the cancel then lands on `SaveAsync(profile, refreshed, cancellationToken)` → `writer.WriteAsync(..., ct)`, so the replacement is never persisted and the user is logged out mid-setup.

`docs/CHANGES.md` asserts "A beat is never cancelled" and then, two paragraphs later, "the cancel aborts it". The code matches the second.

The interrupt path is worse: `BrowserFirstRunFlow.cs:140` disposes the beat *first*, and the relinquish POST that follows on the same client then 401s with an already-spent refresh token and cannot recover either.

Fix is the one token: pass `CancellationToken.None` to `HeartbeatAsync`. The `Task.Delay` bound below and the client's own timeout are what should end a beat — which is exactly what the doc comment describes.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        var waited = await Task.WhenAny(beat, Task.Delay(interval, clock, ct));
+
+        // Still in flight: leave it running and take the next tick. It will finish or time out on the
+        // client's own deadline, and a beat that lands late is still a beat.
+        if (waited != beat) return BeatVerdict.Continue;
```
rsergeenko: **An abandoned beat's verdict is discarded, so a slow 429 or 404/405 is never acted on.**

When the delay wins, the task is dropped and nobody ever reads its outcome. A throttling or overloaded server is precisely the one whose responses take longer than one interval, so the loop never learns the `Retry-After` it now clamps above and keeps posting every 5s — which defeats the throttle handling this class documents as its reason to inspect a status at all.

Same for the availability oracle: a slow `404` — cold DNS + TLS on the very first beat is the likely case — is dropped, and the loop beats on for the rest of the leg against a route that does not exist.

Holding the pending task across ticks and reading its verdict when it completes fixes this and the overlap below.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,158 @@
...
+    ///
+    /// <para>Success is not inspected. A throttle and an absent route are, because both say something no
+    /// later beat can discover for itself.</para>
+    /// </summary>
+    static async Task<BeatVerdict> SendOneAsync(
+            IFirstRunFlowChannel channel, string serverUrl, string flowId, TimeProvider clock,
+            TimeSpan interval, CancellationToken ct) {
+        var beat = Observed(channel.HeartbeatAsync(serverUrl, flowId, ct));
+
+        var waited = await Task.WhenAny(beat, Task.Delay(interval, clock, ct));
```
rsergeenko: **Abandoned beats accumulate, up to `HttpClient.Timeout / interval` concurrent POSTs.**

`WaitForNextTickAsync` returns almost immediately after a 5s abandoned wait — the period has already elapsed — so a wedged network gets a fresh beat every 5s with every earlier one still open.

The browser leg caps that at 3 (`SetupCommand.BrowserFlowHttpTimeout`, 15s). The deferred client from `CreateClientWithAuthStatusAsync` sets no `Timeout`, and `FirstRunFlowClient.HeartbeatAsync` posts through `http.SendAsync` directly rather than the retry helper with its `DefaultTimeout` — so there it is the 100s default: ~20 simultaneous authenticated POSTs, on the machine whose network is the failing thing, sharing a connection pool with `SetupDaemonService.RunAsync` running concurrently.

Skipping a new beat while one is still pending is the cheap mitigation, and it is the same change that fixes the discarded verdict.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -915,9 +915,20 @@ await ConfigMutator.MutateAsync(config, c => {
...
                 // answered — the browser leg skips there only because it has nothing left to do.
-                if (deferredAuth is AuthStatus.Ok or AuthStatus.NoAuthRequired)
+                if (deferredAuth is AuthStatus.Ok or AuthStatus.NoAuthRequired) {
+                    var channel = new FirstRunFlowClient(deferred);
+
+                    // The browser is holding the request this leg is about to perform, and enabling a
+                    // service can outlast any staleness window — so the beat has to span it. Only for a
+                    // flow that finished: every other ending already relinquished, and beating again
+                    // would take that back.
+                    using var beat = browserAnswers.FlowStillLive
```
rsergeenko, line 930: **The gate is right, but the beat does not span the prompts before this point.**

The browser leg's beat dies when `BrowserFirstRunFlow.RunAsync` returns, and this one starts here. In between sit the visibility `SelectionPrompt` (line 593), the confirmation at 630, the API-key `ConfirmationPrompt` (848), and — unconditionally, since the browser leg only runs when `!noPrompt` — the daemon-name `TextPrompt` (865).

A user who steps away at any of those, after a `Finished` flow whose daemon-service request the browser is still holding, produces exactly the silence this PR exists to remove; then this line revives the machine. That flap is what `BrowserFlowAnswers.FlowStillLive`'s own doc calls "worse than the silence it would be covering".

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -127,8 +127,20 @@ public async Task<FirstRunFlowResult> RunAsync(
...
+
+        // Stopped by whichever path sends, INSIDE the send rather than around it — which is what makes
+        // the ordering hold on the interrupt path too. There, Environment.Exit runs no `using` at all,
+        // so a beat scheduled by the timer could otherwise post inside the notice's own budget and tell
+        // the server this machine is here, moments after it said it had gone.
         using var notice = _interrupts.Arm(
-            (reason, token) => RelinquishAsync(serverUrl, flowId, reason, token),
+            (reason, token) => {
+                beat.Dispose();
```
rsergeenko, line 144: **Inside the send callback means it never runs when there is nothing to send.**

`FirstRunNotice.RunAsync` invokes `send` only `if (reason is not null)`, and `InterruptReason` evaluates to null for `Finished` and `Expired`. So an interrupt on one of those claims the notice, sends nothing, and never disposes the beat — which then keeps scheduling through the notice budget and the telemetry flush that follows in `Program.cs` (`FlushAndClose().GetAwaiter().GetResult()` can take seconds).

No relinquish is contradicted on that branch, so the harm is small. But the comment above claims the ordering is stopped "by whichever path sends", and this is the path that sends nothing — worth either disposing before the `reason is not null` check or narrowing the comment.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,158 @@
...
+                    // handling above exists to keep clear. Same oracle the create uses.
+                    if (verdict.Unavailable) return;
+
+                    quietUntil = verdict.QuietUntil;
+                }
+
+                if (!await timer.WaitForNextTickAsync(ct)) return;
+            }
+        } catch (OperationCanceledException) {
+        } finally {
```
rsergeenko: **"The loop ends only on the cancel" is no longer true, so this `Dispose` can race the stop.**

The 404/405 `return` above and a `false` tick both end the loop with `_stopped` still 0, after which this `finally` disposes the source. A later `Dispose()` then calls `Cancel()` on a disposed CTS — handled sequentially by the `ObjectDisposedException` catch, but `CancellationTokenSource.Dispose` is the one member documented as not thread-safe against the others, and the concurrent overlap (this `finally` running while the leg's `using` disposes) is not covered by that catch.

Both comments — this one and the guard's at lines 61-64 — now assert an invariant the early return broke, so they need correcting either way.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
rsergeenko: **`channel.HeartbeatAsync(...)` is evaluated outside `Observed`, so a synchronous throw kills the loop.**

The call is an argument, so an exception thrown *before* a task is returned bypasses `Observed`, propagates out of `SendOneAsync`, and is not caught by `BeatAsync`'s `catch (OperationCanceledException)` — the beat stops for the rest of the leg and `_beating` faults.

`FirstRunFlowClient.HeartbeatAsync` is `async` so it cannot happen today, but `IFirstRunFlowChannel` is public and the repo already has non-`async` implementations of this member (the test fakes), and `A_channel_that_throws_does_not_stop_the_beat` pins only the `async` case. Moving the invocation inside `Observed` closes it.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        try {
+            while (!ct.IsCancellationRequested) {
+                if (clock.GetUtcNow() >= quietUntil) {
+                    var verdict = await SendOneAsync(channel, serverUrl, flowId, clock, interval, ct);
+
+                    // A route this server does not have answers the same way for the rest of the leg, so
+                    // beating on is ~360 authenticated no-ops that can trip the very limiter the throttle
+                    // handling above exists to keep clear. Same oracle the create uses.
+                    if (verdict.Unavailable) return;
```
rsergeenko: **One 404/405 silences liveness for the rest of the leg, with no recovery — and it is not the create's oracle.**

The create already returns `Unavailable` for 404/405, so the only way to observe one here is transient or path-specific: a gateway 405 during a rolling deploy, a proxy rejecting POST on an unknown path. The beat then stops forever while the poll keeps succeeding every 2s on the same client — the browser infers the machine is gone from a machine that is demonstrably still talking to it.

"Same oracle the create uses" is also inaccurate: the create's set is 404/401/403/405, and it abandons the whole leg rather than one route. Requiring two consecutive unavailable answers (or a capped retry) keeps the ~360-no-op protection without making a blip terminal.

rsergeenko (commented):
Third pass, after `a5213f10`.

**Resolved:** the interval-sized bound (the patience now sits on the loop, so a beat that legitimately needs longer can land), the unclamped `Retry-After` (`MaxThrottleBackoff`), the 404/405 stop, and — for the graceful and most interrupt paths — the dispose ordering. The deferred daemon-service leg now gets its own beat, correctly gated on `FlowStillLive`.

**Still reachable:** the credential loss from round 2. The new doc comment states the rule exactly right — "This must not hard-cancel, and that is the load-bearing rule here" — but the call site still hands the request the stop token, so nothing changed behaviourally on that path. That one comment is the blocker; the rest are consequences of moving the bound, plus two comments that now assert invariants the new early return broke.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: You are right, and I need to retract what I said on the previous round: I claimed this was fixed and it is not. Line 129 still passes `ct` — which is `_stopping.Token` — straight into `HeartbeatAsync`, and `Dispose()` still calls `Cancel()`. So the credential-loss path is behaviourally identical to round 2. I changed what the loop waits on and wrote a doc comment asserting the rule, without changing the token the request is actually issued with.

`docs/CHANGES.md` says both things in the same section, three paragraphs apart, and the code matches the wrong one. That is my error twice over — the claim in the reply and the claim in the changelog.

Working through your findings now; the token is the first of them.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Fixed for real this time: `SendAsync` takes no token and calls `HeartbeatAsync(serverUrl, flowId, CancellationToken.None)`. What ends a beat is the client's own timeout; what ends the loop is the token, one level up. The changelog contradiction is gone — the paragraph that said the cancel aborts it was the stale half and has been rewritten.

`A_hung_beat_is_neither_cancelled_nor_joined` asserts `CanBeCanceled` is false on the token the channel actually received, so a token reaching the request again fails the test rather than passing a comment.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Fixed by holding the task rather than dropping it: the loop keeps one `pending` beat across ticks and reads its verdict on the tick after it completes, so a slow 429 or 404 is acted on whenever it lands rather than only when it beats the interval. Same change closes the accumulation below — they were one bug wearing two hats.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Fixed — a new beat is only started when none is pending, so the lane holds at one. Taking your point about the deferred client having no `Timeout`: an outstanding beat now decides how long a wedged connection can silence liveness, so leaving that at the 100s default would have traded one problem for another. It is set to the browser leg's `BrowserFlowHttpTimeout`, so both halves answer for silence on the same scale.

src/Capacitor.Cli/Commands/SetupCommand.cs
George-Payne: Correct, and I have deliberately not fixed it in this PR — I would rather say so than quietly leave it.

Closing it needs the beat to outlive `RunBrowserFlowStepAsync`, and the client it beats on is `using (http)`-scoped inside that method, so spanning the prompts means restructuring the leg to hand its client back and keeping it alive across four prompts. That is a client-lifetime change in `setup`, and getting it wrong breaks setup outright — a worse failure than the one it fixes, and not something I want to bolt onto the end of a review round.

The flap also needs the server to report silence for a finished flow, which it does not: the page suppresses the silent tail on `IsFinished()` and re-folds from the store before claiming lost contact, so the gap is real but currently unobservable. That is a reason to schedule it, not to dismiss it, and I am treating it as follow-up work rather than as answered.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
George-Payne: Right on both counts — `send` is only invoked when the reason is non-null, so the callback never fires for `Finished` or `Expired` and my comment claimed more than it delivered. There is now a `beat.Dispose()` in the leg's `finally` as well, so every ending that reaches it stops the timer, and the comment says which path covers which.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Correct — the early return broke the invariant both comments asserted. The source is no longer disposed by the loop at all: it holds no timer and no registrations once the loop ends, so leaving it to the collector removes the race rather than documenting around it. Both comments have been rewritten to say what is true now.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Fixed — the channel call is now invoked inside `SendAsync` rather than evaluated as an argument, so a synchronous throw is caught like any other. You are right that the existing test only pinned the `async` case; the fakes in this suite are the counter-example sitting in the same file.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Both points taken. It now takes three consecutive not-found answers to stop, so a gateway 405 mid-deploy or a proxy that has not learned the path is transient rather than terminal, and any other status resets the count. `A_single_not_found_does_not_stop_the_beat` pins that directly.

The "same oracle the create uses" line was inaccurate and is gone — the create's set is 404/401/403/405 and it abandons the whole leg, which is a different decision from one route giving up.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -908,16 +908,32 @@ await ConfigMutator.MutateAsync(config, c => {
 
                 using var deferred = deferredHttp;
 
+                // The beat holds its lane until a request finishes, so this client's timeout is how long a
+                // wedged connection can silence liveness. Left unset it is the 100s default; matched to the
+                // browser leg's so both halves of the flow answer for silence on the same scale.
+                deferred.Timeout = BrowserFlowHttpTimeout;
```
rsergeenko, line 914: **This re-opens the credential-stranding hole the withheld token was meant to close.**

`HttpClient.Timeout` is not a bound *outside* the request: it cancels the very `CancellationToken` the handler chain receives, so from the recovery path's point of view it is indistinguishable from the stop token this commit just removed.

A beat that 401s enters `UnauthorizedRetryHandler` → `TokenStore.RecoverForServerAsync` → `RefreshWithCrossProcessLockAsync`, whose lock wait alone allows 15s (`TokenStore.cs:527`) and whose persist is `await SaveAsync(profile, refreshed, cancellationToken)` (`TokenStore.cs:576`) — and all of it, original POST plus refresh plus persist plus the retry POST, has to fit inside this same 15s.

Concretely: a peer `kcap` process (hook/watcher/daemon/MCP) holds `<profile>.lock` for ~8s; the WorkOS rotation POST takes ~4s; at ~13s the rotated token is in hand and `SaveAsync` starts; the timeout fires, the write throws, the `finally` deletes the temp. The single-use refresh token is spent server-side, the replacement is never written, and the user is logged out mid-setup — the same failure as the previous two rounds, reached through the timeout instead of `Dispose`.

Worth noting this line also narrows the deferred leg from the 100s default to 15s, which *increases* the chance the timeout lands inside a recovery rather than reducing it.

Withholding the stop token does not establish the invariant by itself. Either the persist (ideally the rotation too) runs under `CancellationToken.None`, or the beat does not ride the client that owns 401 recovery.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,152 @@
...
+
+                    if (outcome.StatusCode is 429) {
+                        var asked = outcome.RetryAfter ?? ThrottleBackoff;
+
+                        quietUntil = clock.GetUtcNow()
+                                   + (asked > MaxThrottleBackoff ? MaxThrottleBackoff : asked);
+                    }
+                }
+
+                if (pending is null && clock.GetUtcNow() >= quietUntil)
```
rsergeenko: **One-in-flight makes a single stalled request the entire liveness budget.**

A new beat is issued only in the iteration that reads the previous verdict, so silence lasts exactly as long as the request does: with a 15s client timeout that is 15s plus up to a tick, ≈20s — four consecutive missed beats, where `Interval` was chosen so that "a single dropped beat is not a verdict". A channel leaving `HttpClient.Timeout` at the default stretches it to ~105s.

Latency short of a hang degrades the same way: a 6s round trip means the verdict is not ready at the next tick, so the cadence halves to one beat per 10s; 11s latency gives one per 15s. The population this feature exists for — tethered link, cold TLS after a wake — is the one that now reads as dead.

A bounded overlap — a second beat once the pending one has outlived, say, two intervals, capped at two — keeps the anti-pile-up property without turning slow into dead.

src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -908,16 +908,32 @@ await ConfigMutator.MutateAsync(config, c => {
...
                 // answered — the browser leg skips there only because it has nothing left to do.
-                if (deferredAuth is AuthStatus.Ok or AuthStatus.NoAuthRequired)
+                if (deferredAuth is AuthStatus.Ok or AuthStatus.NoAuthRequired) {
+                    var channel = new FirstRunFlowClient(deferred);
+
+                    // The browser is holding the request this leg is about to perform, and enabling a
+                    // service can outlast any staleness window — so the beat has to span it. Only for a
+                    // flow that finished: every other ending already relinquished, and beating again
+                    // would take that back.
+                    using var beat = browserAnswers.FlowStillLive
```
rsergeenko, line 930: **Still open: the beat does not span the interactive stretch, which is the gap it was asked to cover.**

`BrowserFirstRunFlow` disposes its beat in the `finally` around `PollAsync`, and this one starts here. In between, on a `Finished` flow — the only case where `FlowStillLive` is true — setup runs `CodingAgentsStep.RunAsync` (line 789: installs into up to nine harnesses, can prompt), the provider-API-key confirm (847), and, always, since the browser leg implies `!noPrompt`, the Step 5 daemon-name `TextPrompt` (864), which blocks on a keypress for unbounded human time.

The browser is holding the outstanding daemon-service request across all of it, so it watches the machine go stale for minutes and then come back — precisely the false "machine gone" that the 404 leniency and the throttle clamp elsewhere in this PR work to avoid.

Starting the beat as soon as the leg returns a still-live flow, and disposing it after the daemon leg, closes the gap.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,152 @@
...
+                if (pending is { IsCompleted: true }) {
+                    var outcome = await pending;
+
+                    pending = null;
+
+                    // Transient rather than terminal on the first refusal: a gateway 405 mid-deploy or a
+                    // proxy rejecting POST on a path it does not know would otherwise silence liveness for
+                    // the rest of the leg while the poll keeps succeeding on the same client — the browser
+                    // inferring the machine has gone from one that is demonstrably still talking to it.
+                    if (outcome.StatusCode is 404 or 405) {
```
rsergeenko: **Three consecutive refusals is still permanent for the leg — there is no re-probe.**

Three beats span ~10-15s, comfortably inside a rolling deploy or a proxy reload, so a blip the 2s poll on the same client rides out silences liveness for the remaining ~30 minutes of `PollBudget`. By the fix's own rationale — a blip cannot be terminal — a capped backoff (go quiet a minute or two, then probe once more) preserves both the no-op ceiling and recovery.

Related doc drift: `FirstRunHeartbeatTests.cs:276` still asserts in prose that "The create path reads 404 the same way", which is no longer true — the create treats a single 404 as unavailable (`BrowserFirstRunFlow.cs:84`), the beat needs three.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// <summary>
+    /// Stops scheduling. Nothing is cancelled: the request a beat is sitting in may be recovering a
+    /// credential, and killing it there is what strands a rotated refresh token — see
+    /// <see cref="SendAsync"/>. An outstanding beat is left to finish or to time out on the client's
+    /// own deadline, and a beat that lands late costs nothing a stated ending does not outrank.
+    /// </summary>
+    public void Dispose() {
+        if (Interlocked.Exchange(ref _stopped, 1) != 0) return;
+
+        _stopping.Cancel();
```
rsergeenko, line 89: **`_stopping` is now disposed nowhere.**

The loop's `finally { stopping.Dispose(); }` was removed — correctly, for the race — but `Dispose()` only cancels, so every instance leaks its `CancellationTokenSource`. Harmless as the class stands (no timer, and the `WaitForNextTickAsync` registration is released as the await unwinds), and it stops being harmless the moment a linked source or a `CancelAfter` appears.

If you do add one, put it after the observation `ContinueWith`, and note that a loop iteration still heading into `WaitForNextTickAsync(ct)` would then throw `ObjectDisposedException`, which `catch (OperationCanceledException)` does not cover.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
```diff
@@ -127,13 +127,32 @@ public async Task<FirstRunFlowResult> RunAsync(
...
+        // interrupt path: Environment.Exit runs no `using` at all, so a beat scheduled by the timer could
+        // otherwise post inside the notice's own budget and say this machine is here, moments after it
+        // said it had gone.
+        //
+        // Only the paths that SEND come through here — a notice with nothing to say never invokes this —
+        // so the `finally` below is what covers the rest.
         using var notice = _interrupts.Arm(
-            (reason, token) => RelinquishAsync(serverUrl, flowId, reason, token),
+            (reason, token) => {
+                beat.Dispose();
```
rsergeenko, line 144: **Two residuals from no longer cancelling on the interrupt path.**

(a) This dispose runs only when the notice has something to send — `FirstRunNotice.RunAsync` skips `send` for a null reason, and `Environment.Exit` runs no `finally` — so an interrupt arriving after `settled` is set to `Finished`/`Expired` but before the `finally` executes leaves the timer scheduling through the exit. The comment claiming the `finally` covers the rest holds only for non-interrupt endings.

(b) A beat can now be *issued* after the relinquish: if a beat 401s, `UnauthorizedRetryHandler` sends its retry POST once recovery completes, which can be after the relinquish POST has landed — the server then records the machine live moments after it stated it had gone. The old cancel closed both. If that ordering matters, the relinquish has to be terminal server-side rather than "a stated ending outranks an inferred one" by convention.

test/Capacitor.Cli.Core.Tests.Unit/FirstRun/FirstRunHeartbeatTests.cs
```diff
@@ -0,0 +1,313 @@
...
+
+        using var beat = FirstRunHeartbeat.Start(channel, Server, Flow, clock, Beat);
+
+        for (var i = 0; i < 20; i++) {
+            clock.Advance(Beat);
+
+            await Task.Delay(5);
+        }
+
+        await Assert.That(channel.Beats).IsLessThanOrEqualTo(4)
```
rsergeenko: **The headline behaviour of this commit is not pinned.**

Every fake outcome completes synchronously, so no test exercises a verdict that arrives *slower than one interval* — which is the whole point of carrying `pending` across ticks. This test would pass under the old abandon-the-wait code too (`Beats == 1`, ≤ 4), and `IsLessThanOrEqualTo(4)` is satisfied by a beat that stops after one, so it no longer distinguishes the constant it describes: the loop sends exactly 3, so `IsEqualTo(3)` is the assertion.

`FakeChannel.Block`/`Release` already provides the tool — a 429 or a 404 released after two or three clock advances would show a late verdict actually being acted on.

rsergeenko (commented):
Fourth pass, after `01c2b982`.

**Resolved:** the verdict of a slow beat is now read (`pending` carried across ticks), beats no longer pile up behind a wedged POST, the synchronous-throw hole is closed by invoking the channel inside `SendAsync`, the CTS-dispose race is gone and the comments that asserted the broken invariant were corrected, and 404/405 needs three consecutive refusals. Build clean; `FirstRunHeartbeatTests` (12) and `BrowserFirstRunFlowTests` (103) green.

**The one that has not moved** is the credential hazard, now reachable through a third route. Each round has relocated the cancel — per-beat CTS → stop token → client timeout — rather than taking it off the recovery path. `HttpClient.Timeout` cancels the same token the handler chain receives, so `SendAsync`'s "what ends a beat is the client's own timeout" is not the outside-the-request bound the comment reasons from. The durable fix is one line in `TokenStore` (persist the rotated token under `CancellationToken.None`) or a beat client without the 401 handler; everything else in this round is small.

src/Capacitor.Cli/Commands/SetupCommand.cs
George-Payne: You are right, and the correction matters: I had been treating `HttpClient.Timeout` as a bound outside the request. It is not — it links its own source into the token the handler chain receives, so withholding the stop token bought nothing on this path. Three rounds relocated the cancel instead of removing it, which is exactly what you said.

Filed as **AI-2515** — https://linear.app/kurrent/issue/AI-2515/a-cancel-between-the-workos-rotate-and-its-persist-logs-the-user-out — with your fix: persist under `CancellationToken.None`, keep cancellation before the rotate, and a test that a cancelled token still leaves the rotated credential on disk.

Filed rather than fixed here because it is not this PR's bug and not the beat's. `TokenStore` is shared by every command, and the widest exposure is the poll: every 2 seconds for up to the 30-minute browser budget, on a client with the same 15-second timeout and the same 401 handler. A first-run PR is the wrong blast radius for that change, and scoping the fix to the beat would leave the poll exposed to a logout with a narrower-looking diff.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Fixed. The bound is now a count rather than a lane: up to three beats may be outstanding, so a slow answer cannot stop the loop issuing on the tick, and a wedged network still cannot accumulate one open POST per interval. `A_hung_beat_is_neither_cancelled_nor_joined` asserts exactly three against a request that never completes — the cap, not an inequality that a beat stopping after one would also satisfy.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Fixed — it is a pause now, not an ending. Three consecutive refusals go quiet for two minutes and then probe again; any other status resets the count. The test asserts the beat resumes after the window, so a terminal return fails it.

The prose drift is gone too: the line claiming the create path reads 404 the same way was wrong for the reason you gave — the create's set is wider and it abandons the whole leg.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Disposed now, in `Dispose()` after the cancel and the observation, with the loop catching `ObjectDisposedException` alongside the cancel — a tick entering `WaitForNextTickAsync` as the source goes ends exactly as it would on the cancel. Taking your point that it stops being harmless the moment anything here links a source or sets a deadline, which is a change someone could reasonably make without reading the old comment.

src/Capacitor.Cli.Core/FirstRun/BrowserFirstRunFlow.cs
George-Payne: (a) Fixed: the beat is stopped from the reason callback as well as the send. That one is evaluated on every interrupt whatever reason it settles on, so a `Finished` or `Expired` leg now stops the timer too. The comment says which callback covers which case rather than claiming the `finally` covers everything.

(b) Acknowledged and not fixable here as far as I can see. A retry POST issued by the 401 handler after the relinquish cannot be stopped without cancelling it, which is the thing that strands a rotated credential. That leaves server-side terminality carrying it, pinned in kurrent-io/kcap-server#1822 — a relinquished flow keeps its own tail whatever liveness reports.

test/Capacitor.Cli.Core.Tests.Unit/FirstRun/FirstRunHeartbeatTests.cs
George-Payne: Both right, and the first one stings: I wrote that assertion to guard a constant I had just introduced, and `IsLessThanOrEqualTo(4)` does not distinguish it from a beat that stops after one. It is `IsEqualTo(3)` now.

The missing case is covered too — `A_verdict_that_arrives_late_is_still_acted_on` holds every answer past the tick that asked for it, then releases a 429 and asserts the throttle still takes effect. I checked it fails when the harvest is stubbed out, since a test for a behaviour no fake could previously produce is worth proving rather than assuming.

src/Capacitor.Cli/Commands/SetupCommand.cs
George-Payne: Still open, deliberately, and now with a ticket rather than a promise: the client-lifetime change this needs is in `RunBrowserFlowStepAsync`, whose `using (http)` scope is what the beat would have to outlive. Thank you for naming the specific prompts — `CodingAgentsStep` at 789, the API-key confirm at 847 and the unconditional daemon-name `TextPrompt` at 864 is a much better description of the window than "the prompts".

rsergeenko, line 930: **Disposing the client cancels the in-flight beats — a second site for AI-2515's hazard, and this one is inside this PR's own code.**

Agreed on deferring the `TokenStore` fix; the poll being the wider exposure is the right call. This is a separate site rather than a re-raise of it.

`beat.Dispose()` deliberately does not wait, and `deferred` is disposed one scope out (`using var deferred`, line 909). `HttpClient.Dispose` cancels its pending-requests CTS and disposes the handler, so leaving this `if` block cancels up to `MaxInFlight` (3) outstanding beats — no timeout needs to elapse for it, and the cap widened it from one beat to three.

The timing is the worst case: a token that lapsed during a long browser wait is recovered on the *first* beat after it, so the beat most likely to be mid-`RecoverForServerAsync` is the one the scope exit cancels.

Same shape in the browser leg: `using (http)` at line 1495 ends immediately after `BrowserFirstRunFlow` returns, cancelling whatever beats its `finally`-disposed heartbeat left outstanding.

Worth noting this is reachable without AI-2515 being fixed *or* unfixed — a bounded join inside `Dispose` is compatible with "does not cancel", or the beat gets its own client whose lifetime it owns. If it is going to wait on AI-2515, it is worth a line on that ticket, since the fix there (persist under `CancellationToken.None`) would also neutralise this route.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,180 @@
...
+    /// <summary>
+    /// How many beats may be outstanding at once.
+    ///
+    /// <para>More than one, or a single slow request becomes the whole liveness budget: the loop would
+    /// issue nothing until the answer came back, so a request that outlives its interval halves the
+    /// cadence and one that hangs to the client's timeout silences the machine for the whole of it.
+    /// Bounded all the same, because a wedged network would otherwise accumulate one open POST per
+    /// interval on the very machine whose network is failing.</para>
+    /// </summary>
+    const int MaxInFlight = 3;
```
rsergeenko, line 29: **`MaxInFlight` and `UnavailableBeforeBackingOff` are both 3, so "consecutive" no longer means three probes over time.**

With three beats overlapping, a proxy reload can answer all three `405` at the same instant, and one blip trips the full 2-minute quiet window — the exact failure `UnavailableBeforeBackingOff`'s own doc says it prevents ("More than one, so a blip cannot silence it").

Counting refusals per issuing round, or raising the threshold above the in-flight cap, restores the intent.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,180 @@
...
+        using var timer = new PeriodicTimer(interval, clock);
+
+        var inFlight = new List<Task<FirstRunHeartbeatOutcome>>(MaxInFlight);
+
+        var quietUntil  = DateTimeOffset.MinValue;
+        var unavailable = 0;
+
+        try {
+            while (!ct.IsCancellationRequested) {
+                for (var i = inFlight.Count - 1; i >= 0; i--) {
```
rsergeenko: **The harvest walks backwards, so when several beats complete in one drain the oldest answer is applied last and wins.**

A server that throttles hard and then recovers — beat 1 answers `Retry-After: 120`, beat 3 answers `Retry-After: 1` — is obeyed at 2 minutes rather than 1 second. And each late 429 recomputes `quietUntil` from *its own* arrival, so the silence extends up to `(MaxInFlight - 1) * interval` past what any single instruction asked for. A 404-driven backoff can likewise be shortened by a 429 processed after it in the same drain.

Taking the newest verdict (or the max quiet window) makes the outcome independent of iteration order.

test/Capacitor.Cli.Core.Tests.Unit/FirstRun/FirstRunHeartbeatTests.cs
```diff
@@ -0,0 +1,372 @@
...
+    /// uncancellable. A cancel landing between the two spends the credential server-side and never writes
+    /// the replacement, logging the user out mid-setup — so the stop token is kept off the request
+    /// entirely and the client's own timeout is what ends a beat.</para>
+    ///
+    /// <para>The no-second half is the cost of that: an outstanding beat holds the lane, so a wedged
+    /// network is silent until the request times out rather than piling up one open POST per interval on
+    /// the very machine whose network is failing.</para>
+    /// </summary>
+    [Test]
+    public async Task A_hung_beat_is_neither_cancelled_nor_joined() {
```
rsergeenko: **Doc drift, in two tests.**

This one asserts `Beats == 3` — the hung beat *is* joined, twice — while its name and doc still say "never joined by a second" and "an outstanding beat holds the lane, so a wedged network is silent until the request times out". That was true one commit ago and is the opposite of what the cap now does.

`A_route_that_is_never_there_stops_the_beat` (line 293) has the same problem: it pins a pause-and-re-probe, not a stop, and its doc still says the route "answers the same way for the rest of the leg".

A reader with only the current source is told the opposite of what the assertions pin, so both the names and the docs want updating.

rsergeenko (commented):
Fifth pass, after `d60b0f60`. Revised: I had not read the replies on the earlier threads when I first posted this, and three comments re-raised items that already had standing answers. Those are withdrawn — AI-2515 (the rotate/persist cancel), the interactive stretch, and the post-relinquish retry POST pinned in kcap-server#1822 are all answered, and deferring the `TokenStore` fix is the right call: the poll is the wider exposure at every 2s for 30 minutes on the same client, so scoping it to the beat would leave that behind a narrower diff.

**Resolved in this commit:** the in-flight cap is a clean fix for the serialisation problem — 3 x 5s lines up with the 15s client timeout, so a wedged network keeps its cadence; every capped beat is still observed, the list cannot grow, and nothing can deadlock. `UnavailableBackoff` makes the 404 route a pause rather than an ending, `_stopping` is now disposed, `interruptReason` covers the null-reason interrupt, and `A_verdict_that_arrives_late_is_still_acted_on` pins the behaviour the previous commit introduced.

**What remains below** is four items, all new with this commit: the client's disposal cancelling in-flight beats (a second site for AI-2515's hazard, inside this PR's own code), the in-flight cap colliding with the refusal threshold, the harvest order deciding which verdict wins, and two test names whose docs now contradict their assertions.

src/Capacitor.Cli/Commands/SetupCommand.cs
George-Payne: Accepted as a real site, and refuted as something to fix here. Recorded on [AI-2515](https://linear.app/kurrent/issue/AI-2515) as a fourth route to the same window, with your timing point, since the ticket's fix is the one that closes it.

Both call-site remedies make things worse rather than better:

**A bounded join in `Dispose` does not help where it matters.** The disposal that coincides with a recovery is the interrupt path, and there `beat.Dispose()` runs from the SIGINT callback and is followed by `Environment.Exit`. Waiting delays the relinquish notice the browser depends on, and the process then kills the request anyway — a kill strands the credential exactly as a cancel does. On the graceful path it buys a narrower window at the cost of setup latency, and leaves the same hazard reachable through the client's own timeout.

**A separate client for the beat is ruled out by `UnauthorizedRetryHandler`'s own doc:** *"Only ONE component may own 401-retry on a given client, or a single rejection multiplies into several refreshes."* A second client means a second handler with its own `_current`, so both would refresh on the same rejection — two racing rotations of a single-use refresh token. That widens AI-2515 rather than closing it.

Persisting under `CancellationToken.None` closes this route along with the other three, and is the only change that closes all four.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Fixed in `2fa244cd`. `UnavailableBeforeBackingOff` is now `MaxInFlight + 1` rather than a bare `3`, expressed against the cap so the two cannot drift apart again — refusals have to span more than one issuing round, which is what "consecutive" was always meant to buy.

`A_route_that_is_never_there_pauses_the_beat` pins the new count exactly (4, then a probe on the 5th), so lowering the constant back under the cap fails rather than passing silently.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Fixed in `2fa244cd`. The harvest walks forward now — oldest first, removing in place — so a drain ends on the newest answer and the server's latest instruction is the one left standing.

Took "newest wins" rather than "max window": max would keep the `Retry-After: 120` in your example, which is the outcome the finding calls wrong.

`The_newest_answer_in_a_drain_is_the_one_that_stands` pins it: three beats held until the lane is full, answered `120s / 120s / 1s` in issue order, then released together and required to resume well inside 30 seconds. Reverting the loop to walk backwards fails it.

The `(MaxInFlight - 1) * interval` overshoot goes with it — a superseded instruction can no longer be the one that sets `quietUntil`, so the only delay left is a live instruction applied from its harvest rather than its arrival, which is at most one tick.

test/Capacitor.Cli.Core.Tests.Unit/FirstRun/FirstRunHeartbeatTests.cs
George-Payne: Fixed in `2fa244cd`, both of them.

This one is now `A_hung_beat_is_never_cancelled_and_does_not_hold_the_whole_budget`, and the second `<para>` says what the cap does rather than what serialising used to.

`A_route_that_is_never_there_stops_the_beat` is now `..._pauses_the_beat`, and its doc says the beat backs the route off and probes it again, with the rolling-deploy reason for that being the point rather than an aside.

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
```diff
@@ -0,0 +1,193 @@
...
+                        i++;
+
+                        continue;
+                    }
+
+                    var outcome = await inFlight[i];
+
+                    inFlight.RemoveAt(i);
+
+                    if (outcome.StatusCode is 404 or 405) {
```
rsergeenko: **The quiet window is folded answer-by-answer and never re-decided, so reading oldest-first can now enter a pause the server's latest word contradicts.**

`quietUntil` is only ever pushed further out — nothing clears it. This branch trips `UnavailableBackoff` mid-drain, and a *newer* answer processed after it can only reset `unavailable` on line 146; it cannot lift the window the older refusal just caused.

Reproduced at head with a temporary test: beats 1-2 answer `404` immediately, then beats 3-5 are held so they drain together as `404 / 404 / 204`. The 4th refusal trips the backoff, the newer `204` resets only the counter, and the beat stays silent for the full two minutes. Under `d60b0f60`'s backwards loop the same fake resumes beating — the `204` reset the counter before the refusals were counted, so no trip happened.

So the machine goes silent for two minutes *starting at the moment the route came back*, on a client whose poll is succeeding every 2s — which is what `UnavailableBackoff`'s own doc says must not happen ("the browser infer the machine has gone from one that is demonstrably still talking to it"). A rolling deploy where drained pods 404 and new pods 204 is the concrete trigger.

Two corollaries of the same root cause:

- **Across drains the ordering claim does not hold.** An older beat carrying `429 Retry-After: 120` that completes after newer beats have already been harvested — per-connection pool, no response ordering — still sets a two-minute window though the server has since answered `204` twice. So "a superseded instruction can no longer be the one that sets `quietUntil`" is true only *within* one drain, and `docs/CHANGES.md`'s "the newest instruction is the one left standing" overclaims for the same reason. The intra-drain case the fix targets does work; I checked that one separately.
- **Mirrored the other way**, a newer `429` with a short `Retry-After` processed after a 404 trip discards both the two-minute pause and the counter (the `else` reset), so a route that both 404s and 429s never sustains a backoff at all — the no-op POSTs the backoff exists to stop keep flowing.

The shape that closes all three: decide the window from the newest harvested answer rather than applying each in turn — remember the last answer in the pass and clear `quietUntil` when it is neither a throttle nor a refusal, or carry an issue sequence and ignore an answer older than the last one applied.

The forward harvest is still the right direction. The residue is that the fold itself is order-sensitive, not the direction it is read in.

rsergeenko (commented):
Sixth pass, after `1fad553b` + the merge.

All three round-5 items verified fixed. `UnavailableBeforeBackingOff = MaxInFlight + 1` is the right shape — 4 refusals genuinely need two issuing rounds, and expressing it against the cap stops the two drifting apart. The forward harvest is correct as iteration: `i++` only on the not-completed branch, `RemoveAt(i)` without increment, `Count` re-read each pass, and `inFlight` really is issue-ordered. Both renamed tests now match their assertions. The merge touches none of the feature files; both test projects build clean and the two suites pass. I also checked that `deferred.Timeout` cannot throw "already started a request" (the factory never sends on the client it returns) and that overlapping beats cannot double-rotate a refresh token (`needsRefresh: t => t.AccessToken == rejectedAccessToken`, re-read under the cross-process lock).

Agreed on AI-2515 and its four routes, including the two call-site remedies — the `Environment.Exit` point retires the bounded join, and `UnauthorizedRetryHandler`'s one-owner rule (which I checked) retires the second client. Nothing re-raised from (a)-(d).

One finding below, and not a blocker: a self-healing two-minute false silence, no data or credential loss. Worth fixing because it inverts a guarantee the code states, and the fix is small.

rsergeenko (dismissed):

src/Capacitor.Cli.Core/FirstRun/FirstRunHeartbeat.cs
George-Payne: Confirmed and fixed in `879ab065`. Your repro is right: `quietUntil` was only ever pushed out, so a newer answer could reset the counter but not lift the window an older refusal had just bought.

The fold is gone. The newest answer decides the window outright — `429` sets its clamped delay, a refusal sets the backoff only when the run has reached the threshold, and anything else clears it — so a route that answers again mid-drain lifts the pause rather than inheriting it. That closes your second corollary too: a `429` arriving after a refusal trip no longer discards a window it should replace, because there is no accumulated window to discard.

Beats now carry the order they were issued in, and an answer older than the word in force is counted towards the refusal run but does not set the window. That covers the cross-drain case.

**One thing I did not do:** add a test for the cross-drain guard. I wrote one, then found it passed with the guard removed, and the reason is worth stating rather than papering over — the loop harvests every tick regardless of the window, so a stale answer that sets a pause is re-decided by the next answer harvested, usually one tick later. Without the guard the observable damage is bounded by about one interval, not by the two minutes the stale instruction asked for. The guard makes the fold order-independent by construction, which is why it stays, but I am not going to add a test that passes either way to claim it is pinned.

The intra-drain case is pinned: `An_answer_after_the_refusals_lifts_the_pause_they_would_have_bought` runs your rolling-deploy fake — two refusals answered on their own, then a drain of `404 / 404 / 204` — and fails if the window is folded rather than decided.

`docs/CHANGES.md` no longer claims the newest instruction always wins; it now says the newest *decides*, and that an older one is counted but not obeyed.

rsergeenko (dismissed):

rsergeenko (approved):

Issue #707, opened 2026-08-31
Daemon suite flakes on ubuntu CI: GitRepo fixture state vanishes mid-test


alexeyzimarev:
The ubuntu `Build and test` leg intermittently fails one daemon-suite test whose fixture is a `GitRepo`, with git state that should be stable disappearing or mutating between two steps of the same test. Two recent occurrences, different tests, same shape:

- Run [33260948585](https://github.com/kurrent-io/kcap-cli/actions/runs/33260948585) (main, `53be3b75`): `Borrowed_launch_creates_no_worktree_and_runs_in_the_cwd` — the before/after tree snapshots differ by exactly `.git/objects/maintenance.lock`, a transient artifact of git's *background* maintenance, which evidently runs inside the fixture repos on CI.
- Run [33397578952](https://github.com/kurrent-io/kcap-cli/actions/runs/33397578952) (PR #706, App-only diff): `Interactive_codex_launch_threads_the_posture_into_the_runtime_start_context` — `GitRepo.CreateWithCommit` fails 10 ms in with `git add -A` → 128 `fatal: not a git repository`, immediately after a successful `git init` in a uniquely named owned TempDir (`Directory.CreateTempSubdirectory`, so a path collision is ruled out).

Both runs' suite stderr also carries `UnixSpawnerThread terminated unexpectedly` / `Environment.FailFast` — possibly the intentional crash-contract test's child process rather than part of the flake; not yet confirmed either way.

The `maintenance.lock` sighting suggests the assembly-wide git config pin (`Guards/GitConfigGlobalSetup`) does not disable background maintenance/auto-gc in fixture repos; a detached `git maintenance` process from an earlier test also outlives its TempDir. Whether that also explains the `not a git repository` case is open.

PR #768, opened 2026-09-04
Name the repo's projects in the SessionStart memory lead-in

### LINKED ISSUE #756: SessionStart memory index: name the repo's project in the lead-in
## Problem

Agents do not know which project the cwd repo belongs to, so they cannot target `project` on `save_memory`. The SessionStart memory index only tags entries that are already project-scoped (`[project: slug]`); nothing says "this repo is in project X".

## Change

- At SessionStart, fetch the repo's confirmed projects from the server (kcap-server AI-2470 decides the endpoint shape; the index body stays a bare array for older CLIs).
- Render a lead-in line in the memory fragment, for example: `This repo belongs to project "capacitor". Save learnings that span its repos with project: "capacitor".` One line per project when the repo is in several.
- Fail-open like the rest of the index fetch: a missing or failed call emits no lead-in.
- `MemoryIndexEmitter` keeps the existing `[project: slug]` annotation on entries.

## Depends on

kcap-server AI-2470 (part of AI-2468).

alexeyzimarev:
Closes #756 — AI-2474

## What & why

An agent can only land a memory at project scope by passing a slug, and nothing in a session told it which project the cwd repo is in — so cross-repo learnings went to org, the only cross-repo scope reachable without one. The memory fragment now opens with a line per project naming that slug.

The projects ride the existing index call, negotiated by `include=projects`: the body is a bare array and a CLI that predates this drops a fragment that is not one, so the shape could not simply change. A sibling or superseding endpoint costs a second round trip inside a hook budget already tight on Cursor and Claude; a response header cannot carry a non-ASCII project name. The parameter degrades in both directions — an older server ignores it and answers with the array, a newer one answers with an object — and the CLI decides which it got by sniffing the opening token.

The server half is kcap-server AI-2470, not yet built. Until it ships the CLI sees the bare array and emits no lead-in. The wire contract this is built against is written up on that issue.

## Where to look

`include=projects` goes out on every index call, with or without a resolved repo: it declares that this CLI can read the object body, not that a repo is in hand.

Projects alone produce a fragment where entries alone produce none — a project holding no memories yet is exactly the state the agent is being asked to fix, and it cannot without the slug. A repo with an empty index and a project therefore gets one line of context instead of silence.

## Verification

```
Capacitor.Cli.Tests.Unit         3978 passed, 0 failed, 19 skipped (gated live certs)
Capacitor.Cli.Tests.Integration   245 passed, 0 failed
dotnet publish -c Release         exit 0, no IL2026/IL3050
```

Run the two suites separately: concurrently they starve `SessionStartMemoryLeaseStore`'s one-second budget and a handful of lease tests fail on contention alone.


alexeyzimarev:
Review triage:

**1. Partial objects consume lease — accepted, fixed in bfdd506b.** Real, and a regression this PR introduced: the array-only parse used to throw on any object body and come back retryable, so `{}` was already safe before. `ParseIndex` now rejects an object carrying neither member, keeping "absent" distinct from "present but empty" — an empty array is a real answer and still completes the lease. Covered by parameterised cases on both sides of that line (`{}`, `{"entries":null,"projects":null}`, `{"unrelated":1}` retry; `{"entries":[],"projects":[]}`, `{"entries":[]}`, `[]` complete).

**2. Conflicting `[NotInParallel]` — not taking this one.** The 23 other tests in `ClaudeHookCommandTests` pair the same bare method-level attribute with the class-level `[NotInParallel("AuthProviderDiscoveryCache")]`, and the pairing is deliberate rather than accidental: the test captures `Console`, which the repo requires bare exclusivity for, and bare `[NotInParallel]` is exclusive against the whole assembly — it strictly subsumes the keyed constraint, so no isolation is lost. Dropping it to the keyed form would weaken the test; leaving only it in the class would make it the odd one out.

The ubuntu leg failed on `AgentOrchestratorSourceClaimTests.Deferred_launch_claims_then_forwards_without_rebinding_then_first_turns_then_confirms` — the known `GitRepo.CreateWithCommit` flake (#707), in an assembly this PR does not touch.

PR #823, opened 2026-09-08
Resolve vendor harnesses through HarnessRegistry, not the environment

### LINKED ISSUE #810: Resolve vendor harnesses through HarnessRegistry, not the environment
`HarnessRegistry`'s own doc states the rule: *"Built once at an entry point and passed down, so two consumers cannot resolve one vendor's root differently."* Each vendor's `FromEnvironment` repeats it: readers *"reach them through the instance the entry point built, never by resolving the override a second time."*

43 call sites do the opposite. They call `ClaudeHarness.FromEnvironment(home)`, `HarnessPaths.FromEnvironment(home)`, `HarnessRegistry.FromEnvironment(home)` or `BinaryProbe.FromEnvironment()` deep in the call stack, each re-reading `CLAUDE_CONFIG_DIR`, `CODEX_HOME`, `GEMINI_CLI_HOME`, `KIRO_HOME`, `PI_CODING_AGENT_DIR`, `OPENCODE_CONFIG_DIR`, `XDG_CONFIG_HOME`, `XDG_DATA_HOME` or `PATH` at the moment it runs.

## What this actually costs

**`Searching(BinaryProbe)` does not work.** It exists so a GUI launch can rebase vendor lookup onto the login shell's PATH — the launcher inherits a different one. Today that rebase reaches only `HarnessRegistry.Detect`; `ClaudeCliRunner:195`, `CodexCliRunner:76`, `VendorVersionResolver:19` and `CliResolver:27` each call the static and silently get the launcher's PATH instead.

**One vendor root gets derived several ways in one operation.** `AntigravityReviewerHome` re-derives Antigravity from Gemini three times in one file (`:136`, `:159`, `:201`) — the exact double-derivation `AntigravityHarness.Over` was written to prevent.

**Tests cannot describe a machine.** Any suite touching a vendor path has to pin the process environment and carry a `NotInParallel(VendorEnvOverrides)` key, because the value is read during the call rather than passed into it.

## Approach

Make `HarnessRegistry` the single vendor primitive and delete `HarnessPaths`.

1. **Per-vendor layout through `Of<TSelf>()`**, which already exists — `registry.Of<ClaudeHarness>().Paths.Projects` replaces the 27 per-vendor sites. `AntigravityHarness` is an `IHarness<AntigravityHarness>`, so the registry composes it once from the same Gemini instance instead of each caller re-deriving it.

2. **The registry answers binary questions**, since it already holds the `BinaryProbe` and uses it in `Detect`. Adding `Resolve(name)` / `Finds(name)` is what makes `Searching` mean something.

3. **`kcap`'s own PATH probe stays out of it.** The `BinaryProbe.OnPath("kcap")` calls in `PluginCommand` ask about our binary, not a vendor; `SetupCommand` already models this correctly as a `CapacitorOnPath` delegate.

4. **`AgentsPaths` leaves the bundle** — `new AgentsPaths(home)` reads no environment variable and was only there for convenience.

5. **`HarnessPaths` and `TestHarnessPaths` are deleted.** `HarnessPaths`'s stated reason to be a record is `with`-substitution of one vendor, which no production code does; the only substitution is in test helpers, and `HarnessRegistry.Over` / `TestHarnesses.Over` already serve that.

## Scope

`Capacitor.Cli`, `Capacitor.Cli.Core`, `Capacitor.Cli.Daemon`. `Capacitor.App` has one site of the same shape (`AgentsStepViewModel.cs:204`) — follow-up.

Out of scope: `SqliteNativeResolver:91` calls `UserHome.FromEnvironment()` from a static constructor that runs before any caller exists, which its own comment records.

Part of the wider move off runtime environment reads and process-global state toward dependencies resolved once at a composition root.

Inok:
Closes #810 — AI-2562

## What & why

Vendor layout and PATH were resolved from the environment deep in the call stack, so one
operation could derive a vendor root more than once and a search path redirected by
`Searching()` reached only detection. `HarnessRegistry` becomes the single vendor primitive:
`Of<T>().Paths` for layout, `ResolveExecutable(HarnessId)` for a vendor's own binary, and an
injected `BinaryProbe` for arbitrary configured paths. `HarnessPaths` is deleted — its stated
reason to be a record was per-vendor `with` substitution, which only test helpers did.

Each harness now declares `CliBinary`, the command a caller spawns, separately from
`HarnessSignals`. The two answer different questions: Cursor declares no launch signal at all,
because its editor is what says it is installed, yet `cursor-agent` is perfectly spawnable.
Signals are three predicates — `LaunchSignal`, `UserDataSignal`, `Wired` — where `null`
uniformly means the vendor cannot answer, and detection order carries no meaning.

## Where to look

Three behaviour fixes ride along with the refactor.

`AntigravityReviewerHome` derived its isolated home through `GEMINI_CLI_HOME`, so a daemon with
that exported failed *every* Antigravity launch — interactive hosted included — with
`antigravity_reviewer_home_escaped_root`. It now derives from the home alone, and the variable
is stripped from the child, which would otherwise read its result channel and permission grants
from the operator's profile.

`plugin install --kiro` probed `kiro-cli` with `OnPath`, which walks `PATHEXT` and matches an
npm `.cmd` shim, then spawned the bare name through `CreateProcess`, which appends only `.exe`.
On Windows the clone failed right after the probe said yes. It now spawns the resolved path.

A borrowed reviewer given an isolated root also drops the variables that would replace it:
`COPILOT_HOME` under the deny-default sandbox, and `OPENCODE_CONFIG`/`OPENCODE_CONFIG_CONTENT`,
which merge over the empty config dir that exists to remove the nested-flow channel.

Eight `BinaryProbe.OnPath("kcap")` sites stay: they ask about our own binary over the process
PATH, so `Searching()` does not redirect them. One `HarnessRegistry.FromEnvironment` stays in
`Capacitor.App`.

## Verification

Build 0 warnings; `dotnet publish -c Release` reports no IL2026/IL3050.

| Suite | total | failed |
| -- | -- | -- |
| Cli.Tests.Unit | 4010 | 0 |
| Cli.Core.Tests.Unit | 3061 | 0 |
| Cli.Daemon.Tests.Unit | 3076 | 1 |
| App.Tests.Unit | 1548 | 0 |

The daemon failure is `Installed_codex_schema_matches_the_vendored_pin`, failing identically on
`main` — installed codex 0.153.4 against a pin taken from 0.147.0.

Every env scrub is pinned by a test that first establishes the value with `EnvScope.Exclusive`,
since a variable nothing set proves nothing about a scrub, and each is mutation-verified against
reverting it. The `Searching(probe)` claim above is pinned by a test that fails when `Searching`
is reverted to reuse the registry's original probe.


test/Capacitor.Cli.Core.Tests.Unit/Harness/HarnessRegistryTests.cs
```diff
@@ -32,14 +37,87 @@ public async Task Every_harness_carries_a_label() {
         await Assert.That(labels.Distinct().Count()).IsEqualTo(labels.Count);
     }
 
+    /// <summary>Each vendor's layout comes from its own factory rather than the registry reading the
+    /// override variables itself, so relocating one vendor leaves the others where they were.</summary>
+    // Bare: the overrides are inherited by any child a concurrent test spawns.
+    [Test, NotInParallel]
+    public async Task Each_vendor_is_routed_through_its_own_override() {
+        using var relocated = new TempDir();
```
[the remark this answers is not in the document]
Inok: Declining this one.

`CLAUDE.md` makes `using var tmp = new TempDir();` the documented default and reserves the `[TempDir]` injected property for the case where *every* test in a class needs one:

> When every test in a class needs one, inject it instead of holding a field

Neither class qualifies. `HarnessRegistryTests` has 8 tests and 3 need no directory (`Every_harness_is_registered_exactly_once`, `Every_harness_carries_a_label`, `An_id_this_registry_does_not_carry_yields_null`). `AntigravityReviewerHomeTests` has 15 tests and 3 use one.

Injecting would build a directory per test for the majority that never touch it, and the injected property is set after construction, so it cannot be read from a constructor or field initializer — which is why the `using var` form is the default rather than a fallback. The manual construction here is the sanctioned shape, not framework-unmanaged state: `TempDir` is `IDisposable` and each `using` scopes it to its test.

test/Capacitor.Cli.Daemon.Tests.Unit/Harness/Antigravity/AntigravityReviewerLaunchTests.cs
```diff
@@ -225,6 +225,22 @@ await Assert.That(psi.Environment[ConfigRoot.ConfigDirEnvVar])
...
+    [Test]
+    [NotInParallel]
+    public async Task The_turn_child_does_not_inherit_a_gemini_cli_home_override() {
+        const string inherited = "/tmp/kcap-gemini-cli-home-probe";
+        using var _ = EnvScope.Exclusive("GEMINI_CLI_HOME", inherited);
+
+        var psi = BuildPsi(EnabledConfig());
+        psi.Environment.TryGetValue("GEMINI_CLI_HOME", out var carried);
+
+        await Assert.That(carried).IsNotEqualTo(inherited);
```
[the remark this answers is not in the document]
Inok: Taken — the assertion was weaker than the behaviour it pins. `IsNotEqualTo(inherited)` passes if construction stamps some *other* wrong path, so it did not pin the scrub.

Now asserts the child carries no value at all:

```csharp
await Assert.That(psi.Environment.ContainsKey("GEMINI_CLI_HOME")).IsFalse();
```

Worth recording why that is allowed here, since `CLAUDE.md` otherwise forbids it:

> Never assert that an environment variable is *absent* from a built `ProcessStartInfo`

That ban exists because the child block is seeded from the current process, so absence cannot distinguish "the code removed it" from "it was never set". `EnvScope.Exclusive` establishes the value first, which is what makes absence meaningful — and the test carries a comment saying so, to stop the rule being applied back over it.

Mutation-verified: reverting the production `Remove` call turns the test red.

test/Capacitor.Cli.Core.Tests.Unit/Harness/HarnessRegistryTests.cs
```diff
@@ -32,14 +37,87 @@ public async Task Every_harness_carries_a_label() {
...
+
+        using var gemini = EnvScope.Exclusive("GEMINI_CLI_HOME", relocated.Path);
+
+        var harnesses = HarnessRegistry.FromEnvironment(new("/fake/home"));
+
+        await Assert.That(harnesses.Of<AntigravityHarness>().Paths.McpConfigJson)
+            .StartsWith(harnesses.Of<GeminiHarness>().Paths.Root);
+    }
+
+    // ── ResolveExecutable ──
```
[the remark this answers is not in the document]
Inok: Taken — removed. The banner restated the names of the methods below it and named no constraint, which is exactly what the comment rules exclude.

src/Capacitor.Cli.Daemon/Harness/Antigravity/AntigravityHostedAgentRuntimeFactory.cs
```diff
@@ -566,6 +568,11 @@ internal static ProcessStartInfo BuildTurnPsi(
         // KCAP_CONFIG_DIR exported had agy's own hooks reading the real profile by inheritance.
         psi.Environment[ConfigRoot.ConfigDirEnvVar] = ConfigRoot.UnderHome(home).Directory;
 
+        // GEMINI_CLI_HOME REPLACES the Gemini root outright (see AntigravityReviewerHome.LayoutFor),
+        // so an inherited one points agy's config, MCP servers and result channel at the operator's
+        // profile rather than this isolated HOME.
+        psi.Environment.Remove("GEMINI_CLI_HOME");
```
alexeyzimarev, line 574: **Inherited vendor overrides are scrubbed for Antigravity only; the same escape stays open for the other isolated reviewer children.**

`AcpHostedAgentRuntimeFactory.cs:950-962` (the `stateRoot is not null` sandbox block) sets `HOME`, `TMPDIR`, `KCAP_CONFIG_DIR` and the brokered token but never removes `COPILOT_HOME`, and `CopilotPaths` documents that Copilot relocates its whole tree through it. The sandbox profile is deny-default and never grants the operator's home (`BorrowedReviewRuntimeRoots`), so a daemon running with `COPILOT_HOME` exported hands the borrowed Copilot reviewer a config root it cannot read → every borrowed Copilot review fails closed — the exact symptom this hunk fixes for Antigravity.

`OpenCodeLaunchEnvironment.ApplyReviewer` sets `OPENCODE_CONFIG_DIR` only; `OpenCodeReviewerConfigDir`'s own doc names `OPENCODE_CONFIG` / `OPENCODE_CONFIG_CONTENT` as independent sources that merge *over* the dir, so an exported `OPENCODE_CONFIG` carrying `mcp.kcap-flows` reinstates the nested-flow channel the empty dir exists to remove.

Neither has the absence test Antigravity (`AntigravityReviewerLaunchTests:242`) and Kiro (`KiroReviewerLaunchTests:113`) carry. Pre-existing, but this PR is where the mechanism ("an isolated child HOME must also drop the vendor's override variables") gets named — a shared scrub applied by both factories, driven by each harness declaring its override variable(s), would close all three at once instead of one string literal per factory.

src/Capacitor.Cli.Core/Harness/HarnessRegistry.cs
```diff
@@ -108,6 +114,14 @@ public DetectedAgent Detect(HarnessId id) =>
     /// <summary>Whether either signal found this vendor.</summary>
     public bool Detected(HarnessId id) => Detect(id).Detected;
 
+    /// <summary>Full path to the first of <paramref name="id"/>'s declared
+    /// <see cref="HarnessSignals.Binaries"/> that resolves on this registry's search path — so a
+    /// caller resolving Antigravity does not have to know it answers to two names. Null when none
+    /// resolve, when the vendor declares no binaries (Cursor ships no CLI), or when this registry
+    /// does not carry <paramref name="id"/>, matching <see cref="Detect"/>.</summary>
+    public string? ResolveExecutable(HarnessId id) =>
```
alexeyzimarev, line 124: **`ResolveExecutable` returns the *first declared* name, and for both two-name vendors the first name is the IDE launcher, not the agent CLI.**

`AntigravityHarness.Signals.Binaries = ["antigravity", "agy"]` (its own comment: "The CLI binary is `agy`, not `antigravity`") and `KiroHarness` = `["kiro", "kiro-cli"]`. Kiro's docs say typing `kiro` launches the IDE; the Antigravity IDE installs an `antigravity` PATH shim. On a machine with IDE + CLI, the documented use ("a caller resolving Antigravity does not have to know it answers to two names") hands back the GUI launcher. `Detect` used `Any`, so order never mattered until this method; the new test stages only `agy`, so first-wins is unpinned. Today only Claude/Codex (single-name) call it, so this is latent — but the daemon already uses `agy`/`kiro-cli` via `DaemonConfig`, and finding 5 (Kiro install) is exactly the caller that cannot use this as-is.

Either order CLI-first in both harnesses and pin it, or document "first declared" and make `Binaries` order a stated contract.

Related: `Detect` already walks the same list (`Finds` is `Resolve(...) is not null`), so `DetectedAgent` could carry the resolved path and `ResolveExecutable` become `Detect(id).Executable` — one walk, one truth.

src/Capacitor.Cli/Program.cs
```diff
@@ -862,7 +862,7 @@ await Console.Error.WriteLineAsync(
 
 } finally {
     await UpdateNotice.FlushAsync(command, args, profiles, config, Run<NpmRegistryClient>);
-    await HarnessSetupNotice.FlushAsync(command, config, profiles, home);
+    await HarnessSetupNotice.FlushAsync(command, config, profiles, sp.GetRequiredService<HarnessRegistry>());
```
alexeyzimarev: **The registry is now resolved *outside* `HarnessSetupNotice.FlushAsync`'s fail-open `try` and *before* its `ShouldNotify` / `IsErrorRedirected` early returns.**

Pre-PR the line passed `home`, and `FlushAsync` built `HarnessRegistry.FromEnvironment(home)` inside its `try { … } catch { }` only after both guards. Now the argument expression runs in Program's outer `finally`, past both `catch` blocks and `CrashReporter`: anything thrown by `HarnessRegistry.FromEnvironment(home, probe)` escapes as an unhandled exception at exit (a NativeAOT abort) for every command, fail-open hooks included. Today every vendor ctor is pure `Path.Combine`, so there is no live trigger — it is a guard-placement regression, plus the registry is now built for every command (`kcap update`, `report-version`, `config`, …) even when the notice is suppressed, which the CommandServices comment two files over says must not happen.

Mirror the neighbouring `UpdateNotice.FlushAsync(..., Run<NpmRegistryClient>)`: take `Func<HarnessRegistry>`, pass `Run<HarnessRegistry>`, and invoke it after the guards inside the `try`.

src/Capacitor.Cli.Daemon/DaemonRunner.cs
```diff
@@ -1242,17 +1256,17 @@ internal static bool ShouldWarnCursorUnavailable(IEnumerable<IHostedAgentRuntime
...
-            config.GeminiUnattendedReviewerEnabled, config.GeminiPath);
+            config.GeminiUnattendedReviewerEnabled, config.GeminiPath, config);
 
         SeedReviewerAffirmation(
             stateDir, AcpVendorDescriptors.OpenCode.Vendor,
-            config.OpenCodeUnattendedReviewerEnabled, config.OpenCodePath);
+            config.OpenCodeUnattendedReviewerEnabled, config.OpenCodePath, config);
 
-        SeedVersionFloor(stateDir, AntigravityVendor, config.AntigravityPath);
+        SeedVersionFloor(stateDir, AntigravityVendor, config.AntigravityPath, config.Binaries);
```
alexeyzimarev: **This line reads `config.Binaries` unconditionally, so the comment added in `SeedReviewerAffirmation` below ("config.Binaries is read only past the guard above: a daemon with every reviewer disabled must not need it set") is false at the only entry point that matters.**

`Binaries` has a throwing getter; the Antigravity floor is deliberately not flag-gated (the method summary explains why), and the argument is evaluated before `SeedVersionFloor`'s `try`. So any `DaemonConfig` reaching `SeedReviewerFloors` without `Binaries` throws `InvalidOperationException` regardless of reviewer flags. Production sets it in the `RunAsync` initializer and the `Config()` helper in `DaemonRunnerAntigravityFloorTests` sets it too, so nothing exercises the promise — the next all-reviewers-disabled test written against the comment trips it.

Read `var binaries = config.Binaries;` once at the top of `SeedReviewerFloors`, change `SeedReviewerAffirmation`'s last parameter from `DaemonConfig` to `BinaryProbe` (matching `SeedVersionFloor`), and drop the comment.

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -1572,8 +1572,8 @@ public static bool RemoveCopilotHooks(string hooksPath) {
     const string KiroBinary    = "kiro-cli";
```
alexeyzimarev: **A tenth ambient probe survives, and it is a *vendor* binary, not `kcap`:** `InstallKiroHooks` (line 1778) does `if (!BinaryProbe.OnPath(KiroBinary)) return false;` and `RunKiroCli` (line 1853) spawns `new ProcessStartInfo(KiroBinary)` with the bare name, while the enclosing `PluginCommand(PluginEnvironment env)` holds `env.Harnesses`.

That is the pattern this PR removes everywhere else: the probe reads the process PATH (a registry built over a redirected search path never reaches it), and probe→spawn use two different resolution rules — `OnPath` walks PATHEXT while `CreateProcess` appends only `.exe`, the gap `BinaryProbe`'s own doc exists to close. The PR description's "nine `OnPath(\"kcap\")` sites stay" undercounts this one. Pre-existing, so not a regression, but it is in a function this PR rewrote around.

Note `harnesses.ResolveExecutable(HarnessId.Kiro)` is not a drop-in — Kiro declares `["kiro", "kiro-cli"]`, so the IDE launcher would win. Expose the probe (`BinaryProbe` on `PluginEnvironment`, or `HarnessRegistry.Binaries`), resolve `kiro-cli` through it, and spawn the resolved path.

src/Capacitor.Cli/Commands/PluginEnvironment.cs
```diff
@@ -42,18 +42,23 @@ public Func<IEnumerable<StaleAgentTarget>, IReadOnlyList<StaleAgentProcess>> Fin
...
+    /// the ambient read it exists to remove. Held as one instance, so two members of one vendor
+    /// cannot name different roots however the environment moves afterwards.
     /// </summary>
-    public required HarnessPaths Paths { get; init; }
+    public required HarnessRegistry Harnesses { get; init; }
 
-    public static PluginEnvironment FromProcess(ProfileConfig profiles, UserHome home) => new(
+    /// <summary>The cross-vendor <c>~/.agents</c> tree. Derived from <see cref="Home"/> rather than
+    /// supplied: it honours no override, so a second value could only disagree with this one.</summary>
+    public AgentsPaths Agents { get; } = new(Home);
```
alexeyzimarev: **`Agents` goes stale under `with`.** A get-only property initialised from the positional parameter is copied verbatim by the record's synthesized copy constructor — `with` re-runs no initialisers — so `env with { Home = other }` yields `Home == other` but `Agents` still rooted at the original home (same instance). It also has no `init` accessor, so a `with` block cannot correct it (CS8852). That is precisely the disagreement the doc sentence above ("a second value could only disagree with this one") rules out.

No caller uses `with` on a `PluginEnvironment` today, so this is latent — but the type exists as a test seam and `PluginCommand` installs/removes/sweeps under `env.Agents.UserSkillsDir`, so a rebound `Home` in a test would target the wrong tree (the developer's real `~/.agents/skills` if the original was `UserHome.FromEnvironment()`).

```suggestion
    public AgentsPaths Agents => new(Home);
```

src/Capacitor.Cli.Daemon/Harness/Claude/ClaudeLauncher.cs
```diff
@@ -30,9 +31,9 @@ public bool DisablesApprovalPrompts(LauncherContext ctx) =>
     /// equivalence key). Stateless singleton.</summary>
     public IReviewerModelResolver? ReviewerModelResolver => ClaudeReviewerModelResolver.Instance;
 
-    public bool IsAvailable() => CliResolver.Exists(CliPath);
+    public bool IsAvailable() => new CliResolver(config.Binaries).Exists(CliPath);
 
-    readonly ClaudePaths _paths = ClaudeHarness.FromEnvironment(home).Paths;
+    readonly ClaudePaths _paths = harnesses.Of<ClaudeHarness>().Paths;
```
alexeyzimarev, line 36: **`HarnessRegistry.Of<T>()` (`(TSelf)this[TSelf.Id]`, HarnessRegistry.cs:101) casts unchecked, and this PR moves it into a field initialiser here and in `CodexLauncher.cs:17`.**

A registry that carries a different `IHarness` implementation under that id — `TestHarnesses.All()` fills every id with the `TestHarness` record — throws `InvalidCastException` ("Unable to cast TestHarness to ClaudeHarness") instead of the façade's documented "No such harness" `ArgumentOutOfRangeException`; and because it now runs at construction, a subset or stub registry fails the launcher ctor before any method is called. In `WorkItemsNudgeAvailability.IsRegisteredFor` the surrounding `catch { return false; }` would swallow the same cast into a silent "not registered". No current test hits it (every launcher/nudge fixture uses `TestHarnesses.Under(home)`), so latent — but the new primitive should fail with its own message:

`this[TSelf.Id] as TSelf ?? throw new InvalidOperationException($"{TSelf.Id} in this registry is not a {typeof(TSelf).Name}")`.

src/Capacitor.Cli.Daemon/DaemonConfig.cs
```diff
@@ -158,6 +159,15 @@ public UserHome Home {
         set;
     }
 
+    /// <summary>The same PATH probe the DI-registered <c>HarnessRegistry</c> was built over — carried here
+    /// separately because a configured vendor path (<see cref="ClaudePath"/> and its siblings) is
+    /// not a harness question, so its resolvers (<c>CliResolver</c>, <c>VendorVersionResolver</c>)
+    /// take this rather than the registry.</summary>
+    public BinaryProbe Binaries {
```
alexeyzimarev, line 164: **A service (the PATH probe) on the config bag, behind a throwing getter, exists only so consumers can write `new CliResolver(config.Binaries)` inline — while the same probe is already a DI singleton.**

`DaemonRunner.cs:365` `builder.Services.AddSingleton(binaries)` has **no consumer**: no daemon type takes `BinaryProbe` and nothing calls `GetRequiredService<BinaryProbe>()`; every reader goes through `config.Binaries` instead. So there are three routes to one probe (DI probe, DI registry's private `_binaries`, `config.Binaries`), their identity is a `RunAsync` convention rather than a type guarantee (every test fixture that builds both a `DaemonConfig` and a `TestHarnesses.Under(home)` registry breaks it), and roughly 15 test-built `DaemonConfig`s carry no `Binaries` at all — a `[DaemonConfig.Binaries was never set]` throw waiting for the first fixture that routes a real launcher through `IsAvailable()`, `SeedReviewerFloors` or `HandleResolveReviewerModel`.

The doc justifies "this rather than the registry" but not "this rather than DI". Every consumer except two `DaemonRunner` statics is DI-constructed: register `CliResolver` (and `VendorVersionResolver`) as singletons over the injected probe, inject them, pass `binaries` to `SeedReviewerFloors`/`FingerprintUnattendedVendors` from the `RunAsync` local that already exists, and drop this property and its getter.

src/Capacitor.Cli.Daemon/Services/CliResolver.cs
```diff
@@ -3,26 +3,27 @@
...
+/// <see cref="BinaryProbe"/>; a second copy is how the Windows <c>.cmd</c> gap survived, so this
+/// must never grow its own. Used by <see cref="IHostedAgentLauncher.IsAvailable"/> at daemon startup
+/// to decide which vendor launchers to advertise over <c>DaemonConnect</c>.
 ///
 /// <para>This intentionally does NOT execute the binary — startup must stay cheap. False
 /// positives (binary present, exec bit set, but unusable for some other reason) surface
 /// later as the existing <c>LaunchFailed</c> path.</para>
 /// </summary>
-internal static class CliResolver {
+internal sealed class CliResolver(BinaryProbe binaries) {
```
alexeyzimarev: **After this change `CliResolver` is a two-method rename of `BinaryProbe`** (`Exists` ⇒ `binaries.Finds`, `ResolveExecutable` ⇒ `binaries.Resolve`, nothing else) — yet it is constructed **per call** at 8 sites (`ClaudeLauncher:34`, `CodexLauncher:52`, `AcpHostedAgentRuntimeFactory:120/897`, `PiRpcHostedAgentRuntimeFactory:65`, `AntigravityHostedAgentRuntimeFactory:108`, `VendorCliWatcher:45` — a fresh wrapper every 15 s poll tick — and `DaemonRunner:626`), and `VendorVersionResolver` likewise at 4. `VendorCliWatcher.StatCliBinary(CliResolver, string)` and `CursorBorrowedReviewValidation.TryMatchValidatedBuild(CliResolver, …)` then take the wrapper as a parameter type only so callers can pass one in.

Either delete the class and call `config.Binaries.Finds(...)` / `.Resolve(...)` directly (folding `CliResolverTests` into the Core `BinaryProbe` tests they duplicate), or register one `CliResolver` in `DaemonRunner` beside `binaries` and inject it — which also lets `VendorCliWatcher` hold it instead of the `StatBinary` lambda + static two-arg helper pair. `VendorVersionResolver` earns its type (it spawns `--version`) but wants the same single DI instance.

src/Capacitor.Cli/Commands/PluginCommand.cs
```diff
@@ -1295,7 +1295,7 @@ async Task ReportStaleAgentsAsync(
     }
 
     async Task<int> RemoveAntigravity(string[] args) {
-        var hooksPath = GetArg(args, "--antigravity-hooks-path") ?? env.Paths.Antigravity.GlobalHooksJson;
+        var hooksPath = GetArg(args, "--antigravity-hooks-path") ?? env.Harnesses.Of<AntigravityHarness>().Paths.GlobalHooksJson;
```
alexeyzimarev: **The mechanical rewrite left 101 `env.Harnesses.Of<X>().Paths.Y` chains where the deleted record gave one cached field read.** `Of<T>()` is `(TSelf)this[TSelf.Id]` → `ById` → `_harnesses.FirstOrDefault(h => h.Id == id)`: a closure allocation plus a linear scan per call. `RemoveAntigravity` (this method) does it 11 times for `AntigravityHarness`; `RemoveCodex` 6; `RegisterCodexMcpServersAsync` resolves the same `ConfigToml` three times in one `switch`. `kcap plugin` is one-shot so the runtime cost is nothing — the cost is that a reader re-parses the vendor on every line.

`SetupCommand.HandleAsync` in the same PR already shows the shape (eight `var codex = harnesses.Of<CodexHarness>().Paths;` locals at ~715-722): one local per method, or per-vendor cached accessors on `PluginEnvironment` so sites read `env.Antigravity.GlobalHooksJson` as they did with `env.Paths.Antigravity`.

src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
```diff
@@ -674,7 +675,7 @@ public AgentOrchestrator(
         _configRoot        = configRoot;
         _home              = home;
```
alexeyzimarev: **`_home` is dead after this PR.** Its only reader (`TitleGeneration.GenerateAsync(..., _home, ...)` at ~4884) now receives `_harnesses`; the field (line 526) is declared and assigned here and never read, yet the ctor still takes `UserHome home` and all 10 explicit `new AgentOrchestrator(` sites in the daemon tests supply it. Drop the field, the parameter and the argument at each site.

src/Capacitor.Cli.Daemon/Harness/Antigravity/AntigravityReviewerHome.cs
```diff
@@ -147,6 +149,14 @@ internal static string Create(string stateDir, string daemonEpoch, string launch
         return home;
     }
 
+    /// <summary>This home's Antigravity layout, derived from the home directory alone. No vendor
+    /// override is applied: <c>GEMINI_CLI_HOME</c> REPLACES the Gemini root, so honouring it here
+    /// would resolve every file below outside the isolated home — the reviewer's result channel and
+    /// its grants into the operator's own tree, and the plugin-directory check against a directory
+    /// this home does not own.</summary>
+    static AntigravityPaths LayoutFor(string home) =>
+        AntigravityHarness.Over(GeminiHarness.Over(new GeminiPaths(new(home), null))).Paths;
```
alexeyzimarev: **This builds a `GeminiHarness` and an `AntigravityHarness` (each with detection closures nothing here asks) only to unwrap `.Paths`.** `AntigravityPaths` already has the exact ctor for it (`AntigravityPaths.cs:26`: `public AntigravityPaths(UserHome home, string? geminiCliHome) : this(new GeminiPaths(home, geminiCliHome))`), so the whole method is `new AntigravityPaths(new(home), null)` — which also states the behaviour fix ("`null` override") in one token instead of a composition the doc comment then has to explain. (Check the `Gemini` using does not go unused — IDE0005 is a build error here.)

If a production "rooted at a home, no override" factory is wanted, the same shape is hand-rolled again in `TestHarnesses.Under`; an `XHarness.Under(UserHome)` per vendor would serve both.

test/Capacitor.Cli.Core.Tests.Unit/Harness/HarnessRegistryTests.cs
```diff
@@ -32,14 +37,85 @@ public async Task Every_harness_carries_a_label() {
...
+        using var gemini = EnvScope.Exclusive("GEMINI_CLI_HOME", relocated.Path);
+
+        var harnesses = HarnessRegistry.FromEnvironment(new("/fake/home"));
+
+        await Assert.That(harnesses.Of<AntigravityHarness>().Paths.McpConfigJson)
+            .StartsWith(harnesses.Of<GeminiHarness>().Paths.Root);
+    }
+
+    [Test]
+    public async Task Resolves_the_executable_a_vendor_declares() {
```
alexeyzimarev: **The PR's headline claim — a `Searching(probe)`-redirected registry now resolves executables on the *new* probe — has no test.** Every `ResolveExecutable` test here builds the registry directly over its probe via `Over(...)`; no test in the tree calls `HarnessRegistry.Searching(...)` and then `ResolveExecutable`. The implementation is right (`Searching` is `new(_harnesses, binaries)`), but a regression to `new(_harnesses, _binaries)` — or a future cached field — would pass the whole suite; the only `Searching` consumer (`AgentsStepViewModel`) calls `Detected` alone.

While here: `TestHarnesses.Over(BinaryProbe, params IHarness[])` is a verbatim pass-through of `HarnessRegistry.Over` (six callers could use the real one), and `BinaryProbe.Searching(null)` is spelled out at 26 new test sites plus twice inside `TestHarnesses` itself — `TestBinaries` is the existing home for test probes and has no "finds nothing" member; a `TestBinaries.None` would name the intent once.

src/Capacitor.Cli.Daemon/Harness/Antigravity/AntigravityReviewerHome.cs
```diff
@@ -214,26 +224,20 @@ static void WriteSettings(string home, IReadOnlyList<AcpMcpServerSpec> injected)
...
-    /// <para><c>AntigravityPaths</c> → <c>GeminiPaths.Root</c> honors THIS PROCESS's own
-    /// <c>GEMINI_CLI_HOME</c> when set, falling back to <paramref name="home"/> only when it is not — so
-    /// on a daemon that happens to run with <c>GEMINI_CLI_HOME</c> set, the resolved path would escape
-    /// <paramref name="home"/> entirely, writing the reviewer's result channel (or its permission
-    /// grants) outside the isolated home job 3 exists to guarantee, and potentially into the operator's
-    /// own Gemini tree. Not something to fix in <c>AntigravityPaths</c> (that fallback is correct for
-    /// its other, non-isolated callers) — this class's whole purpose is isolation, so it is verified
-    /// here rather than assumed, once for every file the home writes.</para>
+    /// Isolation is this class's whole purpose, so containment is verified for every file the home
+    /// writes rather than assumed from <see cref="LayoutFor"/> — a derivation that reintroduced a
```
alexeyzimarev: **Comment conventions (CLAUDE.md › Comments).** A few PR-added or PR-rewritten comments narrate the change rather than the constraint:

- Here, and in the mirrored test comment (`AntigravityReviewerHomeTests.cs:278` "the one thing standing between a *reintroduced* vendor override…"): "reintroduced" only parses once the reader knows an override derivation existed before this PR. "a derivation that honoured a vendor override would…" says the same thing timelessly.
- `CliResolver.cs:9` was rewritten but kept "a second copy is how the Windows `.cmd` gap survived" — history the reader cannot reconstruct from source; the constraint is just "the PATH/PATHEXT walk lives once, in `BinaryProbe`".
- `CommandServices.cs:29-33` and `DaemonRunner.cs:56-63` are near-twins stating the shared-PATH rationale twice, and "Neither stats disk here — that happens per call, lazily" / "it caches no detection answer" describe `BinaryProbe`/`HarnessRegistry` internals that belong on (and already sit on) those types.
- `AcpHostedAgentRuntimeFactoryTests.cs:1956-1964`: the new `static BinaryProbe Binaries` was inserted between `BrokeredEnv()`'s multi-paragraph `<summary>` and `BrokeredEnv()` itself, so the brokered-credential rationale now documents the probe and `BrokeredEnv()` is undocumented.

test/Capacitor.Cli.Tests.Unit/Commands/UninstallCommandTests.cs
```diff
@@ -8,13 +8,15 @@
...
 
 namespace Capacitor.Cli.Tests.Unit.Commands;
 
-// VendorEnvOverrides: uninstall resolves every vendor layout through HarnessPaths, so it reads
-// each override variable a peer suite clears.
-[NotInParallel(["VendorEnvOverrides", "CwdMutation"])]
+// CwdMutation: uninstall reads the process working directory to find the project git root, and
+// these tests move it. Every vendor layout comes from the registry they inject, so no vendor
+// override variable is read by any route.
+[NotInParallel("CwdMutation")]
```
alexeyzimarev: **Keyed `[NotInParallel("CwdMutation")]` for a class that moves the working directory.** CLAUDE.md › Test conventions › Parallelism: "Bare is what process-global state needs: an environment variable, `Console`, a mutable static in production code, **the working directory**." `Project_flag_strips_project_scope_entries` sets `Environment.CurrentDirectory = tmp.Path` (line ~368) under only this keyed cohort, so any concurrent test outside it that resolves a relative path or spawns a child reads the moved cwd. The pattern is pre-existing, but this PR rewrote the attribute and the comment justifying keyed-for-cwd; the same rewritten comment in `SetupCommandTests.cs:1013-1015` keeps three `HandleAsync` tests keyed for cwd, and the surviving `"VendorEnvOverrides"` cohort now mostly guards PATH mutation (`ImportSkipTitleTests`, `PluginCommandStaleAgentTests`, `PluginCommandVendorSkillsTests` set `PATH` under keyed scopes) — PATH is read by every `BinaryProbe.OnPath("kcap")` outside the cohort and inherited by every spawned child, so the doc says bare there too, and the key's name no longer describes what it guards.

alexeyzimarev (changes requested):
The refactor direction is sound: the DI chain checks out end to end, registry construction is pure, and the call sites are consistent. Both CI legs, both AOT publish checks and the Linear-id scan are green. Fifteen inline comments are posted; the verdict rests on the four below. Each is introduced by this PR and each is a small change.

**Fix before merge**

- `Program.cs:865`. The registry is now resolved in the outer `finally`, outside `HarnessSetupNotice.FlushAsync`'s fail-open `try` and before its `ShouldNotify` / `IsErrorRedirected` guards. On `main` it was built inside both. A throw from the factory would escape every catch and `CrashReporter`, and every command now builds it even when the notice is suppressed. Pass `Run<HarnessRegistry>` and resolve after the guards, as `UpdateNotice.FlushAsync` already does. https://github.com/kurrent-io/kcap-cli/pull/823#discussion_r3961534568
- `HarnessRegistry.ResolveExecutable`. It returns the first declared name, and for both two-name vendors that is the IDE launcher: `antigravity` before `agy`, `kiro` before `kiro-cli`. The daemon's own default is `agy` and `KiroPaths.ProcessName` is `kiro-cli`. The doc names Antigravity as the case this method serves, so a caller following it gets the GUI shim. Put the CLI name first and pin the order with a test, or narrow the doc. https://github.com/kurrent-io/kcap-cli/pull/823#discussion_r3961534441
- `PluginEnvironment.Agents`. `{ get; } = new(Home)` on a positional record is copied verbatim by the synthesized copy constructor, so `with { Home = other }` leaves `Agents` rooted at the old home, which is exactly the disagreement the adjacent doc says cannot happen. No caller uses `with` today. `=> new(Home)` closes it. https://github.com/kurrent-io/kcap-cli/pull/823#discussion_r3961534999
- `DaemonRunner.SeedReviewerFloors`. Line 1269 reads the throwing `config.Binaries` getter unconditionally for the Antigravity floor, so the new comment in `SeedReviewerAffirmation` ("a daemon with every reviewer disabled must not need it set") is false at its only production caller. Hoist the read or drop the comment. https://github.com/kurrent-io/kcap-cli/pull/823#discussion_r3961534673

**Follow-up, not blocking**

- The inherited-override scrub is Antigravity-only. The sandboxed borrowed Copilot launch rewrites `HOME`, `TMPDIR` and `KCAP_CONFIG_DIR` but still inherits `COPILOT_HOME`, which `CopilotHarness` treats as a whole-root override; OpenCode's reviewer inherits `OPENCODE_CONFIG`. Pre-existing, same failure class this PR fixes for Antigravity. Worth an issue. https://github.com/kurrent-io/kcap-cli/pull/823#discussion_r3961534266
- `AgentOrchestrator._home` is dead after this PR (field, ctor parameter, ten test call sites). Trivial, author's call whether here or later. https://github.com/kurrent-io/kcap-cli/pull/823#discussion_r3961535794
- The remaining comments are design and convention items: the Kiro `OnPath` probe left in `PluginCommand`, the unchecked cast in `Of<T>`, `CliResolver` and `DaemonConfig.Binaries` shape, the repeated `Of<X>().Paths` chains, the `LayoutFor` composition, the missing `Searching` to `ResolveExecutable` test, comment conventions, and the keyed `NotInParallel` on a class that moves the working directory.

src/Capacitor.Cli.Core/Harness/HarnessRegistry.cs
Inok: Fixed, but not by ordering the list. The repo owner rejected encoding meaning in list position: signals are signals, and a spawnable command is a different fact. Each harness now declares `CliBinary` (`IHarness`), and `ResolveExecutable` resolves that alone — it no longer walks the detection names, so there is no fallback to an IDE launcher. A machine with Kiro's IDE and no `kiro-cli` resolves null; Cursor now resolves `cursor-agent` though it probes for no name at all. `Signals.Binaries` became `LaunchSignal`, a predicate, so order carries nothing. Pinned by tests that stage both names for Antigravity and Kiro and assert the CLI wins, mutation-verified.

src/Capacitor.Cli.Daemon/Harness/Antigravity/AntigravityHostedAgentRuntimeFactory.cs
Inok: Both gaps confirmed and closed. `COPILOT_HOME` is dropped for a borrowed-snapshot Copilot child, and `OPENCODE_CONFIG`/`OPENCODE_CONFIG_CONTENT` on an OpenCode review launch — the two you named. Kiro needed nothing: it stamps `KIRO_HOME`, so an inherited one is overwritten; Cursor relocates nothing through a vendor variable.

I built the shared declaration-plus-helper you suggested, then removed it: three of its five declarations were inert, and the helper reduced to `foreach (var v in variables) environment.Remove(v)` — a dictionary utility whose name and doc claimed a policy it could not enforce. The removals are now inline at the three isolation sites, each in vendor-specific code. Each is mutation-verified, including hoisting the call out of its condition to prove an interactive launch still keeps its inherited value.

src/Capacitor.Cli/Program.cs
Inok: Fixed as suggested — `FlushAsync` takes `Func<HarnessRegistry>` and invokes it inside the `try`, after both guards, mirroring `UpdateNotice`.

One correction to the finding, from checking `main`: `main` already built the registry inside the `try` after both guards, so the escape you describe was introduced by this PR's first commit rather than being a pre-existing shape, and this restores `main`'s behaviour rather than improving on it. What the delegate does buy over a plain parameter is that a command with redirected stderr never builds the registry at all.

src/Capacitor.Cli.Daemon/DaemonRunner.cs
Inok: Fixed exactly as suggested: `SeedReviewerFloors` reads `var binaries = config.Binaries;` once at the top, `SeedReviewerAffirmation`'s last parameter is now `BinaryProbe` matching `SeedVersionFloor`, and the false comment is gone.

src/Capacitor.Cli/Commands/PluginCommand.cs
Inok: Fixed, and the count in the description was wrong in a different way than you read it: there were nine `OnPath` sites in `src/`, of which eight were `"kcap"` — I had folded the Kiro one into the kcap total. The description now says eight.

`InstallKiroHooks` resolves through the harness flow rather than an injected probe, since `kiro-cli` is a harness binary: `ResolveExecutable(HarnessId.Kiro)` is now a correct drop-in because it resolves the declared `CliBinary` only. `RunKiroCli` spawns the resolved path. That also fixes a live Windows bug — `OnPath` matched an npm `.cmd` shim via PATHEXT, then `CreateProcess` appended only `.exe`, so the clone failed right after the probe said yes.

src/Capacitor.Cli/Commands/PluginEnvironment.cs
Inok: Fixed as suggested — `public AgentsPaths Agents => new(Home);` — with a test pinning that `Agents` follows a rebound `Home` under `with`, mutation-verified against restoring the field initialiser. Nothing depended on reference identity: all reads are of `.UserSkillsDir`, and the one comparison is string equality.

src/Capacitor.Cli.Daemon/Harness/Claude/ClaudeLauncher.cs
Inok: Fixed. `Of<TSelf>()` now throws `InvalidOperationException` naming both the id and the expected type. (`as TSelf` does not compile — `TSelf` has no class constraint — so it is `is TSelf harness ? harness : throw`.) Pinned by a test that fails if the message drops either name.

The field initializers remain, and worth recording: `main` had the same two (`ClaudeHarness.FromEnvironment(home).Paths`), so construction-time evaluation is not new here — only the failure mode for a registry carrying a different implementation under that id.

src/Capacitor.Cli.Daemon/DaemonConfig.cs
Inok: Half taken, half declined — flagging the declined half explicitly.

Taken: `AddSingleton(binaries)` was indeed consumerless and is deleted. I verified it independently — the only `GetRequiredService<BinaryProbe>()` in the tree is the CLI's container, not the daemon's.

Declined: removing `DaemonConfig.Binaries` in favour of constructor injection. All nine `*Path` siblings have production readers beyond availability — spawning, version probing, `CodexMcpInventory`, `SeedVersionFloor`, the ACP descriptors, error strings — and there is a third override source the finding does not mention: `DaemonRunner` seeds `ClaudePath`/`CodexPath` from the persisted profile (`kcap config set daemon.claude_path`), so these carry operator configuration, not an ambient read. `Binaries` reaches daemon types the same way `Home` directly above it does. The doc now says that rather than only "not the registry". The duplication that was real — nine CLI-name literals repeated in the config bag — is gone: those defaults now reference `XHarness.CliBinary`.

src/Capacitor.Cli.Daemon/Services/CliResolver.cs
Inok: Deleted, taking the first of your two options. Every site calls `config.Binaries.Finds(...)` / `.Resolve(...)` directly, and `VendorCliWatcher.StatCliBinary` and `CursorBorrowedReviewValidation.TryMatchValidatedBuild` take `BinaryProbe`. `VendorVersionResolver` stays — it spawns `--version` and earns its type.

`CliResolverTests` was folded away only after checking each case against Core's `BinaryProbeTests`: five were exact duplicates, and the two that were genuine gaps (a rooted path that does not exist; a rooted executable that resolves) are now covered there, mutation-verified.

src/Capacitor.Cli/Commands/PluginCommand.cs
Inok: Fixed with the `SetupCommand` shape you pointed at — one local per method, no new `PluginEnvironment` members. 22 methods gained a local; the chain count went 101 → 35, of which 22 are the declarations and 13 are single-use sites where a local would be noise.

Verified as behaviour-preserving by comparing the multiset of vendor/member accesses on both sides: 26 keys in `PluginCommand`, 35 in `SetupCommand`, 11 in `UninstallCommand`, every count identical, no local reused across two vendors.

src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
Inok: Fixed — field, constructor parameter and all argument sites removed. The 11th site was `AgentOrchestratorHarness`, which subclasses and forwards through `base(...)` so it does not match `new AgentOrchestrator(`.

src/Capacitor.Cli.Daemon/Harness/Antigravity/AntigravityReviewerHome.cs
Inok: Fixed as suggested: `LayoutFor` is now `new AntigravityPaths(new(home), null)`. The `Harness.Gemini` using did go unused and was removed (IDE0005 is a build error here).

On the `XHarness.Under(UserHome)` suggestion — not taken. With `LayoutFor` collapsed to one constructor call, what is left is two small independent expressions with no shared invariant, which is the copy-the-primitive case rather than the hoist-a-helper one.

test/Capacitor.Cli.Core.Tests.Unit/Harness/HarnessRegistryTests.cs
Inok: Fixed — and you were right that it would have passed regardless. The new test builds a registry over one probe, calls `.Searching(probeB)` where probeB sees a binary the first cannot, and asserts `ResolveExecutable` answers on probeB. Mutation-verified: reverting `Searching` to pass `_binaries` fails it.

Also taken: `TestHarnesses.Over` is deleted in favour of `HarnessRegistry.Over`, and `TestBinaries.None` now names the finds-nothing probe at 24 sites. The six inside `BinaryProbeTests` are deliberately left spelled out — that file tests the probe's own null handling, so naming the helper there would hide the subject.

src/Capacitor.Cli.Daemon/Harness/Antigravity/AntigravityReviewerHome.cs
Inok: All taken. "Reintroduced" is gone from both the production doc and the mirrored test comment; `CliResolver.cs` no longer exists so its history sentence went with it; the `CommandServices` / `DaemonRunner` near-twins are cut to one line each stating only the shared-probe fact; and the `BrokeredEnv()` summary is reattached to `BrokeredEnv()` — a static probe field had been inserted between them.

The repo owner also called the new comments too long generally, so the whole diff got a second pass against the one-or-two-line norm.

test/Capacitor.Cli.Tests.Unit/Commands/UninstallCommandTests.cs
Inok: Fixed, and extended past what you flagged. `UninstallCommandTests` and the three `SetupCommandTests` are bare; the `VendorEnvOverrides` cohort is retired entirely, since all three remaining members mutated `PATH`. `ImportSkipTitleTests` also moved off a raw `Environment.SetEnvironmentVariable("PATH", …)` save/restore onto `EnvScope.Exclusive`, so the constraint is enforced rather than commented — that one had bypassed the check that would have named the problem.

One bonus: the method-level bare `[NotInParallel]` on `Project_flag_errors_when_not_inside_git_tree` was shadowing the class-level keyed one; removing it means the class's constraint now actually applies.

Cost measured before assuming: the CLI suite is unchanged at ~3m50s, the two 'after' runs bracketing the 'before' within noise.

PR #846, opened 2026-09-10
Show chat and composer for hosted sessions without a PTY

### LINKED ISSUE #839: Desktop app offers Pi but shows an empty chat for Pi sessions
## Symptom

The desktop app lists Pi as an available harness and launches it fine. The session then appears in the sidebar, but clicking it opens a workspace with no chat: a blank pane and the tab-strip note "This session has no terminal." The same session in the web UI shows the full conversation.

Seen on 2026-09-09 with kcap 0.11.40 / Pi 0.85.1, hosted Pi agents launched by the daemon.

## Root cause

The app never had a chat path for Pi. Three independent gates, each enough on its own:

1. **Chat is only built for PTY vendors.** `WorkspaceViewModel.cs:134` constructs `ChatTabViewModel` on the first status DTO that passes `HostedHarnessCatalog.ShowsTerminal`. `PiRpcHostedAgentRuntime.EmitsTerminalOutput` is `false` and the catalog classes `"pi"` as `"rpc"`, so `Chat` stays null, `ChatTabView` is never shown, and there is no banner template for the no-terminal case, hence the blank pane. `WorkspaceViewModelTests.cs:75-87` pins exactly this behaviour for vendor `"pi"`.
2. **No local Pi projection or chat rules.** `TranscriptChat.For` (`src/Capacitor.Cli.Core/TranscriptChat.cs:29`) and `TranscriptProjection.For` know only `claude` and `codex`; `src/Capacitor.Models.Transcripts/Harness/` has only `Claude/` and `Codex/`. Pi's canonical events are produced server-side by `PiTranscriptNormalizer`. With a null projection the tab would sit in `ChatTabPhase.Unavailable` ("No chat view for this harness").
3. **Nothing to tail.** The app's chat is a local JSONL tailer over `AgentStatusDto.TranscriptPath` (`ChatTabViewModel.SwitchPath`). A hosted Pi session has no transcript file the app could find: the runtime streams `AcpEventEnvelope`s from the daemon to the server over the `AcpSessionEvents` hub (`AgentOrchestrator.StartAcpForwardingAsync`), `PiLaunchEnvironment` sets `KCAP_PI_PURE=1` so the local extension stands down, and `AgentOrchestrator.DetectSessionIdAsync` has locators for `claude` and `codex` only, so the DTO carries no path.

The web works because it reads the server's canonical events, the only place a hosted Pi transcript exists. The app never consumes the daemon's envelope stream.

## Scope

Not Pi-specific. Every non-PTY vendor the app offers (cursor, copilot, gemini, kiro, opencode, antigravity, pi) renders the same empty workspace. Offering a harness the workspace cannot display is the user-facing bug; the missing chat path is the cause.

## Fix directions

- Preferred: have the app consume the daemon's `AcpEventEnvelope` stream over local IPC for non-PTY sessions, and relax the `ShowsTerminal` gate in `WorkspaceViewModel` so a chat tab is built without a terminal. That matches how these runtimes actually produce their transcript.
- Alternative: add a Pi leaf projection under `Capacitor.Models.Transcripts/Harness/Pi/` mirroring the server normalizer, `PiChatRules` under `Capacitor.Cli.Core/Harness/Pi/`, and a Pi arm in `DetectSessionIdAsync` so a file exists to tail. This only covers Pi and duplicates the server's normalizer.
- Until either lands, the app should not offer harnesses whose sessions it cannot display, or the workspace should say so and deep-link to the web view the way remote rows already do (`RailSessionViewModel.cs:67`).

realtonyyoung:
Closes #839 — AI-2625 (first slice of AI-2197)

## What & why

Hosted sessions without a PTY (Cursor, Copilot, Gemini, Kiro, OpenCode, Antigravity, Pi, Codex app-server) had no Chat tab and no way to send a follow-up from the desktop app: the chat needs a local transcript file, and only Claude and Codex wrote one. The daemon now appends every accepted envelope to a per-agent journal under its state directory and reports it as `transcript_path` with a new `transcript_format`; the app tails it with the same `JsonlTail` and renders it through a passthrough projection. Input rides a new one-shot local frame, `SendText`/`SendTextAck` (capability `input/1`), routed into the same delivery core the web's `SendInput` uses, so a local send and a web send behave identically. Design: `docs/superpowers/specs/2026-09-09-ai2197-non-pty-chat-and-composer-design.md`.

## Where to look

`TranscriptJournal`: `Record` is one `TryWrite` on the runtime's thread and never touches the disk; a hung filesystem costs journal lines, not protocol liveness. `HandleLocalSendTextAsync` answers only when the delivery settles — there is no reply timeout on this frame by design.

Two server-visible changes ride the shared delivery core: a failed borrowed-snapshot refresh now reports a `delivery_failed` drop where it reported nothing, and a runtime that refuses a write is logged at Warning with the drop reported to the sender rather than at Error with nothing sent.

## Verification

- `dotnet test --solution Capacitor.slnx`: 12464 total, 6 failed — all pre-existing on untouched files (`Installed_codex_schema_matches_the_vendored_pin`, the two 32 MB flood tests against a 10 s bound, load-sensitive timing tests).
- AOT publish (CLI and daemon): no IL2xxx/IL3xxx.
- Against an isolated daemon built from this branch: `HelloReply` advertises `input/1`; status rows carry `transcript_format`; `SendText` acks `no_such_agent`, `malformed`, `text_empty` and `Ok/delivered`. The journal path still needs a server-driven non-PTY launch to exercise end to end.

### DIFF
--- a/docs/superpowers/plans/2026-09-09-ai2197-non-pty-chat-and-composer.md
+++ b/docs/superpowers/plans/2026-09-09-ai2197-non-pty-chat-and-composer.md
@@ -1,0 +1,3965 @@
+# Non-PTY Chat and Composer (Slice A) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Every non-PTY hosted session (Cursor, Copilot, Gemini, Kiro, OpenCode, Antigravity, Pi, Codex app-server) gets a Chat tab that renders its transcript and a composer that sends follow-ups, in the desktop app, with parity to the PTY path wherever the transport allows.
+
+**Architecture:** The daemon appends every accepted `AcpEventEnvelope` to a per-agent JSONL journal under its state directory and reports the path as `transcript_path` plus a new `transcript_format`; the app tails it with the existing `JsonlTail` through a new passthrough projection. Input rides one new one-shot local frame (`SendText`/`SendTextAck`) routed into the same delivery core the server-origin `SendInput` uses. The app's composer gains an abstraction (`ChatInput`) with a terminal implementation for PTY and a frame implementation for everything else.
+
+**Tech Stack:** .NET 10 (NativeAOT), System.Threading.Channels, System.Text.Json source generation, Avalonia + ReactiveUI + DynamicData, TUnit, `Microsoft.Extensions.Time.Testing.FakeTimeProvider`.
+
+**Spec:** `docs/superpowers/specs/2026-09-09-ai2197-non-pty-chat-and-composer-design.md` (read it first; every task below cites the section it implements).
+
+## Global Constraints
+
+- Build, test and publish with `~/.dotnet/dotnet` (.NET 10); the `dotnet` on PATH is 8.0. Run one suite with `~/.dotnet/dotnet run --project test/<Suite>/<Suite>.csproj -- --treenode-filter '/*/*/<Class>/*'`.
+- `FrameType` is append-only: `SendText = 22`, `SendTextAck = 80`. Never renumber anything.
+- `LocalControlCapabilities.Current` gains `"input/1"` in the SAME commit as the `LocalControlServer` `case FrameType.SendText`.
+- `AgentStatusDto` gains exactly one trailing member `string? TranscriptFormat = null` (`transcript_format`), always emitted; values `"vendor"` / `"envelopes"`.
+- Journal path: `<DaemonStore.StateDirectory(config.Name)>/transcripts/<lowercase sha256 hex of agent id>.jsonl`. Journal channel: `BoundedChannelOptions(4096) { FullMode = Wait, SingleReader = true, SingleWriter = false }`. `Complete` grace 2 s. Path-lock bound 2 s. Retention 30 days, sweep every 24 h.
+- Frame limits: `text` over 256 KiB of UTF-8 → `too_large`. Reason tokens exactly as the spec table: `malformed`, `text_empty`, `too_large`, `no_such_agent`, `protected_kind`, `not_running`, `reaper_claimed`, `reaper_claimed_late`, `queue_full`, `stop_failed`, `delivery_failed`; client-side `transport`.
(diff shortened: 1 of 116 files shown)


realtonyyoung:
Response to the Qodo review (commits `803f5d2b`…`9123fb52`):

**Addressed**
- 1 (`CreatedFile` before the lock): computed under the path lock, immediately before the file open.
- 2 (composer updates off the UI thread): the daemon-status subscription now observes on the main-thread scheduler.
- 3 (env-sourced session ids strip dashes): `ArgParsing.ResolveSessionIdFromEnv` and `HarnessRequesterContext` delegate to `SessionIds.Canonical`, so an ambient `sess-1` and an explicit `sess-1` resolve identically; the stale "dashless" comments are restated.
- 5 (`JsonValueKind` checks): `EnvelopeJournalFormat.TryRead` uses `JsonElementExtensions`.
- 8 (sweep tests build paths by hand): `TempDir` helpers throughout; the shadowed-directory case asserts, and a new case faults enumeration itself.
- 10 (sweep lock ignores shutdown): the sweep's token is passed to the lock acquire.
- 11 (`Dispose` leaves the writer parked): `Dispose` completes the queue and cancels, idempotent against `CompleteAsync`.
- 12 (torn-write test truncates every append): only the gap-bearing buffer is truncated; the preceding record is asserted intact.
- 14 (race test accepts its own timeout): the test asserts the ack settled and faulted.

**Not changed, with reasons**
- 4 (30-day sweep vs a live idle journal): the sweep joins on the PID record by design; a session idle for 30 days is outside the retention policy's intent. Filed as a follow-up rather than special-cased here.
- 6 (`InputIpc.cs` holds several types): it follows the existing `PermissionIpc.cs`/`ConsentIpc.cs` registry shape for one frame family's wire contracts.
- 7 (`new TempDir()` per test): an allowed form in this repo's test conventions; injection is preferred when every test needs one and can follow.
- 9 (open failure leaves the chat waiting): the spec's stated fault behaviour — `transcript_path` null with `transcript_format: "envelopes"` renders Waiting, the same phase a failed PTY discovery produces.
- 13 (same-id relaunch keeps the first projection): the spec scopes the app's same-id relaunch story out of this slice.
- 15 (composer sends bypass the mutation lane): the lane serializes daemon lifecycle mutations; per-agent input follows the permission-resolve precedent of calling the ops client directly.

Issue #859, opened 2026-09-10
Service-installed daemon takes minutes to start, then hangs before binding its socket while status reports running


realtonyyoung:
`kcap daemon service install --name tony` (kcap 1.0.0, `99128af0`, macOS 25.6, 2026-09-10 ~10:19) took minutes to return, and the daemon it started is still not connected to the server ten minutes later: neither the Capacitor web UI nor the desktop app shows it, yet `kcap daemon status` and `kcap daemon service status` both say **running**.

## What the process is doing

launchd job `io.kurrent.kcap.daemon.tony`: `state = running`, `runs = 1`, pid 87184, spawned 10:19:50, never exited.

- `daemons/tony.lock`, `tony.pid`, `tony.version` and `daemons/tony/coverage.json` were all written at **10:21** — about 70 s after spawn — so the daemon got through `DaemonLock.TryAcquire`, the boot checks and the coverage `RecordBoot`.
- `daemon-tony.log` is open for write (fd 8) but has **not received a single byte**: mtime 09:04:06, last line is the previous incarnation's "AppDomain.ProcessExit fired". `daemon-tony.out.log` (launchd stdout/stderr) likewise untouched since 09:04. Whatever the daemon is doing, it is before or outside the host's logging.
- **No control socket is bound**: `lsof -U -p 87184` lists nothing; `daemons/tony.sock` is the stale 09:06 file and a connect to it returns `ECONNREFUSED`. So `LocalControlServer` never started, and nothing that reads the socket (app, CLI Hello probe) can reach it.
- `ps`: state S, 0.70 s CPU total, RSS 11 MB. Idle, not spinning. A `sample` shows the main thread and one thread-pool worker parked deep in stripped frames (release binary, no symbols), so the blocking call cannot be named from outside.

So the daemon is stuck somewhere between `RecordBoot` and the first `ILogger` line — `builder.Build()`, `host.StartAsync`, or the guarded startup's first awaits — with no log line to say which.

## Why the status commands lie

`kcap daemon status` reports running from the validated pid, and `service status` from the launchd job. Neither reflects that the daemon is not serving: no socket, no server connection. A daemon that has held the lock for minutes without binding its socket should be reported as *starting/stuck*, not *running*.

## Context that may matter

[...]

PR #869, opened 2026-09-10
Drain the version probe's pipes while waiting so a chatty --version cannot deadlock startup

realtonyyoung:
Refs #859 — AI-2684

## What & why

The daemon's startup vendor `--version` probe deadlocks on any CLI that prints more than the OS pipe buffer. `ProbeCliVersionOnce` redirected both stdout and stderr but called `WaitForExit(timeout)` before reading either, so a child that writes ~16-64 KB+ (some vendor CLIs emit a banner or an "update available" notice) blocks on the full pipe while the parent blocks in the wait — neither side moves until the per-attempt timeout kills the child, then it retries three times, across several probe sites. That is the multi-minute silent daemon startup: observed live as `agy --version` alive 75+ seconds as a daemon child though it returns in ~1 s standalone. The probe now drains both streams while the child runs, so it exits promptly and the version is read.

This is the pipe-deadlock cause only. The other AI-2684 items — moving the probes off the startup critical path so a genuinely hung CLI cannot delay the socket bind, a startup phase trace for the pre-host window, and `kcap daemon status` distinguishing pid-alive from serving — stay open on the issue.

## Where to look

`DrainToEndAsync` swallows read faults to `""`, so the abandoned timeout path (kill then return) leaves no unobserved task exception behind the two in-flight reads.

## Verification

- New `DaemonRunnerVersionProbeTests.A_version_whose_output_exceeds_the_pipe_buffer_is_drained_not_deadlocked`: a POSIX stub floods 500 KB to stderr before its version line. Before the fix the probe hit its 3 s timeout and returned null; after, it returns `9.9.9`. Red-then-green confirmed.
- Full `Capacitor.Cli.Daemon.Tests.Unit` suite green; daemon AOT publish clean.

### DIFF
--- a/src/Capacitor.Cli.Daemon/DaemonRunner.cs
+++ b/src/Capacitor.Cli.Daemon/DaemonRunner.cs
@@ -1560,14 +1560,31 @@ public static partial class DaemonRunner {
                 RedirectStandardError = true,
                 ArgumentList = { "--version" }
             });
-            if (process is null || !process.WaitForExit(timeoutMs)) {
-                try { if (process is not null) ProcessTree.Kill(process); } catch { }
+            if (process is null) return null;
+
+            // Drain both pipes WHILE the child runs. A --version that writes more than the OS pipe
+            // buffer — some vendor CLIs emit a banner or an "update available" notice — blocks on
+            // write until the parent reads, so reading only after WaitForExit deadlocks: the child
+            // parks on a full pipe, the wait never returns, and the probe hangs until the timeout
+            // kills it. The drains never fault (they swallow to ""), so the abandoned timeout path
+            // leaves no unobserved task exception.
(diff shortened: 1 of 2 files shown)


realtonyyoung:
Ran a Codex review of this branch (`codex exec review --base main`). One finding, now addressed:

- **Unbounded probe buffers (P2).** The concurrent drain kept every byte in a string, removing pipe backpressure, so a CLI that floods `--version` could allocate without limit for the whole probe budget. Fixed by bounding retention to an 8 KB prefix (the version is at the front) while still reading each stream to EOF unretained, so the child never blocks. New test `A_version_at_the_front_survives_a_trailing_flood_on_the_same_stream` guards both halves — it fails if the drain stops at the cap.

No other findings.

PR #907, opened 2026-09-11
Inject the telemetry facade instead of reaching process-wide statics

### LINKED ISSUE #890: Inject the telemetry facade instead of reaching process-wide statics
Telemetry reaches the process through a static facade — `CliTelemetry` (seven
mutable statics behind a `Reset()` whose own doc says every toucher must call it
first), plus `SetupJoin` and `McpTelemetry`. Call sites name those statics from
`Program`, `ConfigCommand`, the MCP servers, the setup/login funnel and the
desktop wizard.

The cost lands on the test suite. Seventeen classes are serialised to keep the
statics apart, six of them on the assembly-exclusive lane. The keyed cohort
guarding the other eleven is unsound: a keyed constraint holds only when every
*reader* joins it, and the funnel's callers do not — so an outsider captures into
whichever sink a concurrent cohort member installed. One of the six says as much
in its own remarks, which is the tell that this was diagnosed once and paid for
rather than fixed.

Construct the facade at each entry and inject it, as `ConfigRoot`, `UserHome`,
`MachineAuth` and the clock already are. `SetupFunnel` and `SetupJoin` become
instances reached through it; `Reset()` and `TestSink` are deleted rather than
injected. It does not split across PRs — a surviving static writer needs a static
handle to the instance, and that handle keeps the cohort alive.

Out of scope: `TelemetryState` and `TelemetryDeviceId` are already pure over a
`ConfigRoot`, and `TelemetrySettings`/`TelemetryEnvironment` already take an
injected environment.

Inok:
Closes #890 — AI-2705

## What & why

Telemetry reached the process through a static facade: seven mutable statics behind a
`Reset()` whose own doc said every toucher must call it first, plus `SetupJoin` and
`McpTelemetry`. Seventeen test classes were serialised to keep those apart, and the keyed
cohort guarding eleven of them was unsound — the funnel's callers read the same sink and
never carried the key. The facade is now a value built at each entry and injected, with
`SetupJoin`/`SetupFunnel` as instances it owns; `Reset()` and `TestSink` are deleted rather
than injected, so there is nothing left to keep apart.

## Where to look

`Start` and `Announce` are deliberately separate: resolving the facade from a container must
not print a privacy disclosure or consume the once-per-device marker. Suppression rides in
`TelemetryStartup` because the spawn marker is removed as it is read, so a second facade in
the same process can only inherit the decision, never re-take it. The desktop wizard is handed
a facade that is off — it shows no notice and offers no opt-out, so its funnel must stay inert.

The second commit is unrelated to telemetry and touches no telemetry file. Those six tests were
winning on timing rather than sequencing, and they fail on `main` too.

## Verification

Core 3263/3263 · CLI 4020/4020 · App 1739/1739 · the two daemon classes 40/40.
`dotnet publish -c Release`: no IL2026/IL3050.

Each assertion rewritten here was checked by mutation, and every mutation reverted:

| mutation | result |
| --- | --- |
| `startup.Debug` never reaches the facade | 3 debug tests fail, including one that previously could not |
| `startup.Suppressed` ignored | both spawn-marker tests fail |
| `SetupJoin` state back to `static` | 10 join tests fail |
| a throwing sink's exception escapes the writer | the rewired IPC test fails |
| losing exceptions reported twice | 4 IPC tests fail |
| the reader stops draining after the first line | both flood tests fail at the 10s timeout |

### DIFF
--- a/src/Capacitor.App/App.axaml.cs
+++ b/src/Capacitor.App/App.axaml.cs
@@ -1,3 +1,4 @@
+using Capacitor.Cli.Core.Telemetry;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Reactive.Subjects;
@@ -51,6 +52,11 @@ public partial class App : Application {
     readonly ProfileOverrides _serverEnv  = ProfileOverrides.FromEnvironment();
     readonly MachineAuth      _machineEnv = MachineAuth.FromEnvironment();
     readonly UserHome   _userHome = UserHome.FromEnvironment();
+
+    /// The wizard signs in through the CLI's own stack, which reports the signup funnel. The app
+    /// shows no privacy notice and offers no opt-out of its own, so it hands that stack a facade
+    /// that is off — nothing a wizard run does can emit.
+    readonly CliTelemetry _telemetry = CliTelemetry.Disabled();
 
     /// Process-lifetime rather than per wizard run — a provisioning poll can outlive the window that
     /// started it, and this client degrades a transport failure but not a disposed handler. What it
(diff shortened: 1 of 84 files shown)


realtonyyoung (approved):
NO FINDINGS

PR #914, opened 2026-09-12
Resolve the auth endpoints at the composition root instead of per read

### LINKED ISSUE #913: Resolve the auth proxy and provisioning endpoints at the composition root
`AuthProxyEndpoint.Url` and `ProvisioningEndpoint.Url` are static properties that
read `KCAP_AUTH_PROXY_URL` and `KCAP_SIGNUP_URL` from the environment on every
read. Twelve production sites reach them that way, across the onboarding facade,
the login flow, `kcap setup`, `kcap login`, `kcap machine`, the join's first-hop
URL and the desktop wizard.

Two costs. The value can change under a single run, since nothing pins it — the
process that resolves an endpoint and the process that reports it to the operator
can answer differently. And a test that wants the override has to set a process
-wide variable: `AuthProxyEndpointTests` guards one with a keyed `NotInParallel`,
which is unsound, because every other reader of that property sits outside the
key.

Resolve both at the composition root and inject them, as `ConfigRoot`, `UserHome`,
`MachineAuth` and the clock already are. `MachineAuth.TokenUrlOverride` is the
shape to copy: the override is carried as data and the URL derived from it, so the
static classes go rather than gaining a setter.

Out of scope: `KCAP_URL` and `KCAP_PROFILE`, which `ProfileOverrides` already
resolves as an injected value.

Inok:
Closes #913 — AI-2730

## What & why

`AuthProxyEndpoint.Url` and `ProvisioningEndpoint.Url` were static properties reading
`KCAP_AUTH_PROXY_URL` and `KCAP_SIGNUP_URL` on every access, so the value could move under a
single run and a test wanting the override had to set a process-wide variable. One
`AuthEndpoints` record now carries both overrides as data, resolved once at each entry —
beside `MachineAuth.FromEnvironment()` in the CLI and in the desktop app's field initializers
— and injected from there. Same shape as `MachineAuth.TokenUrlOverride`.

The join's first-hop host rides on `TelemetryStartup` rather than widening `CliTelemetry.Start`,
which would have handed a signup host to eight MCP servers that can never mint a join.

The second commit is unrelated to the endpoints: it fixes a race where `SessionAccessService`
arms its retry timer one line after publishing the state a test waits on, so a single
`FakeTimeProvider` advance could land in the gap and fire nothing.

## Where to look

`OnboardingFacade` goes from twelve constructor parameters to thirteen, inserted before
`beforeCommit` rather than appended.

`AuthProxyEndpointTests` guarded a process-wide variable with a keyed `[NotInParallel]` while
every other reader of that property sat outside the key. Its replacement touches no
environment; the one test that does reaches for `EnvScope.Exclusive`, which enforces the bare
constraint the old test evaded.

## Verification

Core 3293, CLI 4064, App 1894 — zero failures. Solution builds at 0 warnings; AOT publish
emits no IL2026/IL3050.

Mutations, each reverted:

| mutation | caught by |
| --- | --- |
| override ignored | 2 endpoint tests |
| trailing slash kept | 1 endpoint test |
| `SignupUrl` reads `ProxyOverride` | 2 endpoint tests |
| the two variable names swapped | `FromEnvironment_reads_each_endpoints_own_variable` |
| `services.AddSingleton(endpoints)` removed | both `Every_registered_command_resolves` cases |
| `startup.SignupUrl` replaced by the default | `FirstHopUrl_targets_the_started_host_with_key_and_port` |
| retry timer never armed | 2 session-access tests |


realtonyyoung:
NO FINDINGS

realtonyyoung (approved):

PR #931, opened 2026-09-14
Inject the test-only statics that forced assembly-wide exclusion

### LINKED ISSUE #930: Test-only statics on production types force assembly-wide test exclusion
Several production types carry mutable statics that exist so a test can substitute a collaborator or instrument a method's interior. Because a static is process-global, every test that writes one has to take the whole assembly to itself — bare `[NotInParallel]` — and clear it in a `finally` that a new test can forget.

The affected members:

| Type | Members |
| --- | --- |
| `WatcherManager` | `ProcessStarterForTesting`, `SpawnOverrideForTesting` |
| `SkillsAutoSync` | `ProcessStarterForTesting` |
| `WorktreeManager` | `SnapshotFailurePoint`, `SnapshotRollbackHook`, `SnapshotPreClaimHook`, `SnapshotPostClaimHook` |
| `ServiceTxnMarker` | `FlushDirectory` |

Each should arrive as a constructor argument with a real production implementation, so the substitution is scoped to the object under test and the exclusion can go.

Two things this does not cover:

- `ServiceTxnMarker` is a fully static class with call sites across `ServiceVerify` and `DaemonServiceCommands`, so converting it means reshaping a power-loss durability path. Its exclusion is one method, not a class.
- `WorktreeManager`'s hooks are rendezvous points for a concurrency test, not collaborator substitution. Injecting them scopes the state per caller but leaves a seam in production. Removing it entirely would mean extracting the file-claim protocol into a type the tests can drive step by step.

Inok:
Closes #930 — Linear id pending: the issue was opened for this PR, so its auto-import has no id yet.

## What & why

Six mutable statics existed so tests could substitute a collaborator or hold a caller mid-method. Each is process-global, so every test writing one took the whole assembly to itself and cleared it in a `finally`. They now arrive as constructor arguments with real production implementations: `IProcessStarter`, `IWatcherSpawner` (plus `WatcherPaths` and `WatcherSpawnRequest`), and `ISnapshotBarrier`.

`WatcherManager` was hand-built at seventeen sites, so it moved into the container first — which made the parameters that existed only to build one unread, and the compiler named them.

## Where to look

`ISnapshotBarrier` scopes the claim-protocol rendezvous per caller but does not remove the seam: production still carries an interface whose only non-trivial implementation is a test's. Removing it means extracting the file-claim protocol into a type tests can drive step by step — a reshape of a concurrency-critical daemon path, noted on the issue rather than attempted here.

Six exclusions are gone; the four claim tests keep one, downgraded from bare to a class key. Unconstrained they starve the daemon suite's port and journal tests, so the key throttles real resource contention rather than standing in for exclusion — sound only because none of the state they share is process-global.

`WatcherPaths` resolves `KCAP_WATCHER_DIR` once instead of on every path-helper call, so the probe, the spawn and `kcap cleanup` share one owner for the directory they must agree on.

`SetupFacadeFactoryTests` is new: it pins the picker and requested workspace, which reach the façade only as constructor arguments nothing downstream reports.

## Verification

`dotnet build Capacitor.slnx` at 0 warnings; `dotnet publish -c Release` clean of IL2026/IL3050 for both AOT binaries.

| Suite | Result |
| --- | --- |
| Capacitor.Cli.Tests.Unit | 4078/4078 |
| Capacitor.Cli.Daemon.Tests.Unit | 3192/3192 |
| Capacitor.Cli.Tests.Integration | 286/286 |

Eight mutations, each failing the test that claims the behaviour: watcher-spawn URL guard removed; detached flag dropped from the session-end hand-off; both `IsWatcherAlive` checks removed from `EnsureWatcherRunning`; Cursor's spawn precedence removed; `--auto` dropped from the skills spawn; the snapshot claim weakened from `CreateNew` to `Create`; the requested workspace dropped; the given picker ignored.

The claim tests' parallelism was measured rather than assumed — full daemon-suite runs, counting runs carrying a failure beyond the local codex-pin skip:

| `StandaloneSnapshotTests` | Runs | With an extra failure |
| --- | --- | --- |
| Bare exclusions | 2 | 0 |
| Unconstrained | 5 | 3 |
| Class key | 3 | 0 |


src/Capacitor.Cli/Commands/Harness/KiroHookCommand.cs
```diff
@@ -328,7 +327,7 @@ HookBudget budget
         // transcript (startup is idempotent and agentSpawn fires again next prompt); a killed hook
         // costs the whole session's injection. An abandoned in-flight spawn reconciles on that firing.
         try {
-            await _watchers.EnsureWatcherRunning(sessionId, transcriptPath,
+            await watchers.EnsureWatcherRunning(sessionId, transcriptPath,
```
realtonyyoung, line 329: The comment immediately above still says `SpawnWatcher` performs the parent-PID resolution, but this PR removes that method and moves the behavior to `ProcessWatcherSpawner.SpawnAsync`. Please update the reference so it names the current source. The watcher-directory comment in `CleanupCommand` has the same stale `SpawnWatcher` reference.

realtonyyoung (commented):
Static review only; build and tests were not run as requested.

src/Capacitor.Cli/Commands/Harness/KiroHookCommand.cs
Inok: Both fixed in 1c2eee05 — thank you, and good catch on the `CleanupCommand` one, which I would have missed.

`KiroHookCommand` now names `ProcessWatcherSpawner` as what resolves the kiro-cli pid and passes it as `--parent-pid`. `CleanupCommand` no longer names a method at all: the constraint is that a spawn writes pid files wherever `KCAP_WATCHER_DIR` points, and cleanup has to read that same place, so the comment says that and points at `WatcherPaths` as the one resolver both go through.

I swept the branch for the same defect and found two more — `SpawnWatcher_never_starts_a_process` and `SpawnWatcher_records_the_process_start_token_beside_the_pid`, both named after the deleted method while calling `SpawnAsync`. Renamed in the same commit. `ShouldSpawnWatcher` and `MaybeSpawnWatcherAsync` are live Cursor methods and stay as they are.

test/Capacitor.Cli.Tests.Unit/Harness/Cursor/CursorWatcherSpawnTests.cs
```diff
@@ -120,34 +105,29 @@ public async Task NonQuarantined_toplevel_spawn_invokes_the_watcher_manager_keye
...
+        var parent = NewSessionId();
+        var childFile = tmp.PathTo($"{child}.jsonl");
+        await File.WriteAllTextAsync(childFile, """{"role":"assistant","message":{"content":[]}}""" + "\n");
+        Markers.Quarantine(parent, "test");
+
+        // Seed the ack so the no-ack gate cannot be what suppresses the spawn — otherwise
+        // this test would pass for the wrong reason and prove nothing about quarantine. Then
+        // drive a NON-lifecycle hook: the self-heal spawn path a real child actually reaches.
+        // (This previously used a child `sessionStart`, an event a real Cursor subagent child
+        // never fires.)
```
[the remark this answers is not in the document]
Inok: Fixed in 50540f39.

Worth noting for the record: all the flagged text predates this branch. Collapsing the `try`/`finally` blocks around the removed spawn override re-indented those lines, which is why the diff attributes them to this PR. Rewriting them is expected work for code the branch touches either way, so they were rewritten rather than left.

Seven blocks in that file now state the invariant instead of the history — a real Cursor subagent child emits no `sessionStart`, and transcript delivery stays gated on the durable acknowledgement marker rather than on an empty backlog — with the design-document paths dropped. Five more comments elsewhere on the branch got the same treatment.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
The follow-up resolves the stale watcher-spawn references. Static review only; build and tests were not run as requested.

Issue #781, opened 2026-09-05
Daemon test suite: a test-owned temp tree vanishes mid-test on the ubuntu leg


alexeyzimarev:
The daemon unit suite has a cross-test race on `ubuntu-latest`: a `TempDir` that one test owns disappears while that test is still using it. Four hits in three days, each on a different, unrelated PR or merge, and every re-run passed.

| Date | Run | Test | Symptom |
|---|---|---|---|
| 2026-09-03 | 33783996964 (PR #761) | `BorrowedReviewContextTests.Aggregate_capacity_accepts_exact_limit_and_declares_one_extra_byte_as_omitted` | `DirectoryNotFoundException` from `WorktreeManager.WriteOwnerOnlyFileAsync` writing `manifest.json` into its own `.preparing-<guid>` dir, created moments earlier in the same call |
| 2026-09-04 | 33861186946 (PR #763) | `BorrowedReviewContextTests.Omitted_oversized_config_does_not_consume_capacity_needed_by_later_configs` | `git cat-file -t <sha>` failed: `not a git repository` — the source repo's `.git` vanished mid-snapshot |
| 2026-09-04 | 33888139389 (PR #770) | `BorrowedReviewContextTests.Aggregate_capacity_…` again, plus `AgentOrchestratorBorrowLaunchTests.Borrowed_Cursor_review_flow_runs_in_owned_dirty_snapshot_and_refreshes_between_rounds` | `GitRepo.Create` → `git config user.name` died with `fatal: Unable to read current working directory`, so the fresh `TempDir` was gone before `git init` ran; the orchestrator test threw a `NullReferenceException` at `AgentOrchestratorBorrowLaunchTests.cs:93` in the same run |
| 2026-09-05 | 33967780043 (merge of #766) | `BorrowedReviewContextTests.Omitted_oversized_config_…` | same class, main branch |

The victims are all in the borrowed-review area, but the vanishing point moves: before `git init`, mid-snapshot, mid-manifest-write. That shape says something concurrent removes a tree it does not own, not that these tests are wrong.

[...]

PR #948, opened 2026-09-14
Replace ambient process state with values resolved once

Inok:
Refs #781 — AI-2528

## What & why

A fixture deleted from outside its owner surfaces as a `DirectoryNotFoundException`
wherever the victim next touches its tree, arbitrarily far from whatever removed it.
`TempDir` now reports that at dispose, the one place that can still name the owner —
which immediately caught orchestrator tests handing production cleanup their whole
fixture root instead of a directory under it.

Two ambient reads turned up the same way. The provider router's memo was a mutable
static that concurrent tests cleared under each other, so one test's write could
answer another's probe. The working directory was read from the process wherever it
was wanted. Both are resolved once now and injected; ambient cwd joins
`BannedSymbols.txt`, so a new read is a build error rather than a convention.

## Where to look

The two git children behind `ResolveForRepo` set no working directory of their own,
so they inherited the process's. Handing the value down without also handing it to
them would have changed nothing.

One test still moves the process: a path with no directory component resolves against
nothing else.

`GitProviderRouter` is a required constructor argument on the import sources, so
kcap-server's `PiAi892ImportE2ETests` stops compiling when it advances this submodule.
A matching change there is expected rather than a compatibility overload here.

## Verification

`dotnet test --solution Capacitor.slnx` — 13012 tests, 12943 passed, 68 skipped, 1
failed. That failure, `Installed_codex_schema_matches_the_vendored_pin`, reproduces
unchanged on `main` (installed codex 0.154.0 against a pin taken from 0.147.0) and is
`Skip.When`'d on CI, which has no codex.

Rewritten assertions checked by mutation rather than by going green — pointing the
injected directory away from the repo fails 2/3 `ResolveForRepoTests`, 2/2 Uninstall
`--project` tests and 3/4 Setup acceptance tests; dropping the Cursor workspace-root
guard fails its test. The survivors are the rows whose docs already state they are
directory-independent.

`dotnet publish -c Release`: no IL2026/IL3050. Assembly-exclusive tests 517 → 499.


src/Capacitor.Cli/Harness/Pi/PiImportSource.cs
```diff
@@ -32,10 +33,11 @@ internal sealed class PiImportSource : IImportSource {
     public PiImportSource(
         ConfigRoot                              config,
         string                                  sessionsDir,
+        GitProviderRouter                        router,
```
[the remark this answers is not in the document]
Inok: Noted; the server side will be updated.

test/Capacitor.Cli.Core.Tests.Unit/Harness/Codex/CodexConfigTomlTests.cs
```diff
@@ -125,7 +125,8 @@ public async Task EnableNetworkAccess_is_idempotent() {
     public async Task EnableNetworkAccess_writes_when_config_path_has_no_directory_component() {
         // GetDirectoryName("config.toml") is empty; CreateDirectory("") would throw and
         // silently turn the write into Change.Failed without the guard.
-        using var tmp   = new TempDir();
+        using var tmp = new TempDir();
+#pragma warning disable RS0030 // a path with no directory component resolves nowhere else
```
realtonyyoung: [P2] Fully isolate the remaining cwd mutation

`[NotInParallel("CwdMutation")]` only serializes tests that share that constraint key; this is the only occurrence in the repository, so this test still runs alongside every unconstrained Core test while the block below redirects the process-wide current directory. Any peer doing relative I/O or starting a child without an explicit `WorkingDirectory` can therefore observe (or act on) `tmp.Path`, recreating the ambient-state race this PR is intended to remove. Please use bare `[NotInParallel]` so this test runs completely alone, or move this case behind a subprocess/explicit base-path seam.

realtonyyoung (changes requested):
Static review only; I did not build or run tests. I found one remaining process-wide test-isolation issue.

test/Capacitor.Cli.Core.Tests.Unit/Harness/Codex/CodexConfigTomlTests.cs
Inok: You're right that the key was doing nothing — "CwdMutation" appears exactly once in the repo, so the cohort was this test alone and every unconstrained Core test ran alongside it.

Digging into it, though, the test could not reach the branch it was named for. Update canonicalises the path at CodexConfigToml.cs:461 (`configPath = CanonicalConfigPath(configPath)`, i.e. Path.GetFullPath) about forty lines before the guard at :498 reads `Path.GetDirectoryName(configPath)`. By then the path is always absolute, so GetDirectoryName is never empty. I checked by deleting the guard outright and making the create unconditional: all 43 tests in the class stayed green, including this one.

So moving the process working directory was only giving Path.GetFullPath a base — it exercised nothing. I've deleted the test and the two comment lines above the guard that claimed a directory-less path could arrive there. That drops the last `Environment.CurrentDirectory =` in src/ or test/, so nothing in the suite moves the process any more, which seemed a better answer than isolating it with a bare constraint.

One thing I looked at and deliberately did not change: a relative CODEX_HOME is honoured today and resolved against the process directory. Rejecting it would make kcap write to ~/.codex while Codex still reads the relative path, so I left that alone. It is not unique to Codex either — nine env vars become path roots and only HOME is rooted-checked. Separate concern from this PR.

realtonyyoung:
NO FINDINGS

realtonyyoung (approved):
Static review complete; no actionable findings.

PR #950, opened 2026-09-15
Improve desktop rail feedback for selection, launches and pull requests

realtonyyoung:
No tracker issue on either side: the four changes were requested directly.

## What & why

Four desktop fixes. The open session row was indistinguishable from a hovered one, so it now carries an accent edge, its own background and a heavier title. A launch showed nothing until the daemon finished the runtime handshake; the daemon now reports in-flight launches and their stage over the local status snapshot, and the app adds a placeholder row the moment the server accepts, so the rail row and a stage line in the workspace appear at once. A triple click selected the whole text box; one application-wide tunnel handler selects the line under the pointer instead. The worktree's branch glyph and the PR card's labels share one colour vocabulary: red for closed or failing checks, green ready, muted grey draft, pulsing grey while checks run, purple merged, the warning colour for conflicts.

## Where to look

Conflicts need a `mergeable` field the server does not send yet; the client reads it as null until it does. The tone cache reads every listed session's server-linked PRs every two minutes through the reader's own reveal gate, but keeps a tone past the 30-second window: it is one colour per worktree, not the protected content the window guards, and re-reading each PR every 30 s would multiply GitHub reads. The gh fallback needs the work-context repository the rail rows lack, so local-only PRs stay uncoloured. Markdown bodies keep MarkView's own selection, which takes no click count.

## Verification

```
Capacitor.App.Tests.Unit        total 2014, failed 0 (before the review fixes); 166 across the affected suites after them, failed 0
Capacitor.Cli.Core.Tests.Unit   total 3301, failed 0 (9 skipped); status DTO suites 25 passed after the type move
Capacitor.Cli.Daemon.Tests.Unit (status IPC, pending launch, launch stage suites)  total 48, failed 0
dotnet publish src/Capacitor.Cli -c Release   killed twice locally by memory pressure during native code generation; the CI AOT job is the check
```
The triple-click UI tests fail with the handler uninstalled (`SelectedText` reads the double-click word), so they pin the behaviour rather than Avalonia's default.

https://claude.ai/code/session_01XWRNeNnUJeSbRrQQFwuSui


realtonyyoung:
Qodo findings, addressed in the latest push:

1. **Failed launches stay in the rail** — fixed. `AgentDirectory` now owns a timer on its `TimeProvider` that fires at the earliest placeholder expiry and recomputes, re-armed on every recompute; a test advances a fake clock past ten minutes with nothing else changing.
2. **Local launches can open remote sessions** — fixed. Launch tracking records the lane it launched on; only a published row on that lane settles it, and a same-id remote row neither suppresses the local placeholder nor confirms the launch (test added).
3. **Tones outlive access** — the cache now derives a tone only from a read that passes the reader's own `CanReveal` gate, so an expired or short-lived read yields nothing. It does not schedule removal at the 30-second window: the tone is a single colour per worktree, not the protected content the window guards, and re-reading every 30 s per PR for every listed session would multiply GitHub reads for no user-visible gain. Noted in the PR description as a deliberate call.
4. **Denied PRs keep a tone** — fixed; a denial on any linked PR clears the session's tone (test added).
5. **Signing out leaves stale colours** — fixed; a definitive discovery result (signed out, legacy, unsupported) clears tones, a transient `Unavailable` keeps the last one (tests added for both).
6. **Draft PRs show as open** — fixed; the lifecycle label honours `IsDraft` like the tone does (test added).
7. **Review requests look like running checks** — fixed; `review_required` maps to a distinct `warning` kind that keeps the amber colour and never pulses (test added).
8. **Foreign sessions can colour the rail** — fixed; local rows are excluded while the daemon reports another server, using the directory's own `LocalDaemonOnAppServer` rule (test added).
9. **Local PRs never colour the rail** — not addressed. The gh fallback needs the session's `PullRequestRepository` (provider, host, owner, name, hash), which the workspace gets from the server work-context read; rail rows carry only a local path, so the cache cannot describe sessions and reads server-linked PRs only. Left as a documented limitation.
10. **Rule: one type per file** — fixed; `PendingLaunchDto` moved to its own file.
11. **Cross-repo: conflicts never reach desktop users** — known and documented; orange stays dormant until kcap-server serialises `mergeable`. The client already accepts the field.

realtonyyoung:
NO FINDINGS

PR #974, opened 2026-09-16
Recover the working directory for imported Gemini sessions

George-Payne:
AI-2849 — no GitHub issue (Linear-first for our own work)

## What & why

Gemini names its workspace only in the `<session_context>` bootstrap that opens a recording: the header carries a project hash, and the per-project tmp dir is that same hash. Import therefore had no directory to place a session with, so `excluded_paths` and `excluded_repos` never matched one — and an allow list, which does not admit what it cannot place, skipped every Gemini session. Discovery now reads the bootstrap onto `DiscoveredSession.Cwd`, so the central capture-scope gate scopes these like every other source, and session-start forwards `cwd`/`workspace_root` as the live hook does. Antigravity stays unplaceable: nothing in its transcript records a workspace.

## Where to look

`GeminiSessionContext` rejects far more than it accepts — a turn merely quoting the tag, a drifted heading, an empty list all read as "no workspace". A guessed directory would admit a session the allow list meant to exclude, so unplaceable is the safe answer.

`--cwd` rides on the same recovery, matching whole paths against the recorded workspace; the project-dir basename it can otherwise see is the tmp-dir hash.

## Verification

13 new tests. Each behavioural guarantee falsified by sabotaging the code it covers:

| Sabotage | Tests that fail |
|---|---|
| workspace scan disabled | 2 |
| indentation guard removed | 1 |
| bootstrap-only rule removed | 1 |
| `cwd` dropped from session-start | 1 |

`--treenode-filter '/*/*/*Gemini*/*'`: 80 unit, 9 integration, 36 core — 0 failing.


test/Capacitor.Cli.Tests.Integration/GeminiSubagentImportTests.cs
```diff
@@ -122,6 +122,45 @@ await Assert.That(posts.IndexOf("/hooks/subagent-start"))
...
+    // --- The working directory recovered from the <session_context> bootstrap ---
+
+    [Test]
+    public async Task ImportSession_forwards_the_workspace_recovered_from_the_bootstrap() {
+        const string bootstrap =
+            "<session_context>\\n- **Workspace Directories:**\\n  - /work/demo\\n- **Directory Structure:**\\n</session_context>";
+
+        var chats = Path.Combine(_tempDir, "proj", "chats");
+        Directory.CreateDirectory(chats);
+        File.WriteAllLines(Path.Combine(chats, "session-2026-06-22T14-31-0a900002.jsonl"), new[] {
```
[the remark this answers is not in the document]
George-Payne: Disputed as scoped to this change.

`new TempDir()` is the dominant pattern in this repo, not a deviation from one: 1191 uses across 215 test files, against 69 for the injected property. `GeminiSubagentImportTests` owns `_tmp` and `_tempDir` already, and every test in it builds paths that way — the new test follows the class it lives in.

Converting the class to an injected fixture would touch every test in it for no behavioural gain, in a PR about recovering a working directory. Happy to see the convention settled as its own issue, and applied everywhere at once rather than in whichever file a PR happens to open.

src/Capacitor.Cli/Harness/Gemini/GeminiImportSource.cs
```diff
@@ -265,12 +277,18 @@ CancellationToken                   ct
...
         var payload = new JsonObject {
             ["hook_event_name"] = "SessionStart",
             ["session_id"]      = sessionId,
             ["source"]          = "startup",
         };
+        // Same two fields the live hook forwards, so an imported session attributes to a repo
+        // exactly as a captured one does. Git-root discovery is fail-open.
+        if (cwd is not null) {
+            payload["cwd"] = cwd;
+            if (GitRepository.FindRoot(cwd) is { } workspaceRoot) payload["workspace_root"] = workspaceRoot;
```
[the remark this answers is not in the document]
George-Payne: Correct about the server, and deliberately kept — filed as AI-2856.

The live Gemini hook (`GeminiHookCommand`) already sends `workspace_root` on every session start, derived the same way. So the server dropping it is not something this change introduces, and removing it here would make import the one path that stops sending it — the opposite of the parity this line exists for. Whenever the server accepts the field, both paths gain it together.

The gap is also wider than Gemini. `GeminiSessionExtension` has no `WorkspaceRoot`, and neither do the Antigravity, Copilot, Kiro, Pi or OpenCode extensions; `SessionStartedAccessors.WorkspaceRoot()` reads only `ClaudeCode()` and `Codex()`, while its own doc comment says a Gemini session "must not silently lose it". That is the shape of the fix, and it belongs server-side rather than in this PR.

src/Capacitor.Cli/Harness/Gemini/GeminiImportSource.cs
```diff
@@ -413,32 +431,55 @@ static ImportCommand.SessionClassification MakeClassification(
...
     /// </summary>
-    static (string? SessionId, DateTimeOffset? StartTime) ReadHeader(string path) {
+    static (string? SessionId, DateTimeOffset? StartTime, string? Workspace) ReadHead(string path) {
+        string?         sessionId = null;
+        DateTimeOffset? startTime = null;
+        string?         workspace = null;
+        var             scanned   = 0;
+
         try {
             foreach (var line in File.ReadLines(path)) {
```
realtonyyoung: ReadHead now keeps File.ReadLines open while it scans and parses up to eight records of a Gemini-owned transcript. File.ReadLines opens via StreamReader with FileShare.Read, which denies a concurrent writer on Windows; Gemini appending during import can hit a sharing violation (or this read can fail), even though the later ReadTranscriptStatsAsync path correctly uses FileShare.ReadWrite. Please open this head scan with a FileStream using FileShare.ReadWrite and a StreamReader too. The old one-line header read had the same underlying risk, but this change lengthens that write-denying window while doing workspace extraction.

realtonyyoung (changes requested):
One actionable Windows file-sharing issue in the new transcript head scan; details are in the inline comment. Static review only; I did not build or run tests.

src/Capacitor.Cli/Harness/Gemini/GeminiImportSource.cs
George-Payne: Confirmed and fixed in 820443ba.

Switched to `File.ReadLinesShared`, Core's streamed shared read (`FileShare.ReadWrite | FileShare.Delete`), which is what the other transcript scans already use — `SecondaryRepoRoots`, `WatchCommand`, and `ImportCommand`'s evidence read. Its own doc comment names this exact failure: `File.ReadLines` opens `FileShare.Read`, which "on Windows is mandatory sharing and locks a coding agent out of the file it owns".

Your point about the lengthened window is the part that made it worth fixing rather than noting. The previous one-line header read opened and closed inside a single `MoveNext`; the head scan holds the enumerator across up to eight records while JSON-parsing each one, so the window now spans the workspace extraction rather than a single read.

Gemini suites after the change: Core 37, Unit 80, Integration 9, all passing.

realtonyyoung: Verified on 820443ba: ReadHead now uses File.ReadLinesShared, whose stream opens with FileShare.ReadWrite | FileShare.Delete. This resolves the Windows sharing finding. I am approving this head.

realtonyyoung:
NO FINDINGS on 820443ba. Re-reviewed the Gemini workspace extraction and verified the shared-read fix resolves my prior Windows file-sharing finding. Static review only; I did not build or run tests.

realtonyyoung (approved):
Approved after static re-review; the prior finding is resolved.

PR #977, opened 2026-09-16
Scope unattended agent launches to the profile's capture lists

George-Payne:
AI-2853 — no GitHub issue exists for this; we file in Linear.

## What & why

A review flow's reviewer and a PR review run wherever the flow points them, so nobody opted that directory in — and the daemon reported them regardless. The prompt, the terminal output and the ACP transcript all reach the server with no capture-scope gate in front of any of them.

The line drawn here is whether a human chose the directory, not who dispatched the launch: an agent asked for by hand is left alone, because choosing the directory is the opt-in. An unattended one is judged against the same allow and deny lists a hook applies, and refused before any worktree or borrowed snapshot exists.

## Where to look

The verdict is taken on the originating checkout rather than the execution directory — a snapshot reviewer runs in a daemon-owned copy that matches nothing a user configured. `ResolveMainRepoRoot` collapses borrowed, snapshot and owned-worktree launches onto one repository, so a repo verdict is the same whichever containment a vendor happened to get.

The refusal sent to the server names neither path nor repository. That identity is exactly what the lists exist to withhold, so a reason carrying it would upload the one thing the refusal protects.

## Verification

29 new tests. Every behavioural claim falsified by sabotaging the production code:

| Sabotage | Failures |
| --- | --- |
| Gate applied to every launch, not only unattended ones | 1 (the human-asked-for case) |
| Gate removed entirely | 3 |
| Refusal reason carries the checkout path | 1 |
| Allowlist admits a repo whose key would not resolve | 1 in Core, 1 in the daemon |

`Capacitor.Cli.Core.Tests.Unit` 3370 passed / 0 failed; `AgentOrchestrator*` 211/211; `PathExclusionTests` 35/35 after the move to Core.


src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
```diff
@@ -5356,6 +5375,47 @@ static string ExtractTerminalText(TerminalOutputBuffer buffer) {
...
+    /// has to bind to the next launch rather than the next restart —
+    /// <see cref="ProfileContext.Snapshot"/> says the same of any setting a long-lived process must
+    /// observe. Launches are rare enough for a disk read; the hook path this shares rules with is
+    /// the one that cannot afford one.
+    /// <para>A read that fails falls back to the boot resolution, which still restricts. Treating
+    /// it as "no lists" would turn a transient disk error into an open gate.</para>
+    /// </summary>
+    async Task<Profile?> CurrentCaptureProfileAsync() {
+        try {
+            var snapshot = await AppConfig.LoadProfileConfig(_config.ConfigRoot);
```
realtonyyoung: A failed config reload can silently open this privacy gate. LoadProfileConfig returns ProfileConfig.Fresh() (rather than throwing) when config.json is missing, unreadable, or invalid; Fresh has no capture lists. If the daemon started with an excluded path and a transient read error occurs, this method returns the empty profile, Configured is false, and an unattended review of that excluded checkout proceeds. Please distinguish a successful empty profile from a failed read and fail closed on the latter (or retain a known restrictive snapshot).

src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
```diff
@@ -2165,6 +2174,16 @@ await _server.LaunchFailedAsync(cmd.AgentId,
                     throw new InvalidOperationException($"borrow_auth_failed: {auth.Reason}");
                 }
 
+                // The canonical path is the one that will actually be run, and it is only known
+                // here: a symlink retargeted between the request and this resolution would
+                // otherwise hand the runtime a checkout no list was ever compared against. Before
+                // the snapshot, so a refusal never copies a checkout it is about to discard.
+                if ((isReview || isReviewFlow)
+                 && await RefuseOutOfCaptureScopeAsync(agentId, auth.CanonicalGitRoot ?? auth.CanonicalCwd)
```
realtonyyoung: For a direct borrowed review, the runtime uses CanonicalCwd below, but this passes CanonicalGitRoot to PathExclusion. With a repo at /work/repo, BorrowCwd=/work/repo/private, and excluded_paths: [/work/repo/private], the root is not excluded and the unattended reviewer runs in the excluded directory and uploads its output. An allowlist targeting that subdirectory is likewise rejected. Please evaluate the path lists against CanonicalCwd, while GitRepoKey can still resolve the repository from that directory.

realtonyyoung (changes requested):
Two capture-scope issues need correction before approval; details are in the inline comments. Static review only; I did not build or run tests.

src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
George-Payne: Confirmed and fixed in 7eb92bdc.

You're right that nothing throws, which is what made the fallback useless: `LoadProfileConfig` returns `ProfileConfig.Fresh()` for a missing, unreadable *or* invalid config, and `Fresh()` seeds a `default` profile — so `Profiles.GetValueOrDefault(name)` found a non-null empty profile and the `?? _config.Profiles.Effective` I'd written as the safety net never ran at all.

The read is now proven before its result is trusted, and the two cases are separated: an absent config file is a real answer (nothing configured, nothing scoped), while a file that exists and cannot be read or parsed refuses the launch with its own reason rather than the exclusion one — the cause is a broken config, not a checkout the lists exclude, and only one of those is actionable by whoever sees the flow fail.

Falsified by returning "readable" from that catch: `An_unreadable_config_refuses_rather_than_admitting_the_launch` fails, and passes again once it fails closed.

src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
George-Payne: Confirmed and fixed in 7eb92bdc — the path lists now see `CanonicalCwd`.

Your example is the test: a repo admitted at its root with `excluded_paths: [<root>/private]` and `BorrowCwd` pointing at that subdirectory. Judging the root found nothing excluded and the reviewer ran there anyway.

And as you say, the repo key is unaffected by the change — `GitRepoKey.ForCheckout` walks up from wherever it starts, so it still resolves the repository from the subdirectory.

`A_borrow_in_an_excluded_subdirectory_is_refused_though_its_repo_root_is_not` covers it, and asserts the snapshot is never cut as well as the refusal. Reverting the argument to `CanonicalGitRoot ?? CanonicalCwd` fails it.

src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
```diff
@@ -5356,6 +5378,73 @@ static string ExtractTerminalText(TerminalOutputBuffer buffer) {
...
+            // Proven readable before the result is trusted. LoadProfileConfig answers an
+            // unreadable or invalid config with ProfileConfig.Fresh() rather than an error, and
+            // Fresh() carries a default profile with empty lists — which for a privacy gate is
+            // indistinguishable from a profile that genuinely scopes nothing.
+            using var _ = JsonDocument.Parse(await File.ReadAllTextAsync(path));
+        } catch {
+            return (null, false);
+        }
+
+        var snapshot = await AppConfig.LoadProfileConfig(_config.ConfigRoot);
```
realtonyyoung: The config pre-check only validates JSON syntax. A type-invalid document can pass it, then LoadProfileConfig returns an empty Fresh profile after deserialization fails; the unattended launch is admitted. Please use one parse with an explicit success result and reject on failure.

realtonyyoung (changes requested):
The borrowed-directory finding is fixed, and the unreadable-file path now rejects. One capture-scope fail-open case remains: a syntactically valid config rejected during deserialization becomes an empty Fresh profile. Details are inline. Static review only; no build or tests run.

src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
George-Payne: Right, and it went one layer deeper than the type case. Fixed in 0b66d7c7.

Three shapes land on the same empty profile and only one of them throws. `ConfigMigration.MigrateIfNeeded` answers **not JSON** with `FreshDefault()` in its own `catch (JsonException)`, and **JSON that is not an object** with `FreshDefault()` on the line after — neither reaches `LoadProfileConfig`'s catch at all. Only your case, a typed field that will not deserialize, escapes as an exception. So the syntax pre-parse I had was covering the one shape that would have been caught anyway and missing both that were not, which is the general problem with validating ahead of the parse: it only ever covers what it anticipates.

So the outcome is carried out of the single parse, at both layers — `ConfigMigration.TryMigrate` and `AppConfig.TryLoadProfileConfig` return why they ended where they did, and the daemon refuses on `Unreadable`. `MigrateIfNeeded` and `LoadProfileConfig` keep their signatures and their lenient answers by delegating, so no existing caller changes behaviour.

Worth noting your finding was not theoretical: `ConfigMutator` already carries a hand-written guard validating the document before handing it to migration, for this same reason.

Tests: a config that is not JSON, and one that parses but does not deserialize, each refuse; a missing config still does not. Falsified by reporting `Ok` from the migration's two swallow points — the malformed case fails, and passes again once it reports `Unreadable`. Daemon `Agent*` 291/291, Core 3372/0.

src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
```diff
@@ -5356,6 +5377,63 @@ static string ExtractTerminalText(TerminalOutputBuffer buffer) {
...
+    }
+
+    /// <summary>
+    /// The capture lists as they stand now, read per launch rather than taken from the resolution
+    /// the daemon booted with. A daemon outlives its config, and a list added while one is running
+    /// has to bind to the next launch rather than the next restart —
+    /// <see cref="ProfileContext.Snapshot"/> says the same of any setting a long-lived process must
+    /// observe. Launches are rare enough for a disk read; the hook path this shares rules with is
+    /// the one that cannot afford one.
+    /// <para>A read that fails falls back to the boot resolution, which still restricts. Treating
```
realtonyyoung: This paragraph is now inaccurate: a failed load returns Readable=false, and RefuseOutOfCaptureScopeAsync rejects the launch. Please describe that fail-closed behavior instead of a fallback to the boot profile. This is security-boundary documentation, so the stated behavior should match the code.

test/Capacitor.Cli.Daemon.Tests.Unit/Services/AgentCaptureScopeLaunchTests.cs
```diff
@@ -0,0 +1,311 @@
...
+        await orch.HandleLaunchAgentForTest(new LaunchAgentCommand(
+            "borrow-out", "review", "default", null, cwd, null, null,
+            Vendor: "cursor", Kind: LaunchKind.ReviewFlow, Borrowed: true, BorrowCwd: cwd));
+
+        await Assert.That(server.LaunchFailedCalls.Single(c => c.AgentId == "borrow-out").Reason)
+            .Contains("out_of_capture_scope");
+        await Assert.That(factory.LastContext).IsNull();
+    }
+
+    // Tony's case on #977: a borrow may sit BELOW its repository root. Judging the root would
```
realtonyyoung: Please remove the reviewer/PR reference and describe only the invariant this test pins. The repository comment rule explicitly excludes review artifacts and change-history narration from source and tests; the current behavior is that capture scope evaluates the borrowed cwd, including an excluded subdirectory.

realtonyyoung (changes requested):
The single-load outcome fixes my functional capture-scope finding. Two new comments still need cleanup to match the repository comment policy and current fail-closed behavior; see inline notes. Static review only; no build or tests run.

src/Capacitor.Cli.Daemon/Services/AgentOrchestrator.cs
George-Payne: Fixed in fbbb2512. It now states the fail-closed behaviour: an unreadable config returns `Readable: false` and the caller refuses the launch, because unknown lists are not absent lists.

The stale paragraph was left over from the version where the fallback was to the boot resolution — you're right that it was describing the boundary as more permissive than the code, which is the one direction that misleads.

test/Capacitor.Cli.Daemon.Tests.Unit/Services/AgentCaptureScopeLaunchTests.cs
George-Payne: Fixed in fbbb2512 — the comment now states only the invariant: capture scope judges the directory the runtime is handed rather than the repository root above it, so a borrow into an excluded subdirectory is refused even though its root is admitted.

I also swept the rest of the diff for the same two tells and found nothing else: no reviewer or PR references, and no past-tense narration of the code's own history in any line this PR adds.

src/Capacitor.Cli.Core/Config/ConfigMigration.cs
```diff
@@ -3,23 +3,37 @@
 
 namespace Capacitor.Cli.Core.Config;
 
+/// <summary>Whether a config could be understood at all. <see cref="Unreadable"/> means its
+/// contents are unknown, not empty.</summary>
+public enum ConfigMigrationOutcome { Ok, Unreadable }
```
realtonyyoung: The CLI repository requires one type per file, named after the type; an enum alongside an unrelated class is not one of its exceptions. Please move ConfigMigrationOutcome to ConfigMigrationOutcome.cs and the new ProfileConfigLoad enum from ProfileConfig.cs to ProfileConfigLoad.cs. Existing multi-type files are explicitly not a precedent in the repository instructions.

realtonyyoung (changes requested):
The prior capture-scope and comment findings are resolved. One repository convention remains: the two new public enums should each have their own named file. Details inline. Static review only; no build or tests run.

src/Capacitor.Cli.Core/Config/ConfigMigration.cs
George-Payne: Fixed in b0b604fe. `ConfigMigrationOutcome` and `ProfileConfigLoad` each have their own file now, named after the type.

Agreed on the exception test: neither is an enum-plus-its-extension-methods, a hierarchy, or a descriptor registry — they were simply sitting beside an unrelated class, and the rule says in as many words that the neighbouring files are not a precedent.

I checked the rest of the diff against the same rule rather than just these two: every type this PR adds is now in a file named after it.

realtonyyoung:
NO FINDINGS — re-reviewed the current head. The borrowed-cwd scope check, single-load fail-closed config outcome, corrected comments, and separate enum files address my prior findings. Static review only; no build or tests run.

realtonyyoung (approved):
Prior findings are resolved in the current diff; no remaining actionable findings. Static review only; no build or tests run.

Issue #970, opened 2026-09-16
Flaky daemon unit tests: TranscriptJournal pending gap and version-probe pipe drain


realtonyyoung:
Two timing-sensitive tests in `Capacitor.Cli.Daemon.Tests.Unit` failed once each on PR #969 (run https://github.com/kurrent-io/kcap-cli/actions/runs/35034079520, attempt 1) and passed on the rerun (attempt 2). The PR touches nothing under `src/Capacitor.Cli.Daemon` or that test project.

**ubuntu-latest** — `TranscriptJournalTests.Record_never_blocks_while_the_sink_hangs` (`test/Capacitor.Cli.Daemon.Tests.Unit/Services/TranscriptJournalTests.cs:109`)

```
AssertionException: Expected to be greater than 0 but received 0
at Assert.That(journal.PendingGap).IsGreaterThan(0)
```

**macos-latest** — `DaemonRunnerVersionProbeTests.A_version_whose_output_exceeds_the_pipe_buffer_is_drained_not_deadlocked` (`test/Capacitor.Cli.Daemon.Tests.Unit/DaemonRunnerVersionProbeTests.cs:70`)

```
AssertionException: Expected to be equal to "9.9.9" but received "0123456789ABCDEFGHIJ"
at Assert.That(DaemonRunner.ProbeCliVersionForLaunch(cli)).IsEqualTo("9.9.9")
```

Same assembly was red on main earlier the same day for another test (`Replayed_session_end_with_generate_whats_done_true_drains_without_error`, run 35015885056, macos), so this looks like a shared timing sensitivity in the daemon test host rather than three independent defects. Both assertions read a value that a background pump/process produces; the first reads the gap before the sink has stalled, the second sees the probe's stdout truncated to the first pipe-buffer chunk.

Filed for triage; the rerun was not the fix.

PR #978, opened 2026-09-16
[AI-2857] Spawn detached children without inheriting the agent's handles

Lougarou:
AI-2857 — no GitHub issue; this repo's issues are tracked in Linear only.

## What & why

On Windows a child spawned from a coding-agent hook inherits the agent's stdout pipe and holds it for its whole lifetime, so the agent's read of the hook's stdout never reaches EOF and it abandons the hook at its timeout. `CreateProcess` inherits every handle marked inheritable, and an agent invokes its hook holding more of them than `GetStdHandle` reports, so clearing those three cannot close it. `IProcessStarter` gains `StartDetached`, which on Windows goes through `CreateProcess` with `bInheritHandles: false` — the only cut that holds, and one `System.Diagnostics.Process` offers no way to ask for. Unix keeps the fd >= 3 CLOEXEC sweep, which closes its own differently-shaped leak.

## Where to look

`ClaudeSessionEndHandoff` deliberately keeps the old spawn: it hands its payload to the child through stdin, and on Windows a pipe only reaches a child by being inherited. Closing that one needs the spawn to pass an explicit `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` naming just that pipe.

## Verification

Live Codex (0.154.0) against a pristine `origin/main` control worktree — same machine, same harness, each hook wrapped in a shim timestamping entry and exit, so hook runtime is separated from what the agent does after it exits.

| build | hook itself | gap to next hook | outcome |
| --- | --- | --- | --- |
| `origin/main` | 2.85 s, rc=0 | ~10 min | run killed at 600 s |
| this branch | 2.53 s, rc=0 | 0.5 s | SessionStart, UserPromptSubmit, Stop all `Completed`; 15.4 s total |

A watcher spawned by the patched binary was confirmed alive afterwards, so capture still works rather than the spawn silently failing.

Unit suite: 4273 passed, 3 failed — the three `IsExcluded_*` symlink tests, `A required privilege is not held by the client` (Windows symlink privilege), which fail the same way on `origin/main`.


Lougarou:
@realtonyyoung ready for review — out of draft.

An independent Codex review flow ran this to sign-off over five rounds. It returned `clean` on the final round; eight findings were raised and fixed along the way, five commits on top of the original `ac9bf63b`.

**What it found, in order**

| # | Finding | Fix |
| --- | --- | --- |
| 1 (P1) | `ClaudeSessionEndHandoff` still inherited the agent's handles — it stayed on the old spawn because its payload goes over the child's stdin, and a pipe reaches a child only by inheritance. The original hook hang survived on that path. | `PROC_THREAD_ATTRIBUTE_HANDLE_LIST` naming just that one pipe handle |
| 2 | Unix `Process` wrapper dropped undisposed in `StartDetached` | `using (process)` |
| 3 | `CreateProcessW` failures discarded the Win32 reason | `Win32Exception`, read immediately |
| 4 | The new Unix `StartDetachedWithStdin` repeated finding 2's leak | `ChildStdinStream` owns the `Process` |
| 5 | `DeleteProcThreadAttributeList` could run on a never-initialized buffer | guarded by an `initialized` flag; allocation always freed |
| 6 | The raw pipe handle had two owners if the `FileStream` ctor threw — the second close could land on a reused handle value | ownership moves to a named `SafeFileHandle` first; ctor failure disposes it and terminates the child |
| 7 | `owner.Dispose()` was not in a `finally`, so a throwing pipe close abandoned the process handle | disposal gate + `finally` in both forms |
| 8 | The disposal tests did not constrain the change — built on a real `FileStream`, they stayed green with the gate removed | rebuilt on a `CountingStream` double that records how it was closed and can fail its close |

Findings 4–8 are all defects in fixes for earlier findings, which is the part worth knowing: the first round's fix for the P1 introduced its own leak, and two rounds of tests didn't pin what they claimed to.

**Verification**

- Full `Capacitor.Cli.Tests.Unit`: 4197 passed, 83 skipped, 3 failed — the three `IsExcluded_*` Windows symlink-privilege tests, which fail identically on `origin/main`.
- `dotnet publish -c Release` (NativeAOT): clean, no IL2026/IL3050.
- The disposal tests are mutation-checked: removing the gate and each `finally` in turn fails them, so they constrain the contract rather than decorating it.

**Two things to know before you read it**

1. **The branch does not build without `-p:NoWarn=IDE0200`.** `AgentHookPoster.cs:285` and `McpFlowsServer.cs:87` raise `error IDE0200: Lambda expression can be removed`. These are unchanged from the merge-base parent (confirmed by blob identity), so they pre-date this branch and I left them alone — but CI will be red until they're fixed somewhere.
2. **`ClaudeSessionEndHandoff` now spawns through a new `IProcessStarter.StartDetachedWithStdin`.** That is the riskiest surface here: hand-rolled `CreateProcessW` interop with an attribute list, and handle ownership that has to hold on every throw path. Rounds 2 and 3 both found real defects in it. Worth your eyes even though the reviewer signed off.

realtonyyoung:
I reviewed the latest diff by inspection. Two production concerns remain: on Unix, `skills sync --auto` can still emit a profile-resolution warning before its console writers are nulled, after the detached parent closes the redirected pipe; and the detached-stdin failure paths reacquire a child by PID after releasing its original process handle, creating a PID-reuse kill risk. The existing inline review threads describe both, so I am not duplicating them. The fake-starter fallback test also appears to start a child and then throw before returning its PID when stdin is not redirected, leaving cleanup without ownership. I did not build or run tests.

realtonyyoung (changes requested):
Requesting changes for the detached auto-sync stderr race and detached-child PID ownership on failure; the fake-starter fallback test also needs cleanup. Details are in my PR comment and existing inline threads. Builds and tests were not run.

Lougarou:
@realtonyyoung all three addressed, pushed as `fb9afec8` and `1ebd14f8`. Both production concerns were real.

**1. Auto-sync dying on warnings — fixed, and you're right that this PR caused it.**

The regression is visible in the diff: the old code kept the child's stdout/stderr open and drained them with `CopyToAsync(Stream.Null)`, and its comment said the child only silences its own streams later. This PR replaced that with an immediate close, leaving that startup window uncovered — profile resolution runs at `Program.cs:~108`, the writers weren't nulled until dispatch at `:488`.

The child now nulls its writers at startup, before config and profile resolution can emit anything. The dispatch-site silencing is gone rather than duplicated.

**2. PID-reuse kill — fixed by removing the pid re-acquisition entirely.**

New `DetachedChild` carries the identity the child was created with: the `Process` wrapper on Unix, the `CreateProcess` handle (`SafeProcessHandle`) on Windows. `Terminate()` goes through that handle, so neither failure path resolves a pid any more. Windows keeps `hProcess` open instead of closing it and calling `Process.GetProcessById` later.

This also let `ChildStdinStream` be deleted — `DetachedChild` subsumes it, and the async-disposal path it needed (along with an earlier finding about that path's `finally`) no longer exists.

**3. Fake starter leaving a child alive — already fixed before your review, and it was a real orphan.**

Linux caught it: `A_child_that_never_receives_the_payload_is_killed_before_the_inline_fallback` **failed on Linux** with `pid 1478 still carries identity ... after 10s`. It's skipped on Windows, so no local run could have shown it. The underlying defect was mine, not the test's: the starter threw after spawning, so `TrySpawn` never learned the pid and nothing could clean up. The starter now terminates a child it cannot hand back, and the fake models that same contract.

## On Linux

You noted you didn't build or run. I did, in a .NET 10 container, because this PR's Unix paths have two tests that Windows always skips.

- **Full unit suite on Linux: 4256 passed, 3 failed.** The three are `Sweep_failure_propagates_to_exit_code`, `Config_dir_is_preserved_when_user_level_steps_fail`, `Cursor_hooks_write_failure_propagates_to_exit_code`. They fail **identically on the merge-base parent `ac9bf63b`**, and they pass when the same suite runs as a non-root user. They strip write permission with `File.SetUnixFileMode` and expect `EACCES`; root bypasses DAC. Environment, not this branch.
- **Real-binary end-to-end** (`Capacitor.Cli.Tests.Integration/ClaudeSessionEndHandoffTests`, which stalls the server so only a process outliving the hook can post) passes on Linux.

One thing worth flagging since it is **not** caused by this PR but is close to the edge: that integration test asserts the hook returns in under 1s, against Claude's real 1.5s kill. In a loaded container the baseline `ac9bf63b` failed 4 of 5 runs at 1.29s, 1.42s, 1.29s and 2.76s; this branch failed 1 of 6, worst 1.47s. So the branch is no slower — but the headroom is thin enough that a slow machine can blow the real 1.5s budget. Might deserve its own issue.

## Still open, unchanged

The branch does not build without `-p:NoWarn=IDE0200` (`AgentHookPoster.cs:285`, `McpFlowsServer.cs:87`). Both files are byte-identical to the merge-base parent, so they pre-date this work and I've left them alone — but CI stays red until they're fixed somewhere. Say the word if you'd rather they ride along here.

realtonyyoung:
NO FINDINGS on the updated head. Auto-sync now silences output before profile resolution can warn; the detached child retains its original process handle for failure cleanup; and the fake starter cleans up a child it cannot hand back. I reviewed the changed production and test code by inspection and did not build or run tests.

realtonyyoung (approved):
The detached-child and early-output findings from my previous review are addressed. No further actionable findings by inspection; no build or tests run.

Lougarou:
CI is green — and a correction to something I said earlier.

## Correction: the `IDE0200` build break is not real

In my two earlier comments I said this branch "does not build without `-p:NoWarn=IDE0200`" and that "CI stays red until they're fixed somewhere". **That was wrong, and I'm sorry for the noise** — please don't go looking for a build break.

The build passes on Windows, Ubuntu and macOS, and both AOT publish checks pass. The cause is local to my machine: `global.json` pins SDK `10.0.100` (`rollForward: latestFeature`) and CI installs that, while I was building on `10.0.101`, where that rule's severity differs. Nothing to do in `AgentHookPoster.cs` or `McpFlowsServer.cs`.

## The failing check was a flake, not this PR

The first run failed one test: `TranscriptJournalTests.Gap_note_lands_between_the_last_kept_and_the_first_after_the_loss`, on `await journal.CompleteAsync()` returning false. Re-running the job turned the whole matrix green with no code change.

Why it isn't ours:

- This PR touches **zero** files in `Capacitor.Cli.Daemon`, src or test — only `src/Capacitor.Cli` and `test/Capacitor.Cli.Tests.Unit`.
- That test passes 15/15 locally, three runs in a row.
- The class is already a known flake target: your `30b414d6` ("De-flake the journal pending-gap and version-probe drain tests (#970) (#994)") edits this exact file, and this branch is based on `0d540954`, which predates it — so CI here still runs the pre-de-flake version of that file.
- `main` is tripping timing-sensitive tests too at the moment, a different one most runs: `HardCap_after_resolve_sessionStart_emits_empty_once`, `AConsentLessDaemonSeedsTheFloorThatAdmitsAHostedLaunch`, `A_descendant_holding_the_pipe_open_is_bounded_and_still_reads_the_version`, `A_version_at_the_front_survives_a_trailing_flood_on_the_same_stream`.

Happy to merge `main` in so this branch picks up the three de-flake commits, if you'd rather it ran against them — say the word.

PR #979, opened 2026-09-17
Expire the connecting banner when the daemon never catches up

nortonandreev:
AI-2877 — no GitHub issue

## What & why

A live app lane plus a daemon that never reaches `connected` kept the launcher on "Connecting to the server…" with Sign in hidden: catch-up after auth looks identical to a daemon that 401-retries forever. After 60s (longer than one 30s connect backoff) that line becomes the lost-session notice and Sign in is offered. A daemon that connects inside the bound still clears the banner.

## Where to look

`HomeViewModel` catch-up timer and `NoticeAfterCatchUp`. The pre-bound catch-up tests must still hide Sign in.

## Verification

```
dotnet run --project test/Capacitor.App.Tests.Unit/Capacitor.App.Tests.Unit.csproj -- --treenode-filter "/*/*/HomeViewModelTests/*"
```

82 passed, including four new catch-up-bound cases.


nortonandreev:
Addressed both Qodo findings:

1. **Redundant test comment** — Removed the XML summary above `ALiveAppLaneStopsTreatingDaemonDisconnectAsConnectingAfterTheCatchUpBound`; the name and assertions already pin the behavior.

2. **Stale catch-up callback** — Agreed. `OnCatchUpElapsed` schedules onto the main thread, so a stop/restart (e.g. `NotifySignInCompleted`) could run before a queued expiry and the old handler only checked `_catchUpArmed`. Catch-up now bumps a generation on stop/restart; the scheduled action applies timeout only when that generation still matches and `GetUtcNow()` is past the deadline recorded for the current arm.

`Capacitor.App.Tests.Unit` (2540 tests) passes locally.

PR #989, opened 2026-09-17
Take ambient process state from the composer, not the process

### LINKED ISSUE #990: Production code reads ambient process state instead of taking it from its composer
Several types reach process-wide state directly instead of taking it from whoever composed
them. Production is correct today — the composition roots resolve the real values — so each
one shows up as a testing or isolation problem first, and only becomes a defect when a
caller wants to supply something else.

Found in a sweep following #963, which did the same for the clock.

**Sites**

- `AgentDirectory` takes `TimeProvider? time = null` and falls back to `TimeProvider.System`.
  Eleven of its twelve construction sites take the wall clock without saying so, while the
  type drives a ten-minute placeholder expiry off that provider.
- `ServiceTxnMarker.FlushDirectory` is a reassignable `static Func<string, bool>` whose only
  reason to be writable is letting a test observe the fsync. Swapping it needs bare
  `[NotInParallel]`, which makes that test exclusive against the whole assembly.
- `SettlementBackoff` hides `Random.Shared` behind a constructor default.
- `BinaryProbe.OnPath` reads `PATH` from inside the plugin-install precheck, so two test
  suites repoint process-wide `PATH` to reach it — each forcing a bare `[NotInParallel]`.

**Known to remain after that**

- The drain and offer throttles (`AgentHookPoster.TryClaimDrainAttempt`,
  `HarnessOfferStore.TryClaimCheck`) measure their window by comparing an injected clock
  against the stamp file's mtime. The write side stamps mtime implicitly, so nothing can
  advance it: three suites burn the real window in a `[Before(Test)]` hook because no fake
  clock can expire it. The fix that keeps them fakeable puts the instant in the stamp's
  content, which changes an on-disk format.
- `McpFlowsServer.HandleToolCallAsync` defaults `clock ??= FlowRetryClock.System` and
  `backoff ??= SettlementBackoff.Default`. A retry-path test that omits `clock` gets the
  real wall clock and a real `Task.Delay`. Production always passes one; nothing enforces it.

Not in scope, but adjacent: `InvariantGlobalization` is set on `Capacitor.Cli` and
`Capacitor.Cli.Daemon` only — not on `Capacitor.App`, and on no test project — so the app
ships with ICU and every suite exercises a culture its own production code does not have.
`AttachmentTrayTests.Size_label_and_image_flag` fails on any comma-decimal locale because of it.

Inok:
Closes #990 — AI-2914

## What & why

Four places where production code reached ambient process state instead of taking it
from its composer: a `TimeProvider` that fell back to `TimeProvider.System` when
omitted, a reassignable static that existed only so a test could observe an fsync, an
RNG hidden behind a constructor default, and a `PATH` read buried inside the
plugin-install precheck. Where a dependency became required rather than defaulted, the
compiler enumerated the call sites instead of letting any inherit a value silently.

A last commit stabilizes three timing-sensitive tests unrelated to the above. Each
measured the runner rather than the property under test, and one of them fails on `main`
today.

## Where to look

The `PATH` commit has the reach. Routing the `kcap` precheck through the injected probe
turned 15 tests red at once — every one of them green until now only because this
machine happens to have `kcap` installed. Two suites carried a fixture whose whole job
was repointing process-wide `PATH`; both now stage a probe, which retires two
`EnvScope.Exclusive("PATH", …)` mutations and the bare `[NotInParallel]` each forced.

The session-end hook's elapsed bound no longer asserts Claude Code's 1.5 s grace: that
clock includes an AOT process start, which on a contended two-core runner spends most of
the budget before the hook's first instruction. It now derives from the server stall the
hook is proving it did not wait for — 78 ms observed locally, 1.68 s on the runner that
failed, 3 s threshold, 6 s if the POST were ever awaited.

Left alone: the drain and offer throttles compare an injected clock against their stamp
file's mtime, which nothing can advance. The only fix that keeps them fakeable changes
the stamp's on-disk format.

## Verification

| Suite | Result |
|---|---|
| `Capacitor.Cli.Tests.Unit` | 4308/4308 |
| `Capacitor.Cli.Core.Tests.Unit` | 3401/3401 |
| `Capacitor.Cli.Tests.Integration` | 291/291 |
| `Capacitor.App.Tests.Unit` | 2261/2262 |

The one failure is `AttachmentTrayTests.Size_label_and_image_flag`, which fails the same
way on `main`: `StagedAttachment.SizeLabel` uses a culture-sensitive format. It passes on
CI, whose locale is invariant.

Assertions that changed meaning were checked by mutation: stubbing the fsync barrier to
never report firing, removing `Delete`'s `try/catch`, and tightening the session-end
bound to 1 ms each reddened the intended test and nothing else.


test/Capacitor.Cli.Tests.Unit/Services/ServiceTxnMarkerTests.cs
```diff
@@ -50,23 +51,26 @@ public async Task Delete_of_missing_marker_does_not_throw() {
...
-            await Assert.That(flushed.Count).IsGreaterThanOrEqualTo(2);
-        } finally {
-            ServiceTxnMarker.FlushDirectory = original;
-        }
+    /// <summary>The barrier reports whether it fired, and it fires on a real directory — which is
+    /// what makes the false below a genuine failure signal rather than the only answer it has.</summary>
+    [Test]
+    public async Task Directory_barrier_fires_on_a_real_directory() {
+        if (OperatingSystem.IsWindows()) return; // no portable directory fsync there
+        await Assert.That(ServiceTxnMarker.FlushDirectory(Daemons.Directory)).IsTrue();
```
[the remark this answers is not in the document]
Inok: And why the tests pass then? 🤔

test/Capacitor.Cli.Tests.Unit/Services/ServiceTxnMarkerTests.cs
```diff
@@ -50,23 +51,26 @@ public async Task Delete_of_missing_marker_does_not_throw() {
...
+
+    [Test]
+    public async Task Directory_barrier_reports_failure_rather_than_throwing() {
+        await Assert.That(ServiceTxnMarker.FlushDirectory(Path.Combine(Daemons.Directory, "absent"))).IsFalse();
+    }
+
+    /// <summary>A barrier that cannot fire must not fail the delete it follows: the marker going away
+    /// is the caller's contract, the fsync behind it is hardening.</summary>
+    [Test]
+    public async Task Delete_survives_a_barrier_that_cannot_fire() {
```
realtonyyoung, line 70: The replacement tests check `FlushDirectory` directly, but none observes whether `Write` and `Delete` actually call it. Removing either production call would leave these tests green, whereas the deleted test asserted both calls. Could we preserve coverage of that post-rename/post-delete durability ordering without a process-global mutable delegate (for example, via a per-call/internal injected barrier)? The best-effort behavior is still important, but this test with an absent directory does not pin the call sites.

realtonyyoung (changes requested):
Please preserve coverage that both marker writes and deletes invoke the directory-durability barrier; details are in the inline comment. I did not build or run tests.

test/Capacitor.Cli.Tests.Unit/Services/ServiceTxnMarkerTests.cs
Inok: Right that the call-site assertion is gone. It wasn't worth keeping, though.

A reassignable static is a test seam in production code, which is against policy here. It also proved less than it looked: it asserted that a *substituted delegate* was invoked. The real open/fsync/close never ran in that test, so a barrier that was called but broken passed it just as happily. A seam proves the call; running the real code proves it works.

What replaced it runs the actual libc path — the barrier fires on a real directory, reports false instead of throwing when it can't, and a failed barrier still lets the delete through.

What's genuinely unpinned is that Write and Delete reach the call. I tried returning the barrier's result: it leaks an implementation detail into the API, and Write creates the directory before writing into it, so its barrier can never fail — a hardcoded `true` is indistinguishable from a real one, confirmed by mutation. A properly injected barrier (required, not another optional Func) is the real answer, but it's 81 construction sites and 141 call sites.

realtonyyoung: Thanks for the detail. I rechecked the current `Write` and `Delete` call sites and the replacement tests against the real barrier. I agree that adding a test-only seam throughout this static API would be disproportionate here, so I’m withdrawing this as a blocking finding.

realtonyyoung:
NO FINDINGS — re-reviewed the current head and the reply to my durability-coverage comment. The production call sites still flush after write and delete, and the replacement tests exercise the real barrier. I did not build or run tests.

realtonyyoung (approved):

PR #999, opened 2026-09-18
[AI-2935] Report the harness list as JSON

George-Payne:
## What

`kcap harness list --json` emits the detected / wired / dismissed report as one JSON document on stdout and nothing else — the contract `kcap import --discover --json` already sets. The human table is untouched.

```json
{"harnesses":[{"vendor":"claude","label":"Claude Code","binary_on_path":true,
               "config_found":false,"wired":false,"dismissed":false}]}
```

## Why

First step of AI-2935. A tool setting kcap up for someone has to ask which coding agents to record, and the only honest option list is the one this machine produces. That list already exists — it drives the new-harness nudges — but only as a formatted table, so anything reading it is parsing display text.

## Shape, and why

- **Every known harness is listed, present or not.** A consumer can then tell "unsupported" from "not installed here" without carrying its own vendor list and going stale the day a vendor is added.
- **The two detection signals stay apart** — `binary_on_path` and `config_found` — rather than being ORed the way `HarnessInventory` folds them. Same reasoning as `FirstRunHarnessReport`: a caller offering someone a choice can say which signal it saw, and one that only wants "is it here" ORs them itself.
- **`vendor` is the stable key**, the id `dismiss` and `reset` take. `label` is display text.
- **`--json` is refused on `dismiss` and `reset`** rather than ignored, mirroring the `import --discover` guard — ignoring it would hand a caller expecting JSON a line of prose on a subcommand that writes.

## Testing

Five unit tests on the renderer, in the style of `ServiceStatusJsonTests`: registry order and completeness, snake_case keys, the two signals staying apart, an absent harness, and a dismissal read back from the ledger. Ran the real command in the dev container for both the list and the refusal.


realtonyyoung:
NO FINDINGS. Reviewed the harness-list JSON shape, detection/wiring states, dismissal handling, and the added coverage by inspection. I did not build or run tests.

realtonyyoung (dismissed):
Approved after code review; no actionable findings. Builds and tests were not run.

realtonyyoung:
NO FINDINGS on the updated head. I reviewed the four follow-up commits (ordering assertion, help text, renderer file split, and comment wording); they introduce no new actionable issue. I did not build or run tests.

realtonyyoung (approved):
Approved the updated head after reviewing the follow-up commits. No actionable findings; no build or tests run.

PR #1003, opened 2026-09-18
[AI-2935] Report status as JSON

George-Payne:
AI-2935 — no GitHub issue

## What & why

`kcap status --json` emits one JSON document on stdout and nothing else, so a tool driving a cold install can ask "is this machine set up?" without parsing lines written for a person. `configured` is true when a server is set and this CLI can authenticate to it — it holds credentials, or the server asks for none — and stays true when the server is unreachable: that is a network fact, and reading it as "not set up" would send someone through setup over a dropped VPN. Whether the server still accepts the credentials stays `kcap whoami`'s question.

## Where to look

- Auth resolves in the credential lane's order: no auth, machine, token store. The provider comes from the probe's own `/auth/config` answer, which status records in the provider cache, or from the last answer recorded there when the probe fails. Discovery's outage fallback is not used: it would call an unreachable server with an empty token store configured.
- The text path shares the gatherers, so it moves too: a token bound to another server gets its own Auth line, a server asking for no auth reads `not required`, and the Daemon line validates the start token.

## Verification

- `dotnet run --project test/Capacitor.Cli.Tests.Unit`: 4391 passed, 0 failed, 19 skipped.
- Reintroducing the one-variable-machine and wrong-server bugs fails exactly the three tests that name them, and nothing else.
- A no-auth server probed once, then stopped, still reports its provider: the test for it failed before status recorded its answer.
- `dotnet publish -c Release`: 0 IL2xxx/IL3xxx warnings. The published binary's `status --json` parses as one document, with an empty stderr.


realtonyyoung:
NO FINDINGS. Reviewed the status JSON output and the shared server, authentication, update, harness, and daemon probes by inspection. The earlier machine-credential, wrong-server, footer, and stale-PID concerns appear addressed in the current diff. I did not build or run tests.

realtonyyoung (dismissed):
Approved after code review; no actionable findings. Builds and tests were not run.

realtonyyoung (changes requested):
Finding: `ProbeServerAsync` reads the provider from a successful `/auth/config` response, but does not write that result to `AuthProviderCache`. Its failure path relies on that cache. For a server with no auth, `kcap status --json` can therefore report `configured: true` while reachable and `configured: false` after an outage if status was the only discovery (or the prior entry expired), despite no setup change. Please persist a successful discovery so the documented reachability-independent setup state survives a later failed probe. I did not build or run tests.

PR #1005, opened 2026-09-18
Report the workspaces discovery finds, without choosing one

George-Payne:
AI-2935 — no GitHub issue exists for this work

## What & why

`kcap setup --discover [--json]` signs in, reports the workspaces the account can reach, and stops. That list only exists on the far side of a sign-in, and `setup` learns it in the middle of a run it then carries to completion, so a tool that has to ask someone which workspace to use had no command to ask with.

It reaches the rows through the proxy directly rather than through the choosing flow, which selects a sole workspace before any picker is asked. Nothing is published — no profile, no activation, no token — so the run that follows signs in again; on a device-code flow that is a second approval.

## Where to look

- `can_create` means no workspace was found on the one lane that can create one. It is not a promise: a workspace already being made for the account is only learned by the create call.
- `--discover` takes a closed set (`--json`, `--github`, `--device`, `--no-prompt`); a workspace argument or any other option is refused by name.
- `kcap login --discover` keeps its own meaning: pick a workspace and save it.

## Verification

- `OnboardingFacadeDiscoverOnlyTests`: one, several and zero workspaces leave the config root empty, beside a control that publishes the same script through the choosing route.
- Core unit suite 3,411 passed, CLI unit suite 4,333 passed, 0 failed in either.
- Published AOT binary: `kcap setup --discover acme` and `kcap setup --discover --plugin-scope user` exit 1 with their refusal before any sign-in; the publish shows no IL2026/IL3050.
- Not covered: a complete `--discover` run against a live tenant, which needs an approval in a browser.


src/Capacitor.Cli/Commands/SetupCommand.cs
```diff
@@ -1713,6 +1715,99 @@ internal static (RequestedWorkspace? Workspace, string? Error) ParseRequestedWor
...
+    /// <remarks>A valued flag's value is not a bare token, so the values of the flags that take one
+    /// are excluded before the scan — otherwise <c>--default-visibility private</c> would read as a
+    /// workspace called "private".</remarks>
+    internal static bool NamesAWorkspace(string[] args) {
+        if (GetArg(args, "--server-url") is not null) return true;
+        if (GetArg(args, "--org") is not null) return true;
+        if (GetArg(args, "--slug") is not null) return true;
+
+        string[] valued = ["--server-url", "--org", "--slug", "--default-visibility", "--daemon-name", "--profile"];
+        var values = valued.Select(flag => GetArg(args, flag)).Where(v => v is not null).ToHashSet();
```
realtonyyoung: The values are excluded by text, not by their argument positions. For example, `kcap setup --discover --profile acme acme` treats both `acme` tokens as the `--profile` value, so the extra positional workspace is silently accepted and discovery starts instead of rejecting the invalid pairing. Skip each valued flag and only its following token while scanning, and cover a positional token that equals a flag value.

realtonyyoung (changes requested):
One argument-validation issue remains in `--discover`: a positional workspace can be missed when it matches another option’s value. I left a focused inline comment. This was a source review only; I did not build or run tests.

src/Capacitor.Cli/Commands/SetupCommand.cs
alexeyzimarev: Fixed in 167dfa42, by removing the value list rather than making it positional. `--discover` now takes a closed set (`--json`, `--github`, `--device`, `--no-prompt`, plus the process-wide `--no-update-check`), and the scan refuses the first argument outside it, so there is no set of values for a positional to hide in.

Your example is refused at `--profile` and discovery never starts. That is pinned by `A_workspace_repeating_an_earlier_value_does_not_start_discovery` (a245c52b), and the mirrored order, `--discover acme --daemon-name acme`, by `A_workspace_is_still_caught_when_a_later_value_repeats_it`.

Closed set rather than skipping each valued flag's following token, because a skip list has to know every flag that takes a value. The one here already missed `--plugin-scope` and `--use-provider-api-key`, so `--discover --plugin-scope user` was refused as a workspace called "user". The cost: options `--discover` used to ignore, such as `--daemon-name` and `--default-visibility`, are now refused by name.

Run against the published AOT binary: `kcap setup --discover acme`, `kcap setup acme --discover --json` and `kcap setup --discover --server-url <url>` each exit 1 with the workspace refusal, and `kcap setup --discover --plugin-scope user` exits 1 naming the flag, all before any sign-in.

PR #1016, opened 2026-09-18
Emit eval outcomes and post V4 results

realtonyyoung:
AI-2744 — no kcap-cli GitHub issue exists for this (driven from the Linear issue); the server half is kurrent-io/kcap-server#1939.

## What & why

The server now records an honest per-question evaluation outcome (`assessed` / `insufficient_evidence` / `not_applicable`) with a nullable score and a versioned evidence-coverage record, accepted at `POST /api/sessions/{id}/evals/v4`. This teaches the CLI judge and the dashboard-triggered daemon to emit the same: `ParseVerdict` yields an outcome-bearing assessment (an absent outcome reads as `assessed`, a present malformed one is a parse failure), a failed judge invocation becomes a coded failure instead of a silent null, `Aggregate` builds a V4 payload whose overall is the mean of assessed scores only (null when none scored), and the run POSTs V4. The daemon advertises eval protocol 2 at connect and answers the new `RunQuestionV2`/`FinalizeEvalV2` RPCs; the V1 RPCs and V3 posting stay for an older server.

## Where to look

`ReconcileEvidenceCoverage` — a complete coverage record cannot accompany an `insufficient_evidence` outcome (the server validator rejects that self-contradiction and 400s the whole run), so it is nulled to "unknown". The failure taxonomy in `ClaudeCliRunner.RunDetailedAsync`. `DaemonConnect` sending `eval_protocol_version: 2` (snake_case, pinned by a golden fixture). This PR merges only after the server route is deployed; until the submodule pin bump and npm release, the CLI still posts V3 and reads as fully assessed.

## Verification

Core and integration suites green, including `EvalRunnerV4PostTests` (exactly one POST to `/evals/v4`, none to `/evals/v3`), the D13 parser cases (absent vs present-`null` vs unknown outcome), the `ReconcileEvidenceCoverage` cases, and the `eval_protocol_version` golden fixture; `dotnet publish -c Release` is clean for both `kcap` and `kcap-daemon` (no AOT/trim warnings).


realtonyyoung:
Thanks Qodo — dispositions (9c7c490d):

**Fixed:**
- 🐞 Bug: an internal per-call deadline that fired while streaming stdin was classified as `ProcessFailure`; it now returns `Timeout` (→ `judge_timeout`). Added a runner test for the stdin-timeout path.
- 🐞 Bug: on the protocol-1 relay path an unassessed outcome sent a null-score completion *and* was reported as a failure by the V1 RPC — the observer now relays an unassessed outcome as a single failure (assessed still completes), so an older server gets one consistent signal. New observer test covers it. (protocol-2 / `silentPerQuestion` path unchanged.)
- Rule violations: `ParseVerdict` now uses the shared JsonElement helpers; the README eval section describes the outcomes, the not-scored states, coverage/citation output, and V4 posting.

**Leaving as-is:**
- 🐞 "all failed runs lose results": a run where *every* question is a coded failure (timeout/chat-error/parse-failure) is a failed run with no result by design — unassessed outcomes (`insufficient_evidence`/`not_applicable`) are assessments and still aggregate + post a V4 with a null overall. The server's V4 validator requires at least one category with at least one question, so the empty-category failure-only payload suggested here would be rejected (400).

PR #1026, opened 2026-09-18
Build policy decision events through one factory

### LINKED ISSUE #1025: Build policy decision events through one factory
`PolicyDecisionEventV1` has 16 positional parameters and is constructed positionally at four sites (`ClaudeHostedPolicySeam`, `AcpInteractionBridge`, and two in `ClaudePolicySeam`). An argument-order slip between same-typed string fields (seam/vendor/outcomes) would compile silently.

Add a `PolicyWire.Decision` factory that derives the invariant fields (engine version, decided-at, mode wire string, snapshot id/degraded) and route all constructors through it, so an ordering mistake is unrepresentable.

Linear: AI-2465

realtonyyoung:
Closes #1025 — AI-2465

## What & why

`PolicyDecisionEventV1` was built positionally at four seams, where a slip between adjacent same-typed strings (seam/vendor/outcomes) would compile silently. A single `PolicyWire.Decision` factory now builds every decision event: it derives the invariant fields — engine version, decided-at, the mode's wire spelling, and the snapshot id/degraded pair — so those never appear at a call site, and every remaining field is passed by name.

## Where to look

The null-snapshot arm folds to id `unknown` / undegraded, which was the evaluation-error path's contract and is now applied uniformly; the governed seams always pass a real snapshot, so their output is unchanged.

## Verification

Policy suites green with behaviour unchanged: Core `PolicyWireTests` 8/8 (two added — invariant derivation and the unknown-snapshot fallback), Cli 32/32, Daemon 132/132.


realtonyyoung:
Qodo rule violation (commit subject missing issue suffix): fixed in `22427e13` — subject is now `Build policy decision events through one factory (#1025)`.

